PR #3, opened 2016-06-28
Add CodeClimate test coverage reporter

Fryguy:
@durandom Please review.

Two weird things I noticed locally which may change how we have to run the tests

1.  After `app:test:manageiq-providers-amazon:setup` I see `.I, [2016-06-28T16:38:38.999289 #96868]  INFO -- : Reporting coverage data to Code Climate.`.  This is probably because we run two separate steps.  We might want to try to find a way to make `bundle exec rake` to do the right thing, but that might be tricky.
2. After `app:test:manageiq-providers-amazon` I see `Coverage = 89.35%. Sending report to https://codeclimate.com for branch aws_move_to_gem... done.`  I'm not sure why it's choosing that branch except that that's the branch we use for the "dummy" app.  Once everything is on master, this is fine, but it's still very weird.

I see a bunch of other small, minor, oddities we should do after this is merged, so I'll make issues for those.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,4 @@
 --require manageiq/spec/spec_helper
+--require spec_helper
 --color
 --order random
--- a/.rspec_ci
+++ b/.rspec_ci
@@ -1,4 +1,5 @@
 --require manageiq/spec/spec_helper
+--require spec_helper
 --color
 --order random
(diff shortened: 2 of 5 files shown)


durandom:
:+1:

PR #4, opened 2016-06-28
Update .gitignore

Fryguy:
- Update it to follow what a new gem would have + config values
- Add ignoring of /spec/manageiq dummy app

@durandom Please review


durandom:
:+1:

PR #7, opened 2016-06-29
Virtual Templates API

durandom:
contains the relevant changes from https://github.com/ManageIQ/manageiq/pull/9492
which got merged after we created the providers-amazon repo

@jntullo can you double check that this is the only change to the amazon provider?
@Fryguy pls merge


jntullo:
looks good 👍

PR #10, opened 2016-06-29
Update x1.32xlarge to enhanced and clustered networking.

Fryguy:
http://aws.amazon.com/ec2/instance-types/ updated, so reflecting the
changes here.

@blomquisg @durandom Please review.  I Think this should be darga/yes (which makes things complicated, so I'll have to review with @chessbyte how to backport if necessary).


blomquisg:
:+1: merge when green

durandom:
>  Think this should be darga/yes (which makes things complicated, so I'll have to review with @chessbyte how to backport if necessary).

I think this could be something like
* add this repo as a remote to manageiq checkout locally
* cherry-pick with `git cherry-pick -Xsubtree=gems/manageiq-providers-amazon <sha>`

PR #11, opened 2016-06-30
move vcr cassettes for amazon from manageiq repo over

durandom:
fixes https://github.com/ManageIQ/manageiq-providers-amazon/issues/2

@Fryguy review?


spec/spec_helper.rb
```diff
@@ -2,4 +2,5 @@
 CodeClimate::TestReporter.start
 VCR.configure do |config|
   config.ignore_hosts 'codeclimate.com'
+  config.cassette_library_dir = File.join(ManageIQ::Providers::Amazon::Engine.root, 'spec/vcr_cassettes')
```
durandom: I cant use `ENGINE_ROOT` here, because rspec is executed as a child process

PR #14, opened 2016-07-04
move creation of amazon cloudmanager with different authentication

durandom:
into spec file and remove the factory altogether

fixes CI build failures introduced by https://github.com/ManageIQ/manageiq/pull/7050

@miq-bot add_labels bug,test
@blomquisg review and merge?

### DIFF
--- a/spec/models/manageiq/providers/amazon/cloud_manager_spec.rb
+++ b/spec/models/manageiq/providers/amazon/cloud_manager_spec.rb
@@ -17,8 +17,10 @@ describe ManageIQ::Providers::Amazon::CloudManager do
   end
 
   describe ".discover" do
-    let(:ec2_user) { "0123456789ABCDEFGHIJ" }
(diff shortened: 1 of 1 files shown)


spec/models/manageiq/providers/amazon/cloud_manager_spec.rb
```diff
@@ -147,8 +149,8 @@ def create_ems_on_other_account(name)
           expect(emses.count).to eq(4)
           assert_region(emses[0], "us-east-1")
           assert_region_on_another_account(emses[1], "us-west-1")
-          assert_region_on_another_account(emses[2], "us-west-1 #{ec2_user}")
-          assert_region(emses[3], "us-west-1 1")
+          assert_region_on_another_account(emses[3], "us-west-1 #{ec2_user}")
```
durandom: I know, those array indexes are :-1: but so is this cloudmanager spec too, with its :vhs:

PR #16, opened 2016-07-06
Amazon CloudManager supports provisioning via automate

durandom:
removed hardcoded classes that support provisioning from miq_provision

Managers of providers answer the question wether they support
provisioning via automate or not. The BaseManager does not by default.

related PR https://github.com/ManageIQ/manageiq/pull/8351


bmclaughlin:
@blomquisg, this is needed in Darga per https://bugzilla.redhat.com/show_bug.cgi?id=1376010

PR #19, opened 2016-07-15
[WIP] Ensure right class of network manager

Ladas:
Ensure right class of network manager, rest of the code can be
moved to baseclass then.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -47,6 +47,10 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
   supports :regions
   supports :discovery
 
+  def ensure_network_manager
+    build_network_manager(:type => 'ManageIQ::Providers::Amazon::NetworkManager') unless network_manager
+  end
+
   def ensure_managers
     build_network_manager unless network_manager
     network_manager.name            = "#{name} Network Manager"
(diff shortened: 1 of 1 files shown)


durandom:
@Fryguy what about this ~~Harakiri~~ Hakiri guy? How can we fix his concerns?

Fryguy:
It looks like Hakiri just wants the version of Rails bumped, but shouldn't it be Rails 5?

durandom:
@Ladas can you rebase this branch and push again? Maybe this triggers Hakiri to rescan.
Re-Opening the PR doesnt work :(

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -47,6 +47,10 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
   supports :regions
   supports :discovery
 
+  def ensure_network_manager
```
durandom: Hmm, where is this called from? 
Isnt the `ensure_managers`  method where this should be?

Ladas: @durandom right, that was being leveraged in the next PR https://github.com/ManageIQ/manageiq-providers-amazon/pull/20/files

but I am no longer sure if it needs to be divided. Too many dependencies for my brain :-)

durandom:
@Ladas rebase to trigger CI to restart?

durandom:
@Ladas can you have a look why this is failing?

Ladas:
@bronaghs we should merge this in, on my TODO list, keep falling through the cracks :-D

djberg96:
@Ladas What's the current status of this?

Ladas:
@djberg96 it moved up, it's now on 1459th place in my TODO list. :-)

bronaghs:
srsly? Let's close this @Ladas.

Ladas:
@bronaghs I would like to get it in, but I will need to check what changed in the file, I just can't find time to do it. :-)

PR #25, opened 2016-08-10
Added Rubocop styling

juliancheal:
Keeping styling the same as ManageIQ styles, by reusing MiQ Rubocop


bdunne:
@juliancheal I think we should either:
- Inherit from [ManageIQ/guides](https://github.com/ManageIQ/guides/blob/master/.rubocop.yml) with only the overrides that apply to this repo

--or --
- Inherit from [ManageIQ/manageiq](https://github.com/ManageIQ/manageiq/blob/master/.rubocop.yml) with no changes

juliancheal:
> Inherit from ManageIQ/guides with only the overrides that apply to this repo
> --or --
> 
> Inherit from ManageIQ/manageiq with no changes

so in the `.rubocop.yml` file just put
```
inherit_from:
- https://raw.githubusercontent.com/ManageIQ/manageiq/master/.rubocop.yml
```

bdunne:
Exactly, my only concern is that we may not want everything from manageiq.  What do you think?

juliancheal:
Yeah I realised what you meant on closer inspection.
I've just removed other provider globals, and some things specific to MiQ it's self. Hopefully, (although not entirely sure what all the globals are for) this looks like a good minimum set for an external provider. :)

bdunne:
Thanks for removing the ManageIQ stuff @juliancheal 
Should we remove any global variables that are not referenced in this provider as well?  I assume we won't need stuff like `$vim_log` here.

juliancheal:
> Should we remove any global variables that are not referenced in this provider as well? I assume we won't need stuff like $vim_log, $fog_log, etc. here.

I wasn't sure what ones were needed, as I don't know what `$vim_log` is, but thinking about it the `aws_sdk` doesn't use `Fog` so that can go yes.

I'll narrow it down more, then worst case we can add any back in that we're missing :)

bdunne:
I agree, it's easy to add them later.  It might be worth a quick `git grep` to see what's used

durandom:
@juliancheal do you want to tweak the .rubocop.yml a little more? I think we should get @miq-bot monitor this little repo sooner than later :smile: 

@bdunne do you need anything else than this PR to make this happen?

juliancheal:
Oh yes I totally forgot. Thanks for the reminder.

bdunne:
I checked on the bot, and the repo was incorrectly added.  I corrected it, so it should start commenting soon.

juliancheal:
I've removed all the unused loggers, this should be good to merge.

bdunne:
:+1: LGTM

PR #26, opened 2016-08-11
Amazon elb refresh

Ladas:
LBaaS (Load Balancer as a Service) for Amazon. Patch contains STI models, refresh parser, specs and VCRs containing new HTTP communication

### DIFF
--- a/app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
@@ -6,7 +6,7 @@ class ManageIQ::Providers::Amazon::NetworkManager::RefreshParser
   def initialize(ems, options = nil)
     @ems        = ems
     @aws_ec2    = ems.connect
-    @aws_elb    = @ems.connect(:service => :ElasticLoadBalancing)
+    @aws_elb    = ems.connect(:service => :ElasticLoadBalancing)
     @data       = {}
     @data_index = {}
     @options    = options || {}
@@ -94,11 +94,11 @@ class ManageIQ::Providers::Amazon::NetworkManager::RefreshParser
     load_balancers.each do |lb|
       new_lb = @data_index.fetch_path(:load_balancer_pools, lb.load_balancer_name)
       load_balancer_pool_members = lb.instances.collect { |m| parse_load_balancer_pool_member(m) }
-      load_balancer_pool_members.each do |x|
+      load_balancer_pool_members.each do |member|
         # Store all unique pool members
-        if @data_index.fetch_path(:load_balancer_pool_members, x[:ems_ref]).blank?
-          @data_index.store_path(:load_balancer_pool_members, x[:ems_ref], x)
-          @data[:load_balancer_pool_members] << x
+        if @data_index.fetch_path(:load_balancer_pool_members, member[:ems_ref]).blank?
+          @data_index.store_path(:load_balancer_pool_members, member[:ems_ref], member)
+          @data[:load_balancer_pool_members] << member
         end
       end
 
@@ -232,8 +232,7 @@ class ManageIQ::Providers::Amazon::NetworkManager::RefreshParser
   end
 
   def parse_load_balancer_pool(lb)
-    uid  = lb.load_balancer_name
-    name ||= uid
+    uid = name = lb.load_balancer_name
 
     new_result = {
       :type          => self.class.load_balancer_pool_type,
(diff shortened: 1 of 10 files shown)


Ladas:
@durandom @blomquisg do you want me to separate specs to another PR, so this is smaller?

durandom:
> do you want me to separate specs to another PR, so this is smaller?

@Ladas nope, specs should be in the PR. 
Other than that, is this ready for review?

Ladas:
@durandom yes this is ready for a review, I am just fighting a bit about what everything needs to be in MIQ, so it;s passing CI :-)

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -6,6 +6,7 @@ class ManageIQ::Providers::Amazon::NetworkManager::RefreshParser
   def initialize(ems, options = nil)
     @ems        = ems
     @aws_ec2    = ems.connect
+    @aws_elb    = @ems.connect(:service => :ElasticLoadBalancing)
```
durandom: for consistency I would either use `ems.connect` or change above to `@ems.connect`

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -70,6 +79,49 @@ def get_outbound_firewall_rules(sg)
...
+    get_load_balancer_pool_members
+  end
+
+  def get_load_balancer_pool_members
+    @data[:load_balancer_pool_members] = []
+
+    load_balancers.each do |lb|
+      new_lb = @data_index.fetch_path(:load_balancer_pools, lb.load_balancer_name)
+      load_balancer_pool_members = lb.instances.collect { |m| parse_load_balancer_pool_member(m) }
+      load_balancer_pool_members.each do |x|
```
durandom: minor, but variable like `x` always scream to me :smile: 
I guess x is really a member? so how about  hitting F6 in RubyMine and change it to member?

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -167,6 +219,109 @@ def parse_firewall_rule(perm, direction)
...
+      :ems_ref => uid,
+      :name    => uid,
+    }
+
+    return uid, new_result
+  end
+
+  def parse_load_balancer_pool(lb)
+    uid  = lb.load_balancer_name
+    name ||= uid
```
durandom: where could `name` be defined already?

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -167,6 +219,109 @@ def parse_firewall_rule(perm, direction)
...
+  def parse_load_balancer_health_check(lb)
+    uid = lb.load_balancer_name
+
+    health_check_members = @aws_elb.client.describe_instance_health(:load_balancer_name => lb.load_balancer_name)
+    health_check_members = health_check_members.instance_states.collect do |m|
+      parse_load_balancer_health_check_member(m)
+    end
+
+    health_check = lb.health_check
+    target_match = health_check.target.match(/^(\w+)\:(\d+)\/?(.*?)$/)
```
durandom: as you are parsing an uri here, how about using `URI::parse` ?

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -167,6 +219,109 @@ def parse_firewall_rule(perm, direction)
...
+    protocol     = target_match[1]
+    port         = target_match[2].to_i
+    url_path     = target_match[3]
+
+    matched_listener = @data.fetch_path(:load_balancer_listeners).detect do |x|
```
durandom: you like `x`, right :smile:

durandom:
LGTM, although I dont understand the business model of a load balancer. 
@Ladas @blomquisg do we have somebody to review this from this perspective? Maybe @chriskacerguis ?

durandom:
Oh, and did I say I hate those :vhs: cassettes and the tests for the refresh? Next time we touch this code, we have to adjust all those numbers and strings that come from aws again. 
@blomquisg would be great if we (me) could refactor this soon to use api mocks

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: done

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: shift+F6 in here, cool :-)

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: removed

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: its not url really, they just encode it to string like TCP:22 or HTTP:8080/index.html

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: x,y and z are my best friends :-) it was inline before, which I changed because of too long line

Ladas:
@durandom I think the business model was already review for the MIQ models, so we should be ok there.

Yeah the vcr sucks, I did at least refactor it so it shows diff of all counts at once for now.

Ladas:
@durandom although for other roviders, I adjusted the counts just for >0 check. Then I rather check in detail if all relations are there and correct.

PR #28, opened 2016-08-12
Add gettext catalog to ManageIQ::Providers::Amazon plugin

mzazrivec:
Changes:
* added a gettext catalog to the `ManageIQ::Providers::Amazon` provider plugin. The catalog was
generated automatically by invoking the following rake task in the ManageIQ git checkout:
```
    rake locale:plugin:find[ManageIQ::Providers::Amazon]
```
* The `ManageIQ::Providers::Amazon` plugin now registers its own catalog with the main ManageIQ
rails app. This way the provider plugin makes the fast_gettext engine aware of the plugin's gettext catalog.

### DIFF
--- a/config/initializers/gettext.rb
+++ b/config/initializers/gettext.rb
@@ -1,0 +1,3 @@
+Vmdb::Gettext::Domains.add_domain('ManageIQ_Providers_Amazon',
+  ManageIQ::Providers::Amazon::Engine.root.join('locale').to_s,
+  :po)
--- a/locale/ManageIQ_Providers_Amazon.pot
+++ b/locale/ManageIQ_Providers_Amazon.pot
@@ -1,0 +1,27 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the ManageIQ_Providers_Amazon package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+#, fuzzy
+msgid ""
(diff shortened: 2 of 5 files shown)


durandom:
@mzazrivec can we do without the colon `:` in the filename? It was that on osx that it wouldnt be allowed, although that is no more

lib/manageiq/providers/amazon/engine.rb
```diff
@@ -3,6 +3,12 @@ module Providers
     module Amazon
       class Engine < ::Rails::Engine
         isolate_namespace ManageIQ::Providers::Amazon
+
+        config.to_prepare do
+          Vmdb::Gettext::Domains.add_domain('ManageIQ::Providers::Amazon',
```
durandom: can we move this to an initializer?

mzazrivec:
@durandom You mean the file names like `ManageIQ::Providers::Amazon.pot` ?

Yes, when we create the catalog, we can replace `::` with for example `_`. If this an issue, I can
change that, though I have to change the rake task in the main ManageIQ git first.

durandom:
In case we already have colons in files on miq side, its ok.
But in case not, I would try to avoid them, I think on windows they are still not allowed
Maybe with `underscore` or `tableize` inflections?

mzazrivec:
@durandom We don't have `::` in files in ManageIQ. I can change things to create file names
with underscores, rather than `::` (will do that change in ManageIQ git).

I also moved the initialization code into initializer (`gettext.rb) as you suggested.

mzazrivec:
@durandom I fixed both the thing with initialization and the file paths.

Issue #6, opened 2016-06-28
Reuse setup files from manageiq


Fryguy:
I'm wondering if instead of having a separate tools/ci/before_install.sh we can reuse the /spec/manageiq/bin/setup.  Or perhaps, change ManageIQ's tools/ci/before_install.sh to separate out those shared operations allowing it to be called from this repo's before_install.sh

durandom:
Thats mostly adressed by https://github.com/ManageIQ/manageiq/pull/13278

PR #30, opened 2016-08-18
[wip] EmsNetwork controller specs

Ladas:
EmsNetwork controller specs

### DIFF
--- a/spec/controllers/cloud_network_controller_spec.rb
+++ b/spec/controllers/cloud_network_controller_spec.rb
@@ -1,5 +1,5 @@
-require Rails.root.join('spec/shared/controllers/shared_cloud_network_controller_spec')
+require Rails.root.join('spec/shared/controllers/shared_examples_for_cloud_network_controller')
 
 describe CloudNetworkController do
-  include_examples :cloud_network_controller_spec, %w(amazon)
+  include_examples :shared_examples_for_cloud_network_controller, %w(amazon)
 end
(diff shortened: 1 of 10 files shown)


durandom:
https://github.com/ManageIQ/manageiq-providers-amazon/pull/32 is merged. please rebase and the tests should be failing

Ladas:
@durandom hm ,we might need to add some javascript libraries, so we can have tests that render something?

 couldn't find file 'es6-shim' with type 'application/javascript'

durandom:
But we run the specs inside manageiq as a dummy app.
Does it work locally for you?

Have a look at https://github.com/ManageIQ/manageiq-providers-amazon/pull/32
You can also run the tests from manageiq, does it work then?

Ladas:
@durandom yes, it works from ManageIQ, not sure what we are missing, maybe we need to add the javacript libs in some special way?

Ladas:
@Fryguy hello, would you know what do we miss in the CI setup, that we get Failure/Error: //= require es6-shim ?

Fryguy:
It sounds like the bower install stuff from ManageIQ is not happening, but I'm not exactly sure why.

Fryguy:
I wonder if this step is not sufficient: https://github.com/ManageIQ/manageiq-providers-amazon/blob/87bcab116e6b3a56ac5fb7862056715f84bf961d/bin/setup#L16-L28

From that it looks like it just does a bundle update, but I don't think it ever calls the bower install stuff.  @durandom I have to admit I don't understand the testing plumbing, so I'm not sure here.

durandom:
bower not running is a good shot. I'll re-investigate this with https://github.com/ManageIQ/manageiq-providers-amazon/issues/6 in mind.

Ladas:
@durandom we will need https://github.com/ManageIQ/manageiq/blob/master/.travis.yml#L25

PR #31, opened 2016-08-18
correct classnames

durandom:
fixes warnings. before this would not test the right classes

```
/home/travis/build/ManageIQ/manageiq-providers-amazon/spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb:540: warning: toplevel constant LoadBalancer referenced by ManageIQ::Providers::Amazon::CloudManager::LoadBalancer
```

in https://travis-ci.org/ManageIQ/manageiq-providers-amazon/builds/153278362#L499-L505

### DIFF
--- a/spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
@@ -537,7 +537,7 @@ describe ManageIQ::Providers::Amazon::CloudManager::Refresher do
   end
 
   def assert_specific_load_balancers
-    @elb_non_vpc = ManageIQ::Providers::Amazon::CloudManager::LoadBalancer.where(
+    @elb_non_vpc = ManageIQ::Providers::Amazon::NetworkManager::LoadBalancer.where(
       :name => "EmsRefreshSpec-LoadBalancer").first
     expect(@elb_non_vpc).to have_attributes(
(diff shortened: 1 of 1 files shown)


durandom:
@Ladas please check

Ladas:
@durandom ah yes, this is correct, hm how did this pass the tests?

durandom:
because it's just warnings

@blomquisg please merge

Ladas:
Hm,

ManageIQ::Providers::Amazon::CloudManager::LoadBalancer.class == ::LoadBalancer

@kbrock have you seen this? ^ For some reason, the non existent STI classes do not fail but return a base class. Is that some new thing? In the past, this would just fail with non existent class.

Ladas:
but yes @blomquisg , please merge :-)

durandom:
> ManageIQ::Providers::Amazon::CloudManager::LoadBalancer.class == ::LoadBalancer

http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#when-constants-aren-t-missed

I think this was the case always, thus only the warning

Ladas:
@durandom I see. String::LoadBalancer. Ok so here it throws a warning, but putting that class to STI type will fail with unknown constant String::LoadBalancer? Strange indeed

kbrock:
This reminds me of [aliases](https://github.com/ManageIQ/manageiq/tree/master/app/models/aliases) but since `LoadBalancers` is a global class, doesn't look necessary.

PR #32, opened 2016-08-18
run all specs under spec/ and not only models

durandom:
* removes ENGINE_ROOT to make the specs run from miq

```bash
~/src/manageiq master
❯ bundle exec rails test:providers:amazon
** master - DB vmdb_test on
** master - DB vmdb_test on

Randomized with seed 30066
.................................................................................................................

Finished in 13.21 seconds (files took 5.78 seconds to load)
113 examples, 0 failures

Randomized with seed 30066
```

* renames specs to providers:amazon for clarity

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ addons:
 before_install:
 - source ${TRAVIS_BUILD_DIR}/tools/ci/before_install.sh
 script:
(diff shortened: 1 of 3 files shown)


durandom:
@Fryguy thats needed for https://github.com/ManageIQ/manageiq-providers-amazon/pull/30

PR #33, opened 2016-08-22
Workaround for a secondary fixed IP for ENI

Ladas:
Purpose or Intent
-----------------

AWS can have secondary private IP address assigned to the ENI,
our current model does not allow that. Probably the best fix is,
to expand unique index of the cloud_subnet_network_ports to
include address. But as a backportable fix, we need to display
only the first one, since we can't backport a migration.

Steps for Testing/QA
-----------------

With this fix, having secondary fixed IP address on ENI will not fail refresh. Only first fixed IP will be visible though.


durandom:
@Ladas if you say that a secondary address will not fail the refresh, why slice then?

Ladas:
@durandom bad formulation :-) it will fail now. There is a db unique index that will fail now. We need to extend the index to contain [network_port_id, cloud_subnet_id,  address]. We can't backport that though.

PR #36, opened 2016-08-30
Fix collection of public ip addresses

Ladas:
The collection of the public IP addresses were broken and we were not collecting private addresses for the old ec2 instances.

Fixes Issue:
https://github.com/ManageIQ/manageiq/issues/10586

Fixes BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1371204

### DIFF
--- a/app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
@@ -24,6 +24,7 @@ class ManageIQ::Providers::Amazon::NetworkManager::RefreshParser
     get_load_balancer_pools
     get_load_balancer_listeners
     get_load_balancer_health_checks
+    get_ec2_floating_ips_and_ports
     get_floating_ips
     $aws_log.info("#{log_header}...Complete")
 
@@ -121,17 +122,22 @@ class ManageIQ::Providers::Amazon::NetworkManager::RefreshParser
     process_collection(load_balancers, :load_balancer_health_checks) { |lb| parse_load_balancer_health_check(lb) }
   end
 
-
   def get_floating_ips
     ips = @aws_ec2.client.describe_addresses.addresses
+    # Take only floating ips that are not already in stored by ec2 flaoting_ips
+    ips = ips.select { |floating_ip| @data_index.fetch_path(:floating_ips, floating_ip.public_ip).nil? }
     process_collection(ips, :floating_ips) { |ip| parse_floating_ip(ip) }
   end
 
   def get_network_ports
     network_ports = @aws_ec2.client.describe_network_interfaces.network_interfaces
-    instances     = @aws_ec2.instances
     process_collection(network_ports, :network_ports) { |n| parse_network_port(n) }
-    process_collection(instances, :network_ports) { |x| parse_network_port_inferred_from_instance(x) }
+  end
+
+  def get_ec2_floating_ips_and_ports
+    instances = @aws_ec2.instances.select { |instance| instance.network_interfaces.blank? }
+    process_collection(instances, :network_ports) { |instance| parse_network_port_inferred_from_instance(instance) }
+    process_collection(instances, :floating_ips) { |instance| parse_floating_ip_inferred_from_instance(instance) }
   end
 
   def parse_cloud_network(vpc)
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -121,17 +127,39 @@ def get_load_balancer_health_checks
...
+    # Take only floating ips that are not already in stored by ec2 flaoting_ips
+    ips = ips.select { |floating_ip| @data_index.fetch_path(:floating_ips, floating_ip.public_ip).nil? }
     process_collection(ips, :floating_ips) { |ip| parse_floating_ip(ip) }
   end
 
+  def get_public_ips
+    public_ips = []
+    network_ports.each do |network_port|
+      network_port.private_ip_addresses.each do |private_address|
+        if private_address.association && !(public_ip = private_address.association.public_ip).blank?
```
durandom: I wonder why rubocop does not complain about the assignement inside the conditional.
Could you move that into its own line?

spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
```diff
@@ -2,7 +2,7 @@
   before(:each) do
     _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
     @ems = FactoryGirl.create(:ems_amazon, :zone => zone)
-    @ems.update_authentication(:default => {:userid => "0123456789ABCDEFGHIJ", :password => "ABCDEFGHIJKLMNO1234567890abcdefghijklmno"})
+    @ems.update_authentication(:default => {:userid => "0123456789ABCDEFGHIJ", :password => "ABCDEFGHIJK LMNO1234567890abcdefghijklmno"})
```
durandom: are you sure the password changed for the specs?

spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
```diff
@@ -34,6 +34,7 @@
       assert_specific_vm_powered_on
       assert_specific_vm_powered_off
       assert_specific_vm_on_cloud_network
+      assert_specific_vm_on_cloud_network_1
```
durandom: could that be something more meaningful than `_1` ?
Has it something to do with floating ips?

durandom:
some minor stuff. And @Ladas could you also address miq-bots complaints (those that are reasonable)

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: if it's correctly in the (), the assignment should be valid :-) I use it all the time so save some lines :-)

spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
Ladas: hm, a space

spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
Ladas: will do

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
durandom: yeah, I know :)
I thought this was deprecated in our style guide, but its apparently not.

The thinking behind not using this, is that its easy to slip in a `==`, which could be avoided by [yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions), but nobody really uses them :)

So, I try to avoid this and rather have an additional and explicit line. And imho this makes the code most of the time more readable.

But in the end its also a matter of taste

spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
Ladas: done

spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
Ladas: done

Ladas:
@durandom fixed. I do ignore the rubocop _get and the next, which would lead to complex unless condition. ok?

PR #38, opened 2016-08-30
removed gov cloud from list of regions

durandom:
addresses https://bugzilla.redhat.com/show_bug.cgi?id=1346065

### DIFF
--- a/app/models/manageiq/providers/amazon/regions.rb
+++ b/app/models/manageiq/providers/amazon/regions.rb
@@ -56,11 +56,6 @@ module ManageIQ
           :hostname    => "ec2.sa-east-1.amazonaws.com",
           :description => "South America (Sao Paulo)",
         },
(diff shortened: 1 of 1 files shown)


durandom:
darga pr is at https://github.com/ManageIQ/manageiq/pull/10902

Fryguy:
Whoa...why did we remove this?

chessbyte:
@blomquisg @Fryguy @durandom should we revert this PR?

durandom:
@chessbyte yes, I think so. Can this be done via github buttons or do I have to do this via a PR?

blomquisg:
@durandom I think it needs to be a separate PR

durandom:
> I think it needs to be a separate PR

yes, will incorporate that into https://github.com/ManageIQ/manageiq-providers-amazon/issues/44

PR #43, opened 2016-09-06
Fix vpc floating ips refresh and association

Ladas:
In the provisioning workflow, we need to send allocation_id for a floating IP, otherwise the provisioning will fail, when deploying VPC VM

Fixes BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1373438
    
Fixes issue:
https://github.com/ManageIQ/manageiq/issues/10586

### DIFF
--- a/app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
@@ -350,14 +350,16 @@ class ManageIQ::Providers::Amazon::NetworkManager::RefreshParser
   end
 
   def parse_floating_ip(ip)
-    address = uid = ip.public_ip
+    cloud_network_only = ip.domain["vpc"] ? true : false
+    address            = ip.public_ip
+    uid                = cloud_network_only ? ip.allocation_id : ip.public_ip
 
     new_result = {
       :type               => self.class.floating_ip_type,
       :ems_ref            => uid,
       :address            => address,
       :fixed_ip_address   => ip.private_ip_address,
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -350,14 +355,16 @@ def parse_load_balancer_health_check_member(member)
   end
 
   def parse_floating_ip(ip)
-    address = uid = ip.public_ip
+    cloud_network_only = ip.domain["vpc"] ? true : false
```
durandom: This looks a bit strange. If this is a vpcAddress, then `ip.domain` should return a string and not hash

http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/VpcAddress.html

Ladas: it's a bit weird. Though this code is there from a beginning and works. :-) probably putting a debugger there and check out what is really there will be the best

durandom: ok, heads up, I tested it and yes, it returns a string. This would work because 
```ruby
"vpc"['vpc'] == 'vpc'
"standard"['vpc'] == nil
```

but thats not really intention revealing :)

so do you mind changing it to

```ruby
cloud_network_only = ip.domain == "vpc" ? true : false
```

Ladas: would you mind doing that in another PR? Since that logic was there from the beginning and it should not be this PR changing it.

PR #46, opened 2016-09-09
Change load balancer and instance port to range

Ladas:
Change load balancer and instance port to range


Ladas:
@durandom this is the CI fix

PR #48, opened 2016-09-21
move floating ip factory from manageiq

durandom:
Moving the floating_ip factory over here.

I'm thinking about creating factories for all STI models that are defined in a provider in a dynamic way.


Ladas:
Factory already registered: floating_ip_amazon , cool, so should be enough to remove them from ManageIQ and add them here

But that means move all specs using them here, right?

durandom:
> But that means move all specs using them here, right?

Not yet, see the explanation here https://github.com/ManageIQ/manageiq/pull/11413

blomquisg:
Merging a :red_circle: PR is scary ... But, see [the explanation](https://github.com/ManageIQ/manageiq/pull/11413#issuecomment-248706901) for more information.

PR #50, opened 2016-09-21
more intention revealing detection of vpc

durandom:
ok, heads up, I tested it and yes, it returns a string. This would work because 
```ruby
"vpc"['vpc'] == 'vpc'
"standard"['vpc'] == nil
```

but thats not really intention revealing :)

so do you mind changing it to

```ruby
cloud_network_only = ip.domain == "vpc" ? true : false
```

from this PR
https://github.com/ManageIQ/manageiq-providers-amazon/pull/43#discussion-diff-78062693

@blomquisg please review and merge


Ladas:
:+1: :+1: cool

PR #54, opened 2016-10-04
[EUWE] point euwe branch to manageiq euwe branch

durandom:
ci should run against miq euwe too

probably we can automagically discover what branch we are running on and use that.
but as we dont really want to use miq naming in here, I think explicitily setting it is fine


Fryguy:
I think we should merge this after verify that ManageIQ main repo's euwe branch has gone green

PR #55, opened 2016-10-06
remove stale methods that are superseded by supports :feature blocks

durandom:
`supports :migrate` was added in https://github.com/ManageIQ/manageiq/pull/9990
`supports :smartstate_analysis` was added in https://github.com/ManageIQ/manageiq/pull/9865


durandom:
@blomquisg please merge

PR #56, opened 2016-10-10
Stubbed specs for aws network manager

Ladas:
Stubbed specs for the Amazon NetworkManager


spec/models/manageiq/providers/amazon/aws_stubs.rb
```diff
@@ -0,0 +1,228 @@
...
+        :subnet_id => "subnet_#{i}"
+      }
+    end
+
+    {:subnets => mocked_subnets}
+  end
+
+  def stub_load_balancer_service
+    # TODO(lsmola) seems like Aws doesn't support stubbing of load balancing. Trying to pass stubs using
+    # Aws.config[:ElasticLoadBalancing] fails. We will stub it by rspec for now.
```
durandom: `Aws.config[:elasticloadbalancing] = { stub_responses => true}` works for me. 
Whats the error?

spec/models/manageiq/providers/amazon/network_manager/stubbed_refresher_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+    subject { described_class.discover(ec2_user, ec2_pass) }
+
+    before do
+      EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+
+      stub_load_balancer_service
+    end
+
+    around do |example|
+      with_aws_stubbed(:ec2 => stub_responses) do
```
durandom: If you stub the `elasticloadbalancer` service, you'd have to add that as a key to the options hash here

spec/models/manageiq/providers/amazon/network_manager/stubbed_refresher_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+        example.run
+      end
+    end
+
+    it "will perform a full refresh" do
```
durandom: I see, that you are coming from the `ems_refresh_spec.rb`, which is poorly structured.
now, as this is the only example in here, why not clean up the whole `before` `around` stuff?
Imho, just put it all (the whole setup and teardown) into the example here.

durandom:
I really like that you are stubbing everything and do without the vcr.
I can also see the stubbed spec replace the vcr one.

For the benchmark:
That spec looks the same as the stubbed one. How about:

* make the scaling be settable via an ENV variable
* run the stubbed spec via a script with different scalings
* that script can throw away the rails log and do the parsing

spec/models/manageiq/providers/amazon/network_manager/stubbed_refresher_spec.rb
Ladas: does it work for you? it couldn't find the service for me

spec/models/manageiq/providers/amazon/aws_stubs.rb
Ladas: ah

durandom:
merging as is
comments will be addressed in a follow up PR

kbrock:
good stuff

PR #57, opened 2016-10-11
bump aws-sdk version

durandom:
they say we should use the major version only when defining the
dependency https://github.com/aws/aws-sdk-ruby#installation

we are now locked at 2.2.19 whereas they are at 2.6.6
reading their [Changelog](https://github.com/aws/aws-sdk-ruby/blob/master/CHANGELOG.md)  its mostly API updates and mostly for stuff we dont use - but some API updates for EC2 and ElasticLoadBalancer 

@Fryguy ok?

### DIFF
--- a/manageiq-providers-amazon.gemspec
+++ b/manageiq-providers-amazon.gemspec
@@ -13,5 +13,5 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config.lib}/**/*"]
 
-  s.add_dependency("aws-sdk", ["~>2.2.19"])
+  s.add_dependency("aws-sdk", ["~>2"])
 end


durandom:
@Fryguy merge?

Ladas:
:+1: please merge

Ladas:
@Fryguy  @durandom although, I would prefer an exact version. If it updates automatically, it can potentially break CI in master and many released version. 

For fog-openstack, I moved to fixed version because of that, then we can see the CI failure when we send a PR with bumped version.

durandom:
I havent thought about the miq dependency constraint. Good point. This resolves now to 2.2.37.

As they are using semantic versioning, how about going to '~> 2.6' ?

durandom:
@agrare wdyt? should we lock this at `~> 2` or `~> 2.6` ?

agrare:
@durandom personally I'd prefer being more restrictive and put a PR in to update after its been tested by us, been burned by gem point releases breaking things unexpectedly
Not sure what @Ladas @Fryguy think though

durandom:
closing, beacuse we already have it at `~> 2.6` now

PR #59, opened 2016-10-12
Stubbed specs for aws network manager test deleting

Ladas:
We need to test also if we are deleting the data  correctly, if api does not return them anymore

Depends on:
https://github.com/ManageIQ/manageiq-providers-amazon/pull/60
https://github.com/ManageIQ/manageiq/pull/12021
https://github.com/ManageIQ/manageiq/pull/12020

### DIFF
--- a/spec/models/manageiq/providers/amazon/aws_stubs.rb
+++ b/spec/models/manageiq/providers/amazon/aws_stubs.rb
@@ -109,10 +109,9 @@ module AwsStubs
       }
     end
 
-    ec2_max_instances = (test_counts[:instance_vpc_count] + test_counts[:instance_ec2_count] - 1)
-    (test_counts[:instance_vpc_count]..ec2_max_instances).each do |i|
+    test_counts[:instance_ec2_count].times.each do |i|
       instances << {
-        :instance_id => "instance_#{i}"
+        :instance_id => "instance_ec2#{i}"
       }
     end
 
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -116,6 +116,8 @@ def get_load_balancer_pool_members
   end
 
   def get_load_balancer_listeners
+    # Initialize the data, otherwise the saving code will not catch this when empty
+    @data[:load_balancer_listeners] ||= []
```
Ladas: It fixes what the spec discovered. Without this definition, the listeners are not deleted when API returns []

durandom: so it's likely that we might need to backport it.
can you put this in a PR of its own then, because it can be backported without the other stuff then

Ladas: https://github.com/ManageIQ/manageiq-providers-amazon/pull/67

Ladas:
yes

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
durandom: I wonder why I can merge this PR, as https://github.com/ManageIQ/manageiq-providers-amazon/pull/67 modifies the same file here?!

Ladas: weird, I thought I rebased the commit out, let me do that again

Ladas:
@durandom ok, I think the rebase before failed, now it should be ok. Though it's weird it didn't throw a conflict.

PR #60, opened 2016-10-17
Advanced performace specs

Ladas:
Advanced performace specs. We want to measure create, update and
delete for all refresh configurations we support.


durandom:
I'll merge this, as it is just used for the performance spec.

But I still suggest to extract that to a script which runs the 'normal' spec, as suggested in the original PR.

PR #63, opened 2016-10-20
[WIP] Extend VCR specs to test also DTO refresh

Ladas:
Extend VCR specs to test also DTO refresh by extracting the common asserts used also in original refresh.


durandom:
I dont like that we are mixing the dto refresh specs with the normal refresh. 
This will make it painful to backport future PRs

Ladas:
@durandom those 2 are equvalent now, we will split them when we need to

But we need it tested. In the future, the VCRs will be different but the assert has pass any 'full' refresh

durandom:
> those 2 are equivalent now

Not really. The settings toggle using the Dto refresh or not

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -323,12 +323,6 @@ def parse_load_balancer_health_check(lb)
     port         = target_match[2].to_i
     url_path     = target_match[3]
 
-    matched_listener = @data.fetch_path(:load_balancer_listeners).detect do |listener|
```
durandom: Can you do just this in another PR?

durandom:
It would be great to have this run with the stubs and/or with the vcr

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: nope, this is needed for passing specs

Ladas:
@durandom yeah, but the stubs data are more stupid now, will be harder to put there the interconnection of the real data. But it can be done

durandom:
@Ladas can we close this?

Ladas:
@durandom still working on this, I have it separate locally, need to have some MIQ code merged first

Ladas:
already merged in another PRs

PR #64, opened 2016-10-25
Revert "Add amazon load balancer actions"

durandom:
`ServiceLoadBalancer` does not exist yet in MIQ 
its in https://github.com/ManageIQ/manageiq/pull/10974


durandom:
CI did not catch this, because classes are autoloaded there and there is no spec for that class so it was never loaded

PR #67, opened 2016-10-26
We always need to initialize the top level collection

Ladas:
We always need to initialize the top level collection,
otherwise it would not invoke deleting of all listeners in
DB when API returns none.


durandom:
euwe?

Ladas:
@durandom maybe, it's a well hidden bug

durandom:
>  maybe, it's a well hidden bug

as in, let's wait until someone finds it? or as in yes, before someone finds it?

Ladas:
@durandom no I mean the benefit of fixing is not that big :-) so 'euwe no' for now. That collection is not even exposed in the UI now, so that bug has close to zero impact.

PR #69, opened 2016-10-27
remove descendant loader config

durandom:
no need for adding this since this is in manageiq
https://github.com/ManageIQ/manageiq/pull/12221/commits/83300fbfa871afc08343daeb959299b9a52e4887

### DIFF
--- a/config/initializers/descendant_loader.rb
+++ b/config/initializers/descendant_loader.rb
@@ -1,2 +1,0 @@
-# make sure STI models are recognized
-DescendantLoader.instance.descendants_paths << ManageIQ::Providers::Amazon::Engine.config.root.join('app/models')


durandom:
cc @juliancheal for lenvo

durandom:
@blomquisg please merge if ok

blomquisg:
@durandom hrm, I don't see this in the core MIQ repo.  How is this obviated by the core repo?

durandom:
>  I don't see this in the core MIQ repo. 

See ManageIQ/manageiq@83300fb 
There we register all Rails Engines that are namespaced under `ManageIQ::Providers` as `Vmdb::Plugins` and [here](https://github.com/ManageIQ/manageiq/commit/83300fb#diff-2dffee21730102d591f668045fe85db7R15) we add `/app` to the path

PR #71, opened 2016-10-28
Generator

durandom:
apply stuff from generator

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-- '2.2.5'
+- '2.3.1'
 sudo: false
 cache: bundler
 env:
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -15,10 +15,3 @@ before_install:
 script:
 - bundle exec bin/rails app:test:providers:amazon:setup app:test:providers:amazon
 after_script: bundle exec codeclimate-test-reporter
-notifications:
-  webhooks:
-    urls:
-    - https://webhooks.gitter.im/e/7af45001fe6b7a4039f2
-    on_success: change
-    on_failure: always
-    on_start: never
```
Fryguy: This part should not be removed.  This will *always* be different repo-to-repo, or rather the line of the URL will always be different.

durandom: yup, it's just that the URL in this case is the same as for manageiq. 
can you give me one for the amazon room?
then I'll add it in this PR

durandom:
@Ladas can you have a look and merge if ok?

Ladas:
cool, generator for applying updates for all plugins :-)

looks great :+1:

PR #72, opened 2016-11-02
Amazon CloudManager DTO refresh

Ladas:
Amazon CloudManager DTO refresh


Ladas:
@durandom aaand Amazon Dto Cloud refresh is done :-)

PR #74, opened 2016-11-02
Remove O(n) lookup of listener

Ladas:
Remove O(n) lookup of listener, the association of health
check to listenere doesn't exist in AWS. It's not worth to
approximate it using O(n) lookup.


Ladas:
has to go with a spec change

PR #77, opened 2016-11-04
Updated Codeclimate gem

juliancheal:
Fixes this issue https://github.com/codeclimate/ruby-test-reporter#upgrading-from-pre-10-versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_install:
 - source ${TRAVIS_BUILD_DIR}/tools/ci/before_install.sh
 script:
 - bundle exec bin/rails app:test:providers:amazon:setup app:test:providers:amazon
+after_script: bundle exec codeclimate-test-reporter
 notifications:
   webhooks:
     urls:
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,8 @@
 gemspec
 
 group :test do
-  gem "codeclimate-test-reporter", :require => false
+  gem "simplecov", :require => false
+  gem "codeclimate-test-reporter", "~> 1.0.0", :require => false
 end
 
 # Declare any dependencies that are still in development here instead of in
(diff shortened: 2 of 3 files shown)


Fryguy:
@juliancheal You also have to update https://github.com/ManageIQ/manageiq-providers-amazon/blob/1a30b2877cc76b7419a556eef80ec3e03519cdae/spec/spec_helper.rb#L1-L4 as per the documentation on that link

Fryguy:
Also, you need to update the .travis.yml to send the new results, probably in an after_script step.

juliancheal:
Pfft who reads documentation. But seriously I'm not sure which bit in the docs you're referring to :(

Fryguy:
Oh, it's a little further up in the doc in the Installation section: https://github.com/codeclimate/ruby-test-reporter#installation, but that doesn't describe it very well.  Maybe I'll make a PR to ruby-test-reporter to clarify the documentation.

- Bullet 1 is this PR, so that's good
- Bullet 2 suggests starting simplecov, so you need to replace https://github.com/ManageIQ/manageiq-providers-amazon/blob/1a30b2877cc76b7419a556eef80ec3e03519cdae/spec/spec_helper.rb#L1-L4  with

    ```ruby
    require 'simplecov'
    SimpleCov.start
    ```
- Bullet 3 is the same as existing, so no change there
- Bullet 4 requires a change to the .travis.yml  I think you can just add the following after the `script:` section:

    ```yaml
    after_script: bundle exec codeclimate-test-reporter
    ```

juliancheal:
Specs failed, but I might not have had the newest version of the amazon codebase. So rebased and repushed.

Fryguy:
Merging anyway, since the changes are correct, and something else is broken.

juliancheal:
👍 to living on the edge.

PR #78, opened 2016-11-09
Update inherit_from to use local .rubocop_base.yml

jrafanie:
[skip ci]

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 inherit_from:
-- https://raw.githubusercontent.com/ManageIQ/guides/master/.rubocop.yml
+- .rubocop_base.yml
 #
 # Overrides
 #


Fryguy:
cc @durandom We should probably change the generator to make the rubocop configs consistent.  Please discuss with @jrafanie the best way to do this.

durandom:
I changed that in my generator PR https://github.com/ManageIQ/manageiq/pull/12209 which is yet to be merged.

I dont see another way to propagate changes from miq/.rubocop_base.yml than running the generator again and applying that patch.

This change is also in my PR to bring this repo in line with the generator output https://github.com/ManageIQ/manageiq-providers-amazon/pull/71/

Fryguy:
Closing since this line is part of #71, and I'd prefer merging it in with that.

PR #80, opened 2016-11-15
Stubbed amazon managers spec and fixing ci

Ladas:
Stubbed specs for dto and non dto for Amazon. Also fixing a CI due to recent DTO changes in MIQ

### DIFF
--- a/spec/models/manageiq/providers/amazon/aws_stubs.rb
+++ b/spec/models/manageiq/providers/amazon/aws_stubs.rb
@@ -33,6 +33,21 @@ module AwsStubs
     }
   end
 
+  def assert_do_not_delete
+    allow_any_instance_of(ApplicationRecord).to(
+      receive(:delete).and_raise("Not allowed delete operation detected. The probable cause is a wrong manager_ref"\
+                                 " causing create&delete instead of update"))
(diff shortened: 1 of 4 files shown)


spec/models/manageiq/providers/amazon/aws_stubs.rb
```diff
@@ -3,22 +3,51 @@ def scaling_factor
...
+      :subnet_count                                    => scaling * 20,
+      :network_port_count                              => scaling * 20,
+      :floating_ip_count                               => scaling * 20,
+      :security_group_count                            => scaling * 20,
+      :inbound_firewall_rule_per_security_group_count  => scaling * 5,
+      :outbound_firewall_rule_per_security_group_count => scaling * 5,
     }
   end
 
+  def assert_do_not_delete
```
durandom: can you move that to this `spec_common.rb` file?
I like that this `aws_stubs` only serves for stubbing. Having asserts in here feels wrong

Ladas: hm, it's really used only for the stubbed specs. Can we do that later? I have more refactorings, that could use extracting of asserts for stubbs

durandom: ok, I trust you to do it later :)

PR #81, opened 2016-11-15
Vcr specs for dto refresh

Ladas:
Vcr specs for dto refresh


app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -323,12 +323,6 @@ def parse_load_balancer_health_check(lb)
     port         = target_match[2].to_i
     url_path     = target_match[3]
 
-    matched_listener = @data.fetch_path(:load_balancer_listeners).detect do |listener|
```
durandom: why is this removed?

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -0,0 +1,863 @@
+module AwsRefresherSpecCommon
```
durandom: 👏 for moving this to a common file

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
Ladas: need it for passing of the specs, it's removed from the dto refresh already

durandom: hmm, but it removes the listeners?!
so it alters the current refresh code.

durandom: ok, but why don't we need it?
now it looks like we dont save `loadbalancer_listeners` anymore

PR #85, opened 2016-11-16
Ruby standards has specs ending with *_spec.rb

jrafanie:
I happened to be in this repo and wasn't sure if this file was incorrectly named on purpose or just by mistake.

@Ladas Please review


Ladas:
@jrafanie gah, good catch :-)

PR #86, opened 2016-11-21
move aws api provision spec from manageiq to provider

durandom:
This spec is heavily tied to amazon and should move over to the amazon gem.
It was introduced here https://github.com/ManageIQ/manageiq/pull/6388 

PR on miq: https://github.com/ManageIQ/manageiq/pull/12542
commit that removed it: ManageIQ/manageiq@21e617a9c6e6b5849d79fd61b4f86c38c6fdf0d


durandom:
@Ladas can you merge this?

Ladas:
@durandom seems to be alright :+1:

PR #89, opened 2016-12-02
Allow additional_regions to be specified in settings

agrare:
Add additional regions to be specified in settings, corresponding key in `config/settings.yml` added in https://github.com/ManageIQ/manageiq/pull/12965

https://bugzilla.redhat.com/show_bug.cgi?id=1389068

### DIFF
--- a/spec/models/manageiq/providers/amazon/regions_spec.rb
+++ b/spec/models/manageiq/providers/amazon/regions_spec.rb
@@ -15,4 +15,64 @@ describe ManageIQ::Providers::Amazon::Regions do
       expect(described_class.names).not_to include('us-gov-west-1')
     end
   end
+
+  context "add regions via settings" do
+    context "with no additional regions set" do
+      let(:settings) do
+        {:ems => {:ems_amazon => {:additional_regions => nil}}}
+      end
+
+      it "returns standard regions" do
+        stub_settings(settings)
+        expect(described_class.names).to eql(described_class::REGIONS.keys)
+      end
+    end
+
+    context "with one additional" do
+      let(:settings) do
+        {
+          :ems => {
+            :ems_amazon => {
+              :additional_regions => {
+                :"my-custom-region" => {
+                  :name => "My First Custom Region"
+                }
+              }
+            }
+          }
+        }
+      end
+
+      it "returns the custom regions" do
+        stub_settings(settings)
+        expect(described_class.names).to include("my-custom-region")
+      end
+    end
+
+    context "with additional regions and disabled regions" do
+      let(:settings) do
+        {
+          :ems => {
+            :ems_amazon => {
+              :disabled_regions   => ["my-custom-region-2"],
+              :additional_regions => {
+                :"my-custom-region-1" => {
+                  :name => "My First Custom Region"
+                },
+                :"my-custom-region-2" => {
+                  :name => "My Second Custom Region"
+                }
+              }
+            }
+          }
+        }
+      end
+
+      it "disabled_regions overrides additional_regions" do
+        stub_settings(settings)
+        expect(described_class.names).to     include("my-custom-region-1")
+        expect(described_class.names).not_to include("my-custom-region-2")
+      end
+    end
+  end
 end
--- a/app/models/manageiq/providers/amazon/regions.rb
+++ b/app/models/manageiq/providers/amazon/regions.rb
@@ -69,7 +69,10 @@ module ManageIQ
       }
 
       def self.regions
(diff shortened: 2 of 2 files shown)


agrare:
@durandom please review

agrare:
@durandom middle ground allowing you to add regions without moving all the regions to settings.

spec/models/manageiq/providers/amazon/regions_spec.rb
```diff
@@ -15,4 +15,19 @@
       expect(described_class.names).not_to include('us-gov-west-1')
     end
   end
+
+  context "adds additional regions from settings" do
+    it "includes regions from additional_regions" do
+      custom_region = {
+        "my-custom-region" => {
```
Fryguy: The region will probably be symbolized to keep all keys in settings.yml serialized, so we should take care of that.

Fryguy: e.g. Putting in the config as a string with

```yaml
:ems
  :ems_amazon:
    :additional_regions:
    - my-custom-region:
      :name: My Custom Region
      :hostname: localhost
      :description: Custom Region
```

Then pulling it back out you can see it's a Symbol:

```
::Settings.ems.ems_amazon.additional_regions[0].to_h
# => {:"my-custom-region"=>nil,
#     :name=>"My Custom Region",
#     :hostname=>"localhost",
#     :description=>"Custom Region"}
```

spec/models/manageiq/providers/amazon/regions_spec.rb
```diff
@@ -15,4 +15,19 @@
...
+          :name        => "My Custom Region",
+          :hostname    => "localhost",
+          :description => "Custom Region"
+        }
+      }
+
+      allow(Settings.ems.ems_amazon).to receive(:additional_regions).and_return(custom_region)
```
Fryguy: This is not an accurate way to stub settings.  Prefer using the `stub_settings` helper method.

Fryguy: Additonally, `additional_regions` will be an array, but this inserts a single hash instead.

spec/models/manageiq/providers/amazon/regions_spec.rb
agrare: Used `stringify_keys` so
```
Settings.ems.ems_amazon.additional_regions.to_h.keys
=> [:"custom-region-2", :"custom-region-1"]
```
Returns
```
ManageIQ::Providers::Amazon::Regions.names[-2..-1]
=> ["custom-region-2", "custom-region-1"]
```

spec/models/manageiq/providers/amazon/regions_spec.rb
agrare: Switched to `stub_settings`

I had pictured a hash rather than an array which matches the way `REGIONS` is structures so e.g. this was my config addition:
```
  :ems_amazon:
    :disabled_regions: []
    :additional_regions:
      :custom-region-2:
        :description: 2nd Custom Region
        :hostname: localhost
        :name: Second Custom Region
      :custom-region-1:
        :description: 1st Custom Region
        :hostname: localhost
        :name: First Custom Region
```

Would this be better as an array?

durandom:
@agrare LGTM - do you want to squash the commits before merging?

Also I guess this needs some docs added or we provide an example in the settings.yml
Maybe we do just the latter once we moved settings to providers.
And maybe we can do this first, so you can include the example in this PR?
Looks like there is no pressure from the BZ

agrare:
> Also I guess this needs some docs added or we provide an example in the settings.yml

Yes when we settle on the format I was thinking of leaving a commented out section of what a valid additional region would look like in this PR https://github.com/ManageIQ/manageiq/pull/12965

agrare:
Okay @durandom the PR to settings in the main repo was merged and I squashed the commits

agrare:
@miq-bot remove-label euwe/no
@miq-bot add-label euwe/yes

EUWE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1434152

simaishi:
Euwe backport details:

```
$ git log -1
commit 2f5140d58c5ecaffbaa18fc38624d056c940a3f5
Author: Marcel Hild <hild@b4mad.net>
Date:   Thu Dec 15 11:48:32 2016 +0100

    Merge pull request #89 from agrare/bz_1389068_custom_regions
    
    Allow additional_regions to be specified in settings
    (cherry picked from commit 771d65ad0b95ff908c965f11d6789b70e748d455)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1434152
```

PR #90, opened 2016-12-06
rubocop config

durandom:
* changes rubocop config to point to manageiq repo instead of manageiq/guides
* don't maintain our own copy of rubocop_cc.yml for codeclimate, but use their support for [external configurations](https://codeclimate.com/changelog/582495c32c33066f1b00191d)


durandom:
cc @jrafanie

.rubocop.yml
```diff
@@ -1,5 +1,5 @@
 inherit_from:
-- https://raw.githubusercontent.com/ManageIQ/guides/master/.rubocop.yml
+- https://raw.githubusercontent.com/ManageIQ/manageiq/master/.rubocop_base.yml
```
durandom: Am I right that we use the `.rubocop_base.yml` from ManageIQ/manageiq rather the one from the guides?

jrafanie: Yes @durandom but I think we should be using the local file: `.rubocop_base.yml`, not the URL

durandom: But then I have to maintain a style across all repos?
I thought having one in a central place was the point

.codeclimate.yml
```diff
@@ -34,6 +34,12 @@ engines:
   rubocop:
     enabled: true
     config: '.rubocop_cc.yml'
+prepare:
+  fetch:
+  - url: "https://raw.githubusercontent.com/ManageIQ/manageiq/master/.rubocop_base.yml"
+    path: ".rubocop_base.yml"
+  - url: "https://raw.githubusercontent.com/ManageIQ/manageiq/master/.rubocop_cc.yml"
+    path: ".rubocop_cc.yml"
```
jrafanie: Ok, I'm following, so we fetch the _base and _cc from the main repo as part of the codeclimate run... we tell codeclimate to use .rubocop_cc.yml which pulls in the now local .rubocop_base.yml fetched from the main repo.

What do we do about provider changes to rubocop.yml?

.rubocop.yml
jrafanie: I was misunderstanding... commented above...

.codeclimate.yml
jrafanie: in other words, this PR drops all changes on top of the main repos versions... how do we apply any customizations we want for a specific repo?

durandom:
@jrafanie do you think this good, how I split the files? Or do you prefer another way?

jrafanie:
@durandom I think this is good for now, let's see if it works as we want and refine it as we go.

jrafanie:
@durandom Merging, let's see how it goes

PR #91, opened 2016-12-13
moving settings from manageiq

durandom:
I will remove the settings from manageiq/config/settings.yml once this is merged

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,7 +3,6 @@
   :ems_amazon:
     :disabled_regions: []
     # disable regions by their keys found in app/models/manageiq/providers/amazon/regions.rb e.g.
-    # uncomment the following to disable gov-west-1 region
     # - us-gov-west-1
     #
     # add additional regions as found in app/models/manageiq/providers/amazon/regions.rb
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,0 +1,46 @@
+---
+:ems:
+  :ems_amazon:
+    :disabled_regions: []
+    # disable regions by their keys found in app/models/manageiq/providers/amazon/regions.rb e.g.
+    # uncomment the following to disable gov-west-1 region
+    # - us-gov-west-1
+    #
+    # add additional regions as found in app/models/manageiq/providers/amazon/regions.rb
+    :additional_regions: {}
+    # :spec-region-1:
+    #   :name: Special Region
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -0,0 +1,46 @@
...
+    # disable regions by their keys found in app/models/manageiq/providers/amazon/regions.rb e.g.
+    # uncomment the following to disable gov-west-1 region
+    # - us-gov-west-1
+    #
+    # add additional regions as found in app/models/manageiq/providers/amazon/regions.rb
+    :additional_regions: {}
+    # :spec-region-1:
+    #   :name: Special Region
+    #   :hostname: ec2.spec-region-1.amazonaws.com
+    #   :description: Super Special Region
```
durandom: @agrare I added this as dummy example

agrare: @durandom Nice!

durandom:
@blomquisg care to merge this?

config/settings.yml
```diff
@@ -0,0 +1,46 @@
+---
+:ems:
+  :ems_amazon:
+    :disabled_regions: []
+    # disable regions by their keys found in app/models/manageiq/providers/amazon/regions.rb e.g.
+    # uncomment the following to disable gov-west-1 region
```
Fryguy: I would remove this statement.  We don't want *users* editing this file at all.  Instead they would go either go through Advanced Settings (which stores the changes in the database) or we would have them deploy a settings.local.yml or something.

Fryguy: Note that this comment will not appear in Advanced Settings, so who is the audience for this comment?

durandom: You are right about the `uncomment` statement. 

> Note that this comment will not appear in Advanced Settings

Yeah, I know, but I havent found a better place to add some configuration hints. Same as [here](https://github.com/ManageIQ/manageiq/blob/master/config/settings.yml#L121)

I think it's better than nothing. And advanced users _might_ find their way by googleing or looking at the source. And if its only support staff of ours.

Would it be better to approach the docs team to add that to the docs?

Fryguy: Yeah I was just thinking docs.  That seems the most appropriate place to document new settings.  You may be able to contribute directly in https://github.com/ManageIQ/manageiq_docs, particularly if they already have section somewhere for Advanced Settings.

durandom: Ok, will open a PR to that repo 👍

PR #92, opened 2016-12-15
make spec the default rake task

durandom:
Following somewhat the setup in https://github.com/ManageIQ/manageiq-content/blob/master/Rakefile

now you can just `bundle exec rake` to run the tests.

I think it's worth keeping the tasks at [lib/tasks/spec.rake](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/lib/tasks/spec.rake) as they can be run within the manageiq app then as well.

I also added a explaining error to `Gemfile` in case `spec/manageiq` is not there

@bdunne @Fryguy wdyt?

If this is good, I'd add that to the provider generator

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,14 +1,13 @@
-begin
-  require 'bundler/setup'
-rescue LoadError
-  puts 'You must `gem install bundler` and `bundle install` to run rake tasks'
-end
-
+require 'bundler/setup'
 require 'bundler/gem_tasks'
-require 'rspec/core/rake_task'
 
-APP_RAKEFILE = File.expand_path("../spec/manageiq/Rakefile", __FILE__)
-load 'rails/tasks/engine.rake'
+begin
+  require 'rspec/core/rake_task'
+
+  APP_RAKEFILE = File.expand_path("../spec/manageiq/Rakefile", __FILE__)
+  load 'rails/tasks/engine.rake'
+rescue LoadError
+end
 
 namespace :spec do
   task :setup => 'app:test:providers:amazon:setup'
--- a/Rakefile
+++ b/Rakefile
@@ -10,8 +10,11 @@ rescue LoadError
 end
 
 namespace :spec do
+  desc "Setup environment for specs"
   task :setup => 'app:test:providers:amazon:setup'
 end
 
+desc "Run all amazon specs"
 task :spec => 'app:test:providers:amazon'
+
 task :default => :spec
--- a/Rakefile
+++ b/Rakefile
@@ -4,5 +4,15 @@ rescue LoadError
   puts 'You must `gem install bundler` and `bundle install` to run rake tasks'
 end
 
+require 'bundler/gem_tasks'
+require 'rspec/core/rake_task'
+
 APP_RAKEFILE = File.expand_path("../spec/manageiq/Rakefile", __FILE__)
 load 'rails/tasks/engine.rake'
+
+namespace :spec do
+  task :setup => 'app:test:providers:amazon:setup'
+end
+
+task :spec => 'app:test:providers:amazon'
+task :default => :spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@ env:
 addons:
   postgresql: '9.4'
 before_install: bin/setup
-script:
-- bundle exec bin/rails app:test:providers:amazon:setup app:test:providers:amazon
 after_script: bundle exec codeclimate-test-reporter
 notifications:
   webhooks:
--- a/lib/manageiq-providers-amazon.rb
+++ b/lib/manageiq-providers-amazon.rb
@@ -1,0 +1,2 @@
+require "manageiq/providers/amazon/engine"
+require "manageiq/providers/amazon/version"


Gemfile
```diff
@@ -14,4 +14,8 @@ end
 # your gem to rubygems.org.
 
 # Load Gemfile with dependencies from manageiq
-eval_gemfile(File.expand_path("spec/manageiq/Gemfile", __dir__))
+MANAGEIQ_GEMFILE = File.expand_path("spec/manageiq/Gemfile", __dir__)
+unless File.exist?(MANAGEIQ_GEMFILE)
+  raise 'You must run `bin/setup` to install manageiq into spec/manageiq as a rails dummy app for this provider engine gem'
```
durandom: @NickLaMuro is that an appropriate way to issue that notice?

NickLaMuro: @durandom Hmm... well, you might run into a "chicken or the egg" problem since this error would be raised when `bundler` is parsing the Gemfile, not after it has installed the gems prior to this line, meaning Rails or any other gems required by `bin/setup` might not be available to someone who is using a clean gemset or doing this on a fresh install of Ruby (FYI, as of writing this, I haven't looked at `bin/setup` for this repo... so I might be full of it).

Would this make more sense at the top of the `Rakefile` instead?  Then when someone tries running the tests, they will get the error there instead.

durandom: > Would this make more sense at the top of the Rakefile instead? Then when someone tries running the tests, they will get the error there instead.

That was my first approach. But you will run `bundle exec rake`, so without having `spec/manageiq/Gemfile` it would stop at `bundle` already.

`bin/setup` is PORO. 

Are you suggesting in the Gemfile:
```ruby
eval_gemfile(MANAGEIQ_GEMFILE) if File.exist?(MANAGEIQ_GEMFILE)
```
and issue an error in the `Rakefile`?

NickLaMuro: > `bin/setup` is PORO.

Now that I look at it again, you are right, that would be without issue since you clone then `bundle` right after.  Subsequent commands following that would then have the gems install and it would work fine.

Truth be told, I think putting this either in the Gemfile or the Rakefile is fine, and your suggestion for how to do it in the Rakefile would have probably been what I would have come up with myself.  My only suggestion might be using a:

```ruby
puts "<<YOUR ERROR MESSAGE HERE>>"
exit 1
```

for the error message instead, as it will avoid extra noise from `rake`/`bundler`'s error handling mechanisms.  I assume you just want to deliver a simple message to your users that all of the dependencies haven't been installed (stacktraces can make this unclear in my opinion).

I think the only reason I shy away from putting this Gemfile is traditionally (from my point of view), I tend to see the Gemfile as a list of gems to install, and ideally nothing more.  But this is an interesting and non-standard Gem setup, and I wouldn't call myself a "foremost expert" on the subject, so take my opinion with a grain of salt.

NickLaMuro: Hmm... then maybe my suggestion is better for the rake task then...

Fryguy: We had a similar problem in manageiq-content and solved it similarly.  We should just make them all consistent.  See  https://github.com/ManageIQ/manageiq-content/blob/master/Gemfile#L5-L13

Gemfile
```diff
@@ -14,4 +14,11 @@ end
 # your gem to rubygems.org.
 
 # Load Gemfile with dependencies from manageiq
-eval_gemfile(File.expand_path("spec/manageiq/Gemfile", __dir__))
+manageiq_gemfile = File.expand_path("spec/manageiq/Gemfile", __dir__)
+if File.exist?(manageiq_gemfile)
+  eval_gemfile(manageiq_gemfile)
+else
+  puts "ERROR: The ManageIQ application must be present in spec/manageiq."
+  puts "  run `bin/setup` to clone it from GitHub or symlink it from local source."
```
durandom: @Fryguy I slightly tweaked your wording to indicate that there is `bin/setup` to do the job

durandom:
@Fryguy ok with it? Merge?
I'm fine with spending a longer time on the setup, as this redu

Rakefile
```diff
@@ -4,5 +4,15 @@ rescue LoadError
...
 end
 
+require 'bundler/gem_tasks'
+require 'rspec/core/rake_task'
+
 APP_RAKEFILE = File.expand_path("../spec/manageiq/Rakefile", __FILE__)
 load 'rails/tasks/engine.rake'
+
+namespace :spec do
+  task :setup => 'app:test:providers:amazon:setup'
```
bdunne: If you load your rake files before line 10&11, they won't be prefixed with `app:`.

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -1,12 +1,30 @@
 module ManageIQ::Providers::Amazon::CloudManager::EventParser
+  # these have their own EmsEvent classes in automate
```
bdunne: I don't think this change belongs in this PR.

durandom: hugh, what did git do to me...

durandom:
> I think it's worth keeping the tasks at lib/tasks/spec.rake as they can be run within the manageiq app then as well.

I know that we dont want to run the provider specs in the main manageiq repo. (As in only `vmdb` suite) - But as we split out more repos, I think those tests are still so interdependent, that we might want to move to a staged CI.

* run vmdb specs on PR
* tell miq-bot to merge
* run all specs (providers, content, ui, etc) before merge
* maybe run QE specs before merge

Thus, having the gem specs in `lib/tasks` is convenient, imho.
To get rid of `app:` in the Rakefile I can use `ManageIQ::Content::Engine.load_tasks` from  https://github.com/ManageIQ/manageiq-content/pull/7/commits/9aa54d4ecf7522f47c0cdb6743d68d020912aaf7 

cc @Fryguy @bdunne @chessbyte @blomquisg

Fryguy:
> To get rid of app: in the Rakefile I can use ManageIQ::Providers::Amazon::Engine.load_tasks from

Correct, but as @bdunne mentioned you have to move it before we load the app engine tasks.  See https://github.com/ManageIQ/manageiq-content/commit/4ee8bc2db70fadaa913baae3cc7c342ec1ab0ae3

Rakefile
```diff
@@ -4,5 +4,15 @@ rescue LoadError
   puts 'You must `gem install bundler` and `bundle install` to run rake tasks'
 end
```
Fryguy: Frankly you can get rid if this begin/rescue here.

Basically, can you just steal the Rakefile from manageiq-content directly?  I can't see why they would be any different.

durandom:
@bdunne @Fryguy I was thinking it makes sense to be able to run the provider test suites from the manageiq core repo, because staged CI et. al: https://github.com/ManageIQ/manageiq-providers-amazon/pull/92#issuecomment-268356230

Thats why I wanted to keep the spec Rake tasks in [lib/tasks/spec.rake](https://github.com/ManageIQ/manageiq/blob/master/lib/generators/provider/templates/lib/tasks/spec.rake)

If you both think it's not needed, I'll remove those from spec.rake and the generator and move them to the Rakefile (which indeed can be almost the same as in manageiq-content then).

Rakefile
durandom: > Frankly you can get rid if this begin/rescue here.

ok, that's template code from `rails generate plugin`

PR #96, opened 2017-01-05
Pull default rubocop configuration from ManageIQ/guides

bdunne:
Depends on ManageIQ/guides#178


Fryguy:
Merged ManageIQ/guides#178

Fryguy:
Closing in favor of #97 which included the change at the generator level as well.

PR #98, opened 2017-01-06
Inventory abstraction for data fetching and storing in inventory collections

Ladas:
Inventory abstraction for data fetching and storing in inventory collections

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/targets/event_payload_vm.rb
+++ b/app/models/manageiq/providers/amazon/inventory/targets/event_payload_vm.rb
@@ -1,0 +1,32 @@
+class ManageIQ::Providers::Amazon::Inventory::Targets::EventPayloadVm < ManageIQ::Providers::Amazon::Inventory::Targets
+  def initialize_inventory_collections
+    instance_ems_ref = event_payload(target)["instance_id"]
+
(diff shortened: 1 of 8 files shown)


Ladas:
@durandom @agrare inventory abstraction, so in this way, the parsing code will be the same for everything, then the whole refresh is defined in Inventory::Targets::XY

app/models/manageiq/providers/amazon/inventory/targets/event_payload_vm.rb
```diff
@@ -0,0 +1,25 @@
+class ManageIQ::Providers::Amazon::Inventory::Targets::EventPayloadVm < ManageIQ::Providers::Amazon::Inventory::Targets
+  def initialize_inventory_collections
+    instance_ems_ref = target[:full_data]["configurationItem"]["resourceId"]
```
durandom: so this whole Target class only fills the `ems_ref` and the rest is coming from the db?

Ladas: The ems_ref is for InventoryCollections init, so we know how to compare it. Then the data itself comes from event, in the instances method.

Which will result in create/update/delete based on the state of our db and the event

PR #101, opened 2017-01-11
Introduce Amazon Block Storage  Manager (EBS)

gberginc:
This commit sets up the core structure of the new AWS EC2 block storage
manager. It provides just the basic stubs and tests that are to be
extended to allow for complete management of Amazon block storage

The refresher collects data about existing volumes and private
snapshots. Volumes are not yet linked to disks of virtual machines
collected by the cloud manager.

Depends on [Add storage manager mixins](https://github.com/ManageIQ/manageiq/pull/13384) PR from core repo.

### DIFF
--- a/app/models/manageiq/providers/amazon/block_storage_manager/refresh_parser_inventory_object.rb
+++ b/app/models/manageiq/providers/amazon/block_storage_manager/refresh_parser_inventory_object.rb
@@ -1,0 +1,63 @@
+class ManageIQ::Providers::Amazon::BlockStorageManager::RefreshParserInventoryObject < ::ManagerRefresh::RefreshParserInventoryObject
+  include ManageIQ::Providers::Amazon::RefreshHelperMethods
+
+  def populate_inventory_collections
+    log_header = "MIQ(#{self.class.name}.#{__method__}) Collecting data for EMS name: [#{inventory.ems.name}] id: [#{inventory.ems.id}]"
+
+    $aws_log.info("#{log_header}...}")
+    get_volumes
+    get_snapshots
+    $aws_log.info("#{log_header}...Complete")
+
+    inventory_collections
+  end
+
+  private
+
+  def get_volumes
+    process_inventory_collection(inventory.cloud_volumes, :cloud_volumes) { |volume| parse_volume(volume) }
+  end
+
+  def get_snapshots
+    process_inventory_collection(inventory.cloud_volume_snapshots, :cloud_volume_snapshots) { |snap| parse_snapshot(snap) }
+  end
+
+  def parse_volume(volume)
+    uid = volume['volume_id']
+
+    {
+      :type          => self.class.volume_type,
+      :ems_ref       => uid,
+      :name          => uid,
+      :status        => volume['state'],
+      :creation_time => volume['create_time'],
+      :volume_type   => volume['volume_type'],
+      :size          => volume['size'].to_i.gigabytes
+    }
+  end
+
+  def parse_snapshot(snap)
+    uid = snap['snapshot_id']
+
+    {
+      :ems_ref       => uid,
+      :type          => self.class.volume_snapshot_type,
+      :name          => snap['snapshot_id'],
+      :status        => snap['state'],
+      :creation_time => snap['start_time'],
+      :description   => snap['description'],
+      :size          => snap['volume_size'],
+      :cloud_volume  => inventory_collections[:cloud_volumes].lazy_find(snap['volume_id'])
+    }
+  end
+
+  class << self
+    def volume_type
+      "ManageIQ::Providers::Amazon::BlockStorageManager::CloudVolume"
+    end
+
+    def volume_snapshot_type
+      "ManageIQ::Providers::Amazon::BlockStorageManager::CloudVolumeSnapshot"
+    end
+  end
+end
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -58,8 +58,7 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
            :to        => :block_storage_manager,
            :allow_nil => true
 
-  before_create :ensure_managers,
-                :ensure_block_storage_managers
+  before_create :ensure_managers
 
   supports :provisioning
   supports :regions
@@ -70,14 +69,11 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
     network_manager.name            = "#{name} Network Manager"
     network_manager.zone_id         = zone_id
     network_manager.provider_region = provider_region
-  end
 
-  def ensure_block_storage_managers
     build_block_storage_manager unless block_storage_manager
     block_storage_manager.name            = "#{name} Block Storage Manager"
     block_storage_manager.zone_id         = zone_id
     block_storage_manager.provider_region = provider_region
-    true
   end
 
   def self.ems_type
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -29,6 +29,12 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
           :autosave    => true,
           :dependent   => :destroy
 
+  has_many :storage_managers,
+           :foreign_key => :parent_ems_id,
+           :class_name  => "ManageIQ::Providers::StorageManager",
+           :autosave    => true,
+           :dependent   => :destroy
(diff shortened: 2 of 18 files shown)


gberginc:
@aiperon please review

gberginc:
@agrare could https://github.com/ManageIQ/manageiq/commit/d557f190a406dc7671fe246716806f93db1933b5 be affecting travis tests?

```
     Failure/Error:
       def initialize(inventory)
         @inventory             = inventory
         @inventory_collections = inventory.inventory_collections
       end
     
     ArgumentError:
       wrong number of arguments (given 2, expected 1)
```

Is there any doc on how to approach the refresh parsers now?

agrare:
@gberginc yeah it is failing for one of my branches using `InventoryObjectRefresh` refresh also.
@Ladas can you change `RefreshParserInventoryObject` so it doesn't have to use the experimental inventory abstraction?

gberginc:
Now that you mentioned another branch, I checked also other PRs in Amazon repo and noticed the one from @Ladas #102.

I guess it's best to wait until #102 is passing and then ensure storage manager is using the same approach.

Ladas:
@gberginc the CI should be fixed, but I've caused a conflict in the specs, you'll need to rebase

gberginc:
@Ladas thanks. The conflict is simple to resolve as just the number of `ext_management_system`s is different. However I guess it would be better to use the same approach for storage manager as it is for cloud/network. I'll try to update the PR.

Another thing I wanted to ask: I currently only use stubbed specs, but there are also those that are using actual AWS account - how should I approach this as using our account will change the VCR as well as all the existing spec data.

Ladas:
@gberginc cool

So the VCR kinda sucks there, because we have a VCR env, that we use for that. So option would be that I would record it for you, once you are ready, or we would share some readonly creds for the VCR env. So for now, I think easiest will be that I will record and send the VCRs, once you have all the API calls in place.

We were thinking about getting away from the VCRs and just use stubbed specs, but then it's nice to have a spec that is using real env. data.

gberginc:
@Ladas I've just added additional commit for the inventory object support. I am not sure if you are the one to review this PR (if not, please let me know so that I can ask someone else).

Regarding VCR recording: do I understand correctly, I should now add `EmsRefresh.refresh(@ems.block_storage_manager)` everywhere VCR cassettes are used, for example:

https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/spec/models/manageiq/providers/amazon/cloud_manager/refresher_inventory_object_spec.rb#L38

Then you run the spec and share the VCR with me so that I can update the actual tests?

Ladas:
@gberginc I'll do the review

Regarding the VCR, yes, lets do the VCR specs as a next PR

sergio-ocon:
Hey @Ladas, did you review it?

gberginc:
I think this is partly dependent on the current changes to the refresh parsers, https://github.com/ManageIQ/manageiq-providers-amazon/pull/105.

I would nevertheless appreciate feedback to continue working on storage managers.

app/models/manageiq/providers/amazon/block_storage_manager/refresh_parser_inventory_object.rb
```diff
@@ -0,0 +1,64 @@
...
+    $aws_log.info("#{log_header}...}")
+    get_volumes
+    get_snapshots
+    $aws_log.info("#{log_header}...Complete")
+
+    inventory_collections
+  end
+
+  private
+  def get_volumes
```
aiperon: We're using the same parsing code for RefreshParser and RefreshParserInventoryObject.
Wouldn't it be better if we move it into ManageIQ::Providers::Amazon::RefreshHelperMethods ?

Ladas: there is a slight difference in the parsing code. But yeah, we can abstract the parsing part. But that can be done in another PR.

But my goal is to deprecate/delete the old parsing code and use the new one, so maybe it won't be needed. :-)

app/models/manageiq/providers/amazon/block_storage_manager/refresh_parser_inventory_object.rb
```diff
@@ -0,0 +1,64 @@
...
+    def volume_type
+      "ManageIQ::Providers::Amazon::BlockStorageManager::CloudVolume"
+    end
+
+    def volume_snapshot_type
+      "ManageIQ::Providers::Amazon::BlockStorageManager::CloudVolumeSnapshot"
+    end
+  end
+
+
```
Ladas: extra blank lines

app/models/manageiq/providers/amazon/block_storage_manager/refresh_parser_inventory_object.rb
```diff
@@ -0,0 +1,64 @@
...
+    $aws_log.info("#{log_header}...}")
+    get_volumes
+    get_snapshots
+    $aws_log.info("#{log_header}...Complete")
+
+    inventory_collections
+  end
+
+  private
```
Ladas: blank line before and after private

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -54,6 +72,14 @@ def ensure_managers
     network_manager.provider_region = provider_region
   end
 
+  def ensure_block_storage_managers
+    build_block_storage_manager unless block_storage_manager
+    block_storage_manager.name            = "#{name} Block Storage Manager"
+    block_storage_manager.zone_id         = zone_id
+    block_storage_manager.provider_region = provider_region
+    true
```
Ladas: why we return true?

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -54,6 +72,14 @@ def ensure_managers
     network_manager.provider_region = provider_region
   end
 
+  def ensure_block_storage_managers
+    build_block_storage_manager unless block_storage_manager
```
Ladas: I think we can just put this body to ensure_managers. Or calling this from the body of ensure_managers, then we should extract the current into :ensure_network_manager, then add :ensure_block_storage_manager

Ladas:
@gberginc just few nits, mostly rubocop that should be fixed. After that I am ready to merge, so we can move forward. :-)

app/models/manageiq/providers/amazon/cloud_manager.rb
gberginc: @Ladas I've merged the two into `ensure_managers` as there is currently no need to have them separated. I also fixed the other rubocop issues.

gberginc:
@Ladas I amended the commit, so the following comment is not visible. Just reposting to make sure it doesn't get lost.

I've merged the two into ensure_managers as there is currently no need to have them separated. I also fixed the other rubocop issues - I think the remaining issues can be left as they are following existing code.

Ladas:
@gberginc awesome :-) Thank you for a great contribution. 

Yeah, the rest of the robocop warnings are on the edge of a correctness. :-)

sergio-ocon:
👍

roliveri:
@gberginc @Ladas @Fryguy  Actually, this shouldn't have been merged yet. We've been naming managers based on concrete names instead of type. This enables providers to support different/multiple implementations of the same type of manager. So, I would suggest a name change.


From:
```ruby
ManageIQ::Providers::Amazon::BlockStorageManager < ManageIQ::Providers::StorageManager
```
To:
```Ruby
ManageIQ::Providers::Amazon:: StorageManager::EBS < ManageIQ::Providers::StorageManager
```

gberginc:
@roliveri do you think we could proceed with the pending PRs on the manageiq repo and then introduce this change now that it's merged?

roliveri:
@gberginc Is there a benefit in doing that? It seems that 2 of the pending PRs rely on the same naming scheme:

https://github.com/ManageIQ/manageiq/pull/13458
https://github.com/ManageIQ/manageiq/pull/13457

So I would make the same comment on those. If I merge those, you'll only have to submit new PRs updating those as well.

gberginc:
@roliveri the only consideration was that `ManageIQ/manageiq` was failing tests because of this merge. I though it would be faster to make the existing PRs merged and then do a followup patch that would fix the names.

roliveri:
@gberginc Ah, i see. It looks like https://github.com/ManageIQ/manageiq/pull/13458 has been merged, and doing so has fixed the test failures.

So, https://github.com/ManageIQ/manageiq/pull/13457 remains. I can hold off merging that until the names are fixed, or I can merge it, but then I'd like for you to open a Github issue for the name change - just to make sure it doesn't get overlooked.

let me know which way you'd like to go.

gberginc:
@roliveri I already started on renaming the class. I therefore suggest you wait with the worker as it is not required for functional test.

PR #105, opened 2017-01-13
Ems event collection refresh

Ladas:
A targeted refresh based on AwsConfig events. Creating an EmsEventCollection that is used for grouping EmsEvent targets and refreshing them in one graph Refresh. This way, we are able to save all needed cross-links, in the case many AWS object are created at once (e.g. Vm, Interface, EIP, SecurityGroup, etc.)

This targeted refresh doesn't do any API call, it uses Event Payload data to refresh MIQ DB.

Implements:
https://www.pivotaltracker.com/story/show/137733251


app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
```diff
@@ -26,6 +26,14 @@ def parse_targeted_inventory(ems, _target, inventory)
     hashes, = Benchmark.realtime_block(:parse_inventory) do
       if refresher_options.try(:[], :inventory_object_refresh)
         ManageIQ::Providers::Amazon::CloudManager::RefreshParserInventoryObject.new(inventory).populate_inventory_collections
+
+        if inventory.kind_of?(ManageIQ::Providers::Amazon::Inventory::Targets::EmsEventCollection)
+          # For EmsEventCollection, we want to parse everything, regardless whether it belongs to Cloud or Network
+          # Manager
+          ManageIQ::Providers::Amazon::NetworkManager::RefreshParserInventoryObject.new(inventory).populate_inventory_collections
```
gberginc: Here is where we'll have to pipulate other refresh parsers as well?

app/models/manageiq/providers/amazon/inventory/ems_event_collection.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def all_related_ems_events(ems)
+    # We want all EmsEvents around the time of collected EmsEvents, and to the future since AWS send all related events
+    # at once
+    EmsEvent.where(:ems_id => ems.id).where(:timestamp => datetime_min..DateTime::Infinity.new).order("timestamp DESC")
+  end
+
+  def datetime_min
+    targets.collect(&:timestamp).min - 3.minutes
```
gberginc: I suggest adding a comment why 3 mins are used here.

app/models/manageiq/providers/amazon/inventory/ems_event_collection.rb
```diff
@@ -0,0 +1,25 @@
...
+  attr_accessor :targets
+
+  def initialize(targets)
+    @targets = targets
+  end
+
+  def all_related_ems_events(ems)
+    # We want all EmsEvents around the time of collected EmsEvents, and to the future since AWS send all related events
+    # at once
+    EmsEvent.where(:ems_id => ems.id).where(:timestamp => datetime_min..DateTime::Infinity.new).order("timestamp DESC")
```
gberginc: Is there any nicer way for checking `timestamp >= datetime_min`?

app/models/manageiq/providers/amazon/inventory/targets.rb
```diff
@@ -23,7 +27,7 @@ def add_inventory_collections(inventory_collections, inventory_collections_data
 
   def add_remaining_inventory_collections(inventory_collections_data)
     # Get names of all inventory collections defined in InventoryCollectionDefaultInitData
-    all_inventory_collections = ManageIQ::Providers::Amazon::Inventory::InventoryCollectionDefaultInitData.
+    all_inventory_collections     = ManageIQ::Providers::Amazon::Inventory::InventoryCollectionDefaultInitData.
```
gberginc: Are the spaces intentional?

app/models/manageiq/providers/amazon/inventory/targets/ems_event_collection.rb
```diff
@@ -0,0 +1,186 @@
...
+      unique_checker_id = "#{resource_type}__#{resource_id}"
+
+      next if resource_type.blank? || resource_id.blank?
+      # We take only the last change and ignore the rest. Events are ordered so the last ones come first.
+      next if uniq_checker[unique_checker_id]
+
+      uniq_checker[unique_checker_id] = true
+
+      event_payload   = event_payload(ems_event)
+      collection_name = case resource_type
```
aiperon: Should we also expect AWS::S3 events here?

gberginc: Yes, we are going to add storage events, but I was planning to add them after we have at least the basic logic working.

app/models/manageiq/providers/amazon/cloud_manager/refresh_parser_inventory_object.rb
```diff
@@ -143,6 +147,7 @@ def parse_flavor(flavor)
 
     {
       :type                     => ManageIQ::Providers::Amazon::CloudManager::Flavor.name,
+      :ext_management_system    => ems,
```
Ladas: The ems relation is usually filled automatically, when it's saved like ems.association.build, but for a targeted refresh, we don't build it like this, so we need to provide it.

app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
Ladas: Events are going only through a cloud manager now, so no need.

app/models/manageiq/providers/amazon/inventory/ems_event_collection.rb
Ladas: yes, using arel_table

app/models/manageiq/providers/amazon/inventory/ems_event_collection.rb
Ladas: it's a first draft condition to find related events, usually the similar events should have a CloudTrail id, but it's not true for some reason, when deploying through stack.

app/models/manageiq/providers/amazon/inventory/targets.rb
Ladas: hm no, this one is a leftover, not caught by auto-align for some reason :-)

app/models/manageiq/providers/amazon/inventory/targets/ems_event_collection.rb
Ladas: yeah, we should, haven't played with those yet

Ladas:
@aiperon @gberginc btw. this whole feature is hidden under a configuration now and it's possible, we will end up not using it :-(

So benefit of this is, that we can refresh our data, just based on the payload of the events we receive, so we do not need to make any API calls.

But a disadvantage of this is that AwsConfig events are slow, it takes like 5-10 minutes for them to arrive. And also not all services are covered by them, I don't get events for the whole CloudFormations, images, etc, so we still need to make a periodical full refresh.

So I will be exploring CloudWatch events next, where we will need to make API requests, since there is not a full payload sent and we will need to batch them, to not hit API limits. But they are very quick in turn, we receive them almost instantly.

gberginc:
With "this whole feature" you mean event collection refresh or the entire inventory that was introduced in #102?

We'll focus on other aspects of storage managers now until we better understand how to use events.

Ladas:
@gberginc just this PR with refresh based on Events payload. Right now, each event will cause a full refresh of a CloudManager (including the storage events) and after that NetworkManager refresh is queued, so StorageManagers should be too (here: https://github.com/Ladas/manageiq-providers-amazon/blob/81fcc78bf015e6fbb706e9afe069afa4d6331769/app/models/manageiq/providers/amazon/cloud_manager/refresher.rb#L69-L69)

So yeah, ignore events for now, as by default, storage events are already caught.

durandom:
Does this PR depend on https://github.com/ManageIQ/manageiq/pull/13408 ?
And could you describe what this PR does in the description? It's massive!

app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
Ladas: Ah, yeah, sorry, I answered to something else it seems. Yes. In here, it should parse all data, if this was an event based refresh.

durandom:
While I think it's great to see Inventory updates by the payload of an event, I'm still not sure if using the _normal_ refresh code path is the right direction. As mentioned [here](https://github.com/ManageIQ/manageiq/pull/13408#pullrequestreview-16821115) in the corresponding [core PR](https://github.com/ManageIQ/manageiq/pull/13408).

I think there are basically two scenarios:

1. Use eventpayload to update inventory with the stuff in the payload
2. Use eventpayload to augment / support an update which requests more information from the provider

While 2. might re-use some of the refresh part, 1. probably will not.

I'd rather see something like `Event` -> `Automate` -> `InventoryUpdate`

Also this should be designed carefully as it will be the blueprint (as in copy - paste) for other providers.

@Ladas @blomquisg @agrare wdyt?

Ladas:
I can't do Event -> Automate -> InventoryUpdate, since I want to batch events. The Event -> Automate is there now for determining ** this events should be involved in refresh, send it there as a target**

With this way, we add a lot of targets to the refresh queue, lets say there can be 1000EmsEvent, 500Vm, 2Ems. Now in refresher we say, it's time for a refresh and it needs to decide what to do with all those targets.

So now based on configurations, we should have a new factories, that are able to process targets and output InventoryCollections and Data that will be used in a refresh. The configurations can be.

1. Always do a full ems refresh
2. Use data from the events to do a refresh without API calls
3. Use the data from a refresh to get ems_refs, based on those we will do batched targeted refreshes (using API calls to get a subset of the entities, mentioned in the events)
4. Combination of ^? 

And ultimately, all the EmsEvents should be sent as an 'update_only node' to each graph refresh, so all the references are filled. (remember that we need a post refresh hook now, to tie an EmsEvent to a created entity, since the entity exists in our DB after the refresh and parser tries to tie them before the refresh. We are missing this in most of the providers.)

So what I think we should have, is a new class (Refresher::TargetProcessor) that is able to look at what was sent for refresh and it will determine what should be refresh based on configuration. So the output should be what InventoryCollections we will be saving (when we create IC, we also say how we will save it). And how do we fill those ICs with data (I am experimenting with 1 parser code, that have changed input data collections)

Ladas:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-amazon/pull/115

The refresh based on event's payload can be used in another cloud providers, but it doesn't work well for AWS.

PR #108, opened 2017-01-18
dont run bundle twice on travis

durandom:
without an install script travis defaults to bundle install
because there is a Gemfile in the project root.
This effectively runs bundle twice, because we also run it from
bin/setup


durandom:
cc @bdunne

Ladas:
:+1:

PR #111, opened 2017-01-23
Set executable permission to bin/update script

miha-plesko:
There are three scripts in bin/ folder:

bin/
   |- rails
   |- update
   |- setup

with only bin/update file lacking execute permission. Fixed.


miha-plesko:
@miq-bot add_label bug

gberginc:
Do you need to call the update script from this repo? I always run`bin/update` from the main to update everything.

miha-plesko:
I was following [this](http://www.manageiq.org/docs/guides/developer_setup) setup guide. At the bottom it says:
```
cd plugins/manageiq-providers-amazon
ln -s ~/src/manageiq spec/manageiq
bin/update
# == SKIPPING update of spec/manageiq because its symlinked ==
```
I'm not sure why, but it feels like the script may be useful. My guessing was that since the script is here, it should at least have appropriate permissions.

gberginc:
I see. This has changed recently and I still seem to use the old way :-)

https://github.com/ManageIQ/guides/commit/450bc969cd99b623214335e0dafc5f2b59bac570

miha-plesko:
Yes, it's amazing, the speed of development on MIQ😄
Would you perhaps know who should I assign this PR to?

gberginc:
Azure provider has `+x` on `bin/update`, while UI classic doesn't even have `bin/update`.

Perhaps @durandom can shed some light due to the change in the guides repo?

durandom:
development for pluggable providers is done in a checkout of the provider, so the guides are uptodate. You can still bin/update in you miq checkout and it will update the providers, but only as a gem and not your checkout. Makes sense? It's a bit tricky, but in the end it's just gem dependencies.

Now a rails plugin needs a rails dummy app to run tests. That dummy is just our miq core codebase. Hence the linking or bin/setup to clone it.

The fs layout of providers are streamlined by a rails provider generator in miq core, `rails g provider amazon` 

Looks like this `+x` change did not make it here.

The UI is a totally different story 😄

durandom:
@miha-plesko thanks

PR #112, opened 2017-01-23
Add links between cloud volumes and base snapshots

gberginc:
This patch adds the ability to identify base snapshot, i.e. the snapshot
that was used to create a specific volume. Both legacy parser and
inventory object parser are updated, along with the spec. The latter
adds tests for specific volume and snapshot to make sure proper values
are identified.

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
@@ -15,8 +15,9 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParser
     $aws_log.info("#{log_header}...")
     get_volumes
     get_snapshots
-
     $aws_log.info("#{log_header}...Complete")
+
+    link_storage_associations
 
     @data
   end
@@ -41,7 +42,8 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParser
       :status        => volume.state,
       :creation_time => volume.create_time,
       :volume_type   => volume.volume_type,
-      :size          => volume.size.to_i.gigabytes
+      :size          => volume.size.to_i.gigabytes,
+      :snapshot_uid  => volume.snapshot_id
     }
 
     return uid, new_result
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
```diff
@@ -57,13 +59,21 @@ def parse_snapshot(snap)
       :status        => snap.state,
       :creation_time => snap.start_time,
       :description   => snap.description,
-      :size          => snap.volume_size,
+      :size          => snap.volume_size.to_i.gigabytes,
```
gberginc: While adding new specs, I realised I stored GB and not bytes for snapshots.

spec/models/manageiq/providers/amazon/storage_manager/ebs/stubbed_refresher_spec.rb
```diff
@@ -53,15 +53,15 @@
         it "2 refreshes, first creates all entities, second deletes all entitites" do
           @data_scaling = 1
           2.times do
-            refresh_spec
+            refresh_spec(false)
```
gberginc: in L64 I introduce additional parameter that controls whether tests for specific entities are needed. Since this test deletes all entities, there is none to actually make a specific test in this case. All other cases should pass specific tests as well.

gberginc:
With this patch, the UI properly populates the `Base Snapshot` field, as indicated in the screenshot.
screen shot 2017-01-23 at 22 23 57

spec/models/manageiq/providers/amazon/storage_manager/ebs/stubbed_refresher_spec.rb
miha-plesko: I'd go with two functions instead of this boolean:
* `refresh_spec_table_counts`
* `refresh_spec_specific_objects` (should assert both table counts and specific snapshot/volume)

Or at least rename the parameter `with_specific_tests` into `assert_counts_only`. I was wondering what specific tests are you referring to and could only understand when I read the comment here...

spec/models/manageiq/providers/amazon/storage_manager/ebs/stubbed_refresher_spec.rb
```diff
@@ -183,4 +187,35 @@ def assert_ems
...
+      :status      => "in-use",
+      :volume_type => "standard",
+      :size        => 1.gigabyte
+    )
+
+    expect(@volume.ext_management_system).to eq(@ems.ebs_storage_manager)
+    expect(@volume.base_snapshot).to eq(@snapshot)
+  end
+
+  def assert_specific_snapshot
```
miha-plesko: Move this above `assert_specific_volume`, please. It makes more sense to define functions in same order as they should be executed.

gberginc: Changed the order.

spec/models/manageiq/providers/amazon/storage_manager/ebs/stubbed_refresher_spec.rb
gberginc: Changes to `refresh_spec_counts` and `refresh_spec_full`.

Ladas:
@gberginc looks great :+1:

PR #114, opened 2017-01-25
Amazon event parser for CloudWatch events

Ladas:
Amazon event parser for CloudWatch events. The parser is able
to recognize CloudWatch ec2 events and CloudWatch API events
collected via CloudTrail.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
@@ -159,12 +159,29 @@ class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Stream
   # @param [Aws::SQS::Types::Message] message
   def parse_event(message)
     event = JSON.parse(JSON.parse(message.body)['Message'])
-    $log.info("#{log_header} Found SNS Message with message type #{event["messageType"]}")
-    return unless event["messageType"] == "ConfigurationItemChangeNotification"
+
+    if event["messageType"] == "ConfigurationItemChangeNotification"
+      # Aws Config Events
+      event["eventType"]   = parse_event_type(event)
+      event[:event_source] = :config
+    elsif event.fetch_path("detail", "eventType") == "AwsApiCall"
+      # CloudWatch with CloudTrail for API requests Events
+      event["eventType"]   = event.fetch_path("detail", "eventName")
+      event[:event_source] = :cloud_watch_api
+    elsif event["detail-type"] == "EC2 Instance State-change Notification"
+      # CloudWatch EC2 Events
+      state                = "_#{event.fetch_path("detail", "state")}" if event.fetch_path("detail", "state")
+      event["eventType"]   = "#{event["detail-type"].gsub(" ", "_").gsub("-", "_")}#{state}"
+      event[:event_source] = :cloud_watch_ec2
+    else
+      # Not recognized event, ignoring...
+      return
+    end
+
+    $log.info("#{log_header} Found SNS Message with message type #{event["eventType"]} coming from #{event[:event_source]}")
 
     event["messageId"] = message.message_id
-    event["eventType"] = parse_event_type(event)
-    $log.info("#{log_header} Parsed event from SNS Message #{event["eventType"]}")
+    $log.info("#{log_header} Parsed event from SNS Message #{event["eventType"]} coming from #{event[:event_source]}")
     event
   rescue JSON::ParserError => err
     $log.error("#{log_header} JSON::ParserError parsing '#{message.body}' - #{err.message}")
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
@@ -163,16 +163,16 @@ class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Stream
     if event["messageType"] == "ConfigurationItemChangeNotification"
       # Aws Config Events
       event["eventType"]   = parse_event_type(event)
-      event[:event_source] = :config
+      event["event_source"] = :config
     elsif event.fetch_path("detail", "eventType") == "AwsApiCall"
       # CloudWatch with CloudTrail for API requests Events
       event["eventType"]   = event.fetch_path("detail", "eventName")
-      event[:event_source] = :cloud_watch_api
+      event["event_source"] = :cloud_watch_api
     elsif event["detail-type"] == "EC2 Instance State-change Notification"
       # CloudWatch EC2 Events
       state                = "_#{event.fetch_path("detail", "state")}" if event.fetch_path("detail", "state")
       event["eventType"]   = "#{event["detail-type"].gsub(" ", "_").gsub("-", "_")}#{state}"
-      event[:event_source] = :cloud_watch_ec2
+      event["event_source"] = :cloud_watch_ec2
     else
       # Not recognized event, ignoring...
       return
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
@@ -163,16 +163,16 @@ class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Stream
 
     if event["messageType"] == "ConfigurationItemChangeNotification"
       # Aws Config Events
-      event["eventType"]   = parse_event_type(event)
+      event["eventType"]    = parse_event_type(event)
       event["event_source"] = :config
     elsif event.fetch_path("detail", "eventType") == "AwsApiCall"
       # CloudWatch with CloudTrail for API requests Events
-      event["eventType"]   = event.fetch_path("detail", "eventName")
+      event["eventType"]    = event.fetch_path("detail", "eventName")
       event["event_source"] = :cloud_watch_api
     elsif event["detail-type"] == "EC2 Instance State-change Notification"
       # CloudWatch EC2 Events
-      state                = "_#{event.fetch_path("detail", "state")}" if event.fetch_path("detail", "state")
-      event["eventType"]   = "#{event["detail-type"].gsub(" ", "_").gsub("-", "_")}#{state}"
+      state                 = "_#{event.fetch_path("detail", "state")}" if event.fetch_path("detail", "state")
+      event["eventType"]    = "#{event["detail-type"].tr(" ", "_").tr("-", "_")}#{state}"
       event["event_source"] = :cloud_watch_ec2
     else
       # Not recognized event, ignoring...
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
@@ -1,10 +1,11 @@
 #
-# Uses the AWS Config service to monitor for events.
+# Uses the AWS Config or CloudWatch service to monitor for events.
 #
-# AWS Config events are collected in an SNS Topic.  Each appliance uses a unique
+# AWS Config or CloudWatch events are collected in an SNS Topic.  Each appliance uses a unique
 # SQS queue subscribed to the AWS Config topic.  If the appliance-specific queue
 # doesn't exist, this event monitor will create the queue and subscribe the
 # queue to the AWS Config topic.
+
 #
 class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Stream
   class ProviderUnreachable < ManageIQ::Providers::BaseManager::EventCatcher::Runner::TemporaryFailure
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
@@ -1,48 +1,41 @@
 module ManageIQ::Providers::Amazon::CloudManager::EventParser
-  # these have their own EmsEvent classes in automate
-  INSTANCE_EVENTS = %w(
-    AWS_EC2_Instance_CREATE
-    AWS_EC2_Instance_UPDATE
-    AWS_EC2_Instance_DELETE
-    AWS_EC2_Instance_running
-    AWS_EC2_Instance_shutting-down
-    AWS_EC2_Instance_stopped
-  ).to_set.freeze
-
-  def self.map_event_type(aws_event_type)
-    if INSTANCE_EVENTS.include?(aws_event_type)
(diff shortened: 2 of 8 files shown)


Ladas:
@blomquisg @bronaghs changes needed to store events coming from CloudWatch.

@gberginc I'll be adding CloudWatch events for EBS and EBS snaphots. EBS is now covered only in the API events now. Although EBS CloudWatch events do not work for me, I see no events emitted when creating/attaching/detaching volumes. So it's probably a bug, I'll send message to AWS support

@durandom as part of this I am also reverting https://github.com/ManageIQ/manageiq-providers-amazon/pull/93 . We might want to revert it separately, then rebase this? That is, if we decide to go with https://github.com/ManageIQ/manageiq-content/pull/26

gberginc:
@Ladas Great, thanks!

I also checked EBS events sent to SNS but didn't receive any change related to the changes I did in EBS. Since there is no "deleteVolume" event in the "EBS Volume Notification" type, API monitoring will probably have to be used for this kind of events.

Interestingly, though, "EBS Snapshot Notification" works (although it also doesn't trigger upon removal):

```
  "Message" : "{\"version\":\"0\",\"id\":\"47147f12-1b86-4de8-8afa-929128e7739a\",\"detail-type\":\"EBS Snapshot Notification\",\"source\":\"aws.ec2\",\"account\":\"692986702767\",\"time\":\"2017-01-25T09:58:27Z\",\"region\":\"us-west-2\",\"resources\":[\"arn:aws:ec2::us-west-2:snapshot/snap-0ff3038f4be81510f\"],\"detail\":{\"event\":\"createSnapshot\",\"result\":\"succeeded\",\"cause\":\"\",\"request-id\":\"\",\"snapshot_id\":\"arn:aws:ec2::us-west-2:snapshot/snap-0ff3038f4be81510f\",\"source\":\"arn:aws:ec2::us-west-2:volume/vol-06f5539b700c57b0f\",\"startTime\":\"2017-01-25T09:56:48.000Z\",\"endTime\":\"2017-01-25T09:58:26.803Z\"}}",
```

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -1,48 +1,41 @@
...
-      "#{event['eventType']}: #{changed_properties}"
-    else
-      "#{event['eventType']}"
-    end
+  def self.parse_cloud_watch_api_event!(event, event_hash)
+    event_hash[:message]                   = "request: #{event.fetch_path("detail", "requestParameters")}, "\
+                                             "response: #{event.fetch_path("detail", "responseElements")}"
+    event_hash[:timestamp]                 = event["time"]
+    # TODO(lsmola) so Event can be tied to more Vms, we will need to change the modeling
+    event_hash[:vm_ems_ref]                = event.fetch_path("detail", "responseElements", "instancesSet", "items").
```
gberginc: Should there be a similar setter for EBS `volume_ems_ref` or is it sufficient to just store the event?

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -1,48 +1,41 @@
...
       :source     => "AMAZON",
-      :message    => format_message(event),
-      :timestamp  => event["notificationCreationTime"],
       :full_data  => event,
       :ems_id     => ems_id
     }
 
-    event_hash[:vm_ems_ref]                = parse_vm_ref(event)
-    event_hash[:availability_zone_ems_ref] = parse_availability_zone_ref(event)
+    send("parse_#{event[:event_source]}_event!", event, event_hash)
```
gberginc: Should there be a check for supported `event_source`s?

blomquisg: Heh, probably, since that's the test that's failing right now :smile:

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
```diff
@@ -159,12 +159,29 @@ def queue_url_to_arn(queue_url)
   # @param [Aws::SQS::Types::Message] message
   def parse_event(message)
     event = JSON.parse(JSON.parse(message.body)['Message'])
-    $log.info("#{log_header} Found SNS Message with message type #{event["messageType"]}")
-    return unless event["messageType"] == "ConfigurationItemChangeNotification"
+
+    if event["messageType"] == "ConfigurationItemChangeNotification"
+      # Aws Config Events
+      event["eventType"]   = parse_event_type(event)
+      event[:event_source] = :config
```
gberginc: @bronaghs my understanding is that we are getting AWS Config events along side the others that are used in the `elsif` parts below.

The check for `AwsApiCall` (CloudTrail) is because currently EBS events don't work (even though you configure CW to publish EBS-related events, they are never delivered over SNS; @Ladas mentioned he needs to check with AWS support). CloudTrail supposedly pushes all AWS API calls, including those that occur when you are creating, deleting, ... volumes/snapshots.

Last `elsif` checks for `EC2 Instance State-change Notification` which is a CW event, specific to EC2 instances.

durandom:
> as part of this I am also reverting #93 . We might want to revert it separately, then rebase this? That is, if we decide to go with ManageIQ/manageiq-content#26

So, this PR adds all event types again? This means you have to add all known events to settings.yml?

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
Ladas: @gberginc I haven't decided here yet, but we have like 15 more entities we want to connect there. Maybe we could use polymorphic relation. There is already a polymorphic relation :target, we could probably use.

Ladas:
@durandom yeah, the settings.yml is for filtering what we want to show in timelines, which is good to filter out I think. Also, each of the event should go to a specific section.

But the refresh should still work on all events, with a possibility of redefining or changing the behavior in automate.

bronaghs:
LGTM.
I will merge once the tests are green and @durandom has given it a 👍

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
```diff
@@ -159,12 +160,29 @@ def queue_url_to_arn(queue_url)
...
+
+    if event["messageType"] == "ConfigurationItemChangeNotification"
+      # Aws Config Events
+      event["eventType"]    = parse_event_type(event)
+      event["event_source"] = :config
+    elsif event.fetch_path("detail", "eventType") == "AwsApiCall"
+      # CloudWatch with CloudTrail for API requests Events
+      event["eventType"]    = event.fetch_path("detail", "eventName")
+      event["event_source"] = :cloud_watch_api
+    elsif event["detail-type"] == "EC2 Instance State-change Notification"
```
durandom: I wonder why there are so many conventions

* camelcase: `eventType` 
* underscore: `event_source` 
* dash: `detail-type`

So, it looks the `detail-type` is coming from CloudWatch, `messageType` is coming from AWS Config and `eventType` and `event_source` is something we set.

How about 
* `eventType` => `miq_aws_event_type`
* `event_source` => `miq_aws_event_source`

This makes it at least clearer, that we add the underscore keys, and we prepend them with something not already set and with `miq_aws_`

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -1,48 +1,45 @@
...
       :ems_id     => ems_id
     }
 
-    event_hash[:vm_ems_ref]                = parse_vm_ref(event)
-    event_hash[:availability_zone_ems_ref] = parse_availability_zone_ref(event)
+    unless %i(config cloud_watch_api cloud_watch_ec2).include?(event["event_source"])
+      raise "Unsupported event source #{event["event_source"]}"
+    end
+
+    send("parse_#{event["event_source"]}_event!", event, event_hash)
```
durandom: While this looks clever, I think something like this is a perfect case for a switch statement:

```ruby
case event["event_source"]
when 'config'
  parse_config_event! event, event_hash
else 
  raise "Unsupported event source #{event["event_source"]}"
end
```

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -1,48 +1,45 @@
...
-  ).to_set.freeze
-
-  def self.map_event_type(aws_event_type)
-    if INSTANCE_EVENTS.include?(aws_event_type)
-      aws_event_type
-    else
-      'ConfigurationItemChangeNotification'
-    end
+  def self.parse_config_event!(event, event_hash)
+    event_hash[:message]                   = event["configurationItemDiff"]
```
durandom: How about keeping / restricting the number of `eventTypes` at least for aws config?
See, what I don't like is, that we open up the amount of events that are possibly coming from aws.
Now, it's impossible to configure all events in `settings.yml` - and I dont know how you can query the DB via the UI for all events that __are__ in the DB.

That was my rationale in https://github.com/ManageIQ/manageiq-providers-amazon/pull/93

@blomquisg could you please weigh in?

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
```diff
@@ -1,10 +1,11 @@
 #
-# Uses the AWS Config service to monitor for events.
+# Uses the AWS Config or CloudWatch service to monitor for events.
```
Ladas: @bronaghs is actually working on automated setup. But yeah, you need to configure it, then both AWSConfig and ClouWatch can send event into the Topic we have there. But in reality, that costs money, you are fine with one of them. :-)

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
Ladas: so only event_source is what I am adding, to know how the parse the event in parser. Rest is AWS API.

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
Ladas: So in all providers, we store all events that just pass the evenType check. Then you are able to change settings/automate in production, so it shows and does anything special. I think we need to keep this behavior.

In settings.yml, we can define just few interesting events. Settings is used to determine which event should show in which section in timelines. So given to requirement of 'section', we cannot do this automatically. But at any time, you can extend this in customer environment to show events of interest (they are stored just not shown in timelines)

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
Ladas: well, I had it like this, it was too many lines :-)

durandom: but you can better understand the lines, but up to you - that's a matter of style I guess

Ladas: then lets keep it this way, since I changed it from a case to this :-)

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
durandom: So, you are saying, all other providers also just write all events to the db without gatekeeping them?
How can they introspect the events in the DB?

If it's like this, fine. I'd just like to keep it consistent.

Ladas: Yeah, usually only condition is that event needs to have some event_type, so it has a nice name.

Then all are stored in our DB. Using the Automate, you can then react on any of them, same as using settings, you can show any of them in some section in timelines.

There is not really any introspection except for going to the DB and looking at the unique event names.

Ladas: That being said, we could have something like show me more events. But I think the way of listing them in settings is supposed to make it more understandable when showing it.

But that being said, I would like to see :all in timelines and do a smart grouping of the events. So you don't see 1 event but a (30 events of VmCreated) which will be a list to those events.

With a smart grouping we can display large number of events and have a nice overview of what is happening.

Ladas: But grouping like that is a bit hard to do in SQL :-) So maybe when we have the timeseries db, this will be a piece of cake.

durandom: yeah, all you say is true but really out of scope and a huge undertaking 🗻

durandom:
@blomquisg we need you to make a call on wether all events should make it into the db or if we should gatekeep them and funnel unknown events into a common thing like `ConfigChanged`

Plus the documentation part of it is probably taken care of in https://github.com/ManageIQ/manageiq-providers-amazon/issues/126

Ladas:
@blomquisg @durandom we do this in other providers, event should be saved under its event name and we do collect all events that have an 'event name'

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
blomquisg: @durandom right, the only gatekeeping that's done by other providers is the "filtered events" setting to specifically ignore certain event types.  Beyond that, *all* events are captured even if all events are not handled.

That means that we can see things that happened in the `event_streams` table. Then handling the event requires adding a handler, and graphing the event on the timeline requires adding the event to the timeline configuration.

blomquisg:
> we need you to make a call on whether all events should make it into the db or ...

@durandom @Ladas 

All other providers already allow all non-filtered events into the DB as is from the provider.  The only gatekeeping that's done is via the event filtering.

That said, AWS Config might be a unique situation if the multitude of events amounts to hundreds of unique event types because of minute configuration changes being tracked.  If that's the case, I don't mind if we bundle some disparate events into common "event types" for AWS Config.  But, I don't think we should drop events on the floor unless the events are specifically filtered.

Ladas:
@blomquisg @durandom right, we should do the same filtering as e.g. OpenStack has, connected to the settings. If i remember correctly, you were writing the @blomquisg , right?

But the filtering is already on the AWS side, you should not send Events, that are ignored to the MIQ topic, since it costs $. Also, with the autoconfigure, we will subscribe only to Events we are interested in.

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
durandom: ok, thanks for the clarification.

durandom:
ok. sorry for the lengthy review - nice work.

https://github.com/ManageIQ/manageiq-providers-amazon/pull/93 is partially reverted by this

PR #115, opened 2017-01-25
Separate collector from inventory and targeted refresh

Ladas:
Separate collector from inventory, Collector class focus directly on collecting related data from a provider. Introduce targeted refresh.


app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
```diff
@@ -35,6 +44,29 @@ def parse_targeted_inventory(ems, _target, inventory)
     hashes
   end
 
+  def preprocess_targets
```
durandom: For now its ok, but I dont think we should allow this to be overriden

app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
```diff
@@ -35,6 +44,29 @@ def parse_targeted_inventory(ems, _target, inventory)
...
+    @targets_by_ems_id.each do |ems_id, targets|
+      if targets.any? { |t| t.kind_of?(ExtManagementSystem) }
+        ems             = @ems_by_ems_id[ems_id]
+        targets_for_log = targets.map { |t| "#{t.class} [#{t.name}] id [#{t.id}] " }
+        _log.info "Defaulting to full refresh for EMS: [#{ems.name}], id: [#{ems.id}], from targets: #{targets_for_log}" if targets.length > 1
+      end
+
+      # We want all targets of class EmsEvent to be merged into one target, so they can be refreshed together, otherwise
```
durandom: So, this thing about `EmsEvent` being a target. Is https://github.com/ManageIQ/manageiq/pull/13408 needed for this to work?

app/models/manageiq/providers/amazon/inventory/collectors.rb
```diff
@@ -0,0 +1,126 @@
+class ManageIQ::Providers::Amazon::Inventory::Collectors
```
durandom: I dont like this class being a plural. It should be `Collector`

app/models/manageiq/providers/amazon/inventory/collectors.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  def initialize(ems, target)
+    @ems     = ems
+    @target  = target
+    @options = Settings.ems_refresh[ems.class.ems_type]
+
+    initialize_inventory_sources
+  end
+
+  def initialize_inventory_sources
```
durandom: So every collector has all this? Shouldnt it be per Subclass different?

app/models/manageiq/providers/amazon/inventory/targets.rb
```diff
@@ -1,4 +1,9 @@
 class ManageIQ::Providers::Amazon::Inventory::Targets < ManageIQ::Providers::Amazon::Inventory
```
durandom: Same here, why is that a plural class name? And why does it inherit from `Inventory` ?

app/models/manageiq/providers/amazon/inventory/collectors.rb
Ladas: we can do singular, I have there Targets and Collectors

app/models/manageiq/providers/amazon/inventory/collectors.rb
Ladas: no, we need is always defined, so by default all data are []

app/models/manageiq/providers/amazon/inventory/targets.rb
Ladas: The target is really an inventory here

We can revisit that Inventory could have target and collector

durandom:
merging to move forward ontop of this

PR #116, opened 2017-01-25
Parse availability zone of an EBS volume

gberginc:
Refresh parser and inventory object will retrieve the availability zone from the parent manager or inventory and save it into the cloud volume hash.

Depends-On: https://github.com/ManageIQ/manageiq/pull/13654

@miq-bot assign @Ladas

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
@@ -24,16 +24,6 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParser
 
   private
 
-  def parent_manager_fetch_path(collection, ems_ref)
-    @parent_manager_data ||= {}
-    return @parent_manager_data.fetch_path(collection, ems_ref) if @parent_manager_data.has_key_path?(collection,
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
```diff
@@ -22,6 +22,18 @@ def ems_inv_to_hashes
...
+    @parent_manager_data ||= {}
+    return @parent_manager_data.fetch_path(collection, ems_ref) if @parent_manager_data.has_key_path?(collection,
+                                                                                                      ems_ref)
+
+    @parent_manager_data.store_path(collection,
+                                    ems_ref,
+                                    @ems.public_send(collection).try(:where, :ems_ref => ems_ref).try(:first))
+  end
+
   def get_volumes
```
bronaghs: @gberginc - this method already exists in ```ManageIQ::Providers::Amazon::NetworkManager::RefreshParser```
Maybe it should be moved to ```ManageIQ::Providers::Amazon::RefreshHelperMethods```  which is already included in both the NetworkManager parser and this file.

gberginc:
Thanks @bronaghs.

I have verified that both network and EBS storage refresh parsers pass the tests!

PR #120, opened 2017-01-26
Queue EBS storage refresh after cloud inventory is saved

gberginc:
While parsing volumes immediately after new AWS manager is created, EBS
manager failed to get proper links to availability zones, because it was
parsing data in parallel to the cloud manager. This patch queues the EBS
storage manager parse in the same way as the network manager, i.e.
immediately after the cloud inventory is saved, EBS refreshes the data
once more.

@miq-bot assign @Ladas 

@Ladas I assigned you because there is a todo for you and would like to check if this is still valid way to go.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
@@ -40,6 +40,7 @@ class ManageIQ::Providers::Amazon::CloudManager::Refresher < ManageIQ::Providers
   def save_inventory(ems, target, inventory_collections)
     EmsRefresh.save_ems_inventory(ems, inventory_collections)
     EmsRefresh.queue_refresh(ems.network_manager) if target.kind_of?(ManageIQ::Providers::BaseManager)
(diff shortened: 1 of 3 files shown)


Ladas:
@gberginc yes, this will be the default way of refresh. Full refresh is big and slow, but it's very easy to keep consistent data.

Ladas:
@gberginc it will be enough to increase the MiqQueue count in specs, you are adding a refresh to a Queue, so it has risen +1

The VCR update will be needed when you will add the storage refresh to the spec.

gberginc:
@Ladas I've updated the queue counts so it should pass (hopefully; while trying to run the test locally, I received authentication errors).

With respect to VCR update, I created a [WIP PR ](https://github.com/ManageIQ/manageiq-providers-amazon/pull/122) that adds corresponding EBS manager refreshes. Could you record a new cassette and send it to me so that I can update the specific tests?

gberginc:
@miq-bot add_label euwe/no, enhancement

Ladas:
great :+1:, will send you the VCR tomorrow

PR #121, opened 2017-01-26
Parse the name of EBS volume or snapshot

gberginc:
Name is defined as a tag and can be optional. Patch uses the id of
the volume/snapshot if name is not given.

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
@@ -33,7 +33,7 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParserInventoryOb
       :type                  => self.class.volume_type,
       :ext_management_system => ems,
       :ems_ref               => uid,
-      :name                  => uid,
+      :name                  => get_from_tags(volume, :name) || uid,
       :status                => volume['state'],
       :creation_time         => volume['create_time'],
       :volume_type           => volume['volume_type'],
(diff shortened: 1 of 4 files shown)


gberginc:
@miq-bot add_label euwe/no, enhancement

bronaghs:
@miq-bot assign @bronaghs

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
```diff
@@ -59,6 +59,11 @@ def parse_snapshot(snap)
     }
   end
 
+  # Overridden helper methods, we should put them in helper once we get rid of old refresh
+  def get_from_tags(resource, item)
+    (resource['tags'] || []).detect { |tag, _| tag['key'].downcase == item.to_s.downcase }.try(:[], 'value')
+  end
+
```
bronaghs: Once this PR is merged, the ```get_from_tags``` method will be overridden in 3 different RefreshParserInventoryObject subclasses (NetworkManager, CloudManager, StorageManager) - each one identical.
For the record, as the comment on line 62 indicates, the above implementation will replace the current version in the RefreshHelperMethods module when the RefreshParser class becomes obsolete.

gberginc: It hurt very much copying it once more. I was thinking about using a different name. For example, I thought about renaming the existing `get_from_tags` in the helpers module and using this one instead, but just couldn't find a proper name (`get_from_tags_legacy` or `get_from_tags_refresh_parser` didn't sound ok to me).

Any idea?

bronaghs: @gberginc - Haha I completely know what you mean. If the RefreshParser wasn't on the cusp of being deprecated i would like to see it addressed, but refactoring this disrupts a number of classes and I'm not sure it is worth the effort when it will be addressed properly for the G release.

gberginc:
@bronaghs Updated the stubs with an additional tag for the name for both volumes and snapshots. I'll be updating another specs once I get the VCRs for #122.

bronaghs:
@gberginc - great, thanks.

gberginc:
@bronaghs is this safe to merge or should we wait for the VCR as well?

sergio-ocon:
@bronaghs @ladas, is there any other change needed? I see we are dependent on #122, but is that the only thing we need for testing this?

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
Ladas: Yeah it sucks, but it should be an easy cleanup later :-)

bronaghs:
@gberginc - A separate PR using the VCR specs is fine.

PR #122, opened 2017-01-26
[WIP] Update VCR cassettes with EBS parser

gberginc:
This is a temporary PR that is to be used to get the updated VCRs for the EBS storage manager.

@miq-bot add_label wip

### DIFF
--- a/spec/models/manageiq/providers/amazon/cloud_manager/refresher_inventory_object_spec.rb
+++ b/spec/models/manageiq/providers/amazon/cloud_manager/refresher_inventory_object_spec.rb
@@ -28,6 +28,7 @@ describe ManageIQ::Providers::Amazon::CloudManager::Refresher do
 
         allow(Settings.ems_refresh).to receive(:ec2).and_return(settings)
         allow(Settings.ems_refresh).to receive(:ec2_network).and_return(inventory_object_settings)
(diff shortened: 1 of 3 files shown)


gberginc:
@miq-bot assign @Ladas

gberginc:
@miq-bot add_label wip

sergio-ocon:
Hi @Ladas can we merge this? I think that there are two other PR depending on this.

Thanks,
Sergio

gberginc:
@sergio-ocon I didn't plan to merge this one, because it will break the specs.

I marked it WIP because we discussed with @Ladas it would be best if I try to extend the specs for EBS based on his AWS account. I already have some other tests so that I can work on the provider. Hopefully, I'll then just be able to migrate those specs.

Ladas:
@gberginc also, I am thinking that I will prepare a set of CloudFormation templates, that would deploy all kinds of data combinations we would need to test. Hopefully packed into one big template. With a little script around that, it should be a nice way to turn any AWS account into a VCR env. Also will be easier to get rid of the data after.

Maybe if you would create CloudFormation template for Vms with Volumes, with the combinations you need to test. I could just deploy that in the VCR AWS Account and we would be testing the same data and crosslinks?

Ladas:
:+1: right, I've already sent these

PR #124, opened 2017-01-30
Changelog 53

mfeifer:
Update for Sprint 53

[skip ci]


bronaghs:
@mfeifer - 
I think we could include: https://github.com/ManageIQ/manageiq-providers-amazon/pull/116 and https://github.com/ManageIQ/manageiq-providers-amazon/pull/109

PR #125, opened 2017-01-30
Link volumes to corresponding VM disks

gberginc:
This patch currently works only for the legacy refresh parser. It will look for the instance (VM) referenced in the attachment section of the volume data and create a backing link between the volume and instance's disk. Linking is done similar to the OpenStack (Cinder) provider.

Inventory object doesn't work yet as I can't figure out how to properly get the hardware and disks from the corresponding `:vm` inventory object.

Specs will also fail because I am using local VCR with our own account. They are nevertheless provided for inspection.

@miq-bot add_label wip, enhancement
@miq-bot assign @Ladas 

Depends on: #122. The spec will be updated with the data from the VCR received from this PR.

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
@@ -49,6 +49,8 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParser
       :availability_zone => parent_manager_fetch_path(:availability_zones, volume.availability_zone),
     }
 
+    link_volume_to_disk(new_result, volume.attachments)
+
     return uid, new_result
   end
 
@@ -75,6 +77,57 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParser
       base_snapshot = @data_index.fetch_path(:cloud_volume_snapshots, base_snapshot_uid)
       cv[:base_snapshot] = base_snapshot unless base_snapshot.nil?
     end if @data[:cloud_volumes]
+  end
+
+  def link_volume_to_disk(volume_hash, attachments)
+    uid = volume_hash[:ems_ref]
+
+    attachments.each do |a|
+      if a['device'].blank?
+        _log.warn "#{log_header}: Volume: #{uid}, is missing a mountpoint, skipping the volume processing"
+        _log.warn "#{log_header}:   EMS: #{@ems.name}, Instance: #{a['instance_id']}"
+        next
+      end
+
+      dev = File.basename(a['device'])
+
+      vm = @ems.parent_manager.vms.find_by(:ems_ref => a['instance_id'])
+      unless vm
+        _log.warn "VM referenced by attachment (#{a['instance_id']} not found."
+        next
+      end
+
+      hardware = vm.hardware
+      disks = hardware.disks
+      unless disks
+        _log.warn "#{log_header}: Volume: #{uid}, attached to instance not visible in the scope of this EMS"
+        _log.warn "#{log_header}:   EMS: #{@ems.name}, Instance: #{a['instance_id']}"
+        next
+      end
+
+      disk_hash = {
+        :size           => volume_hash[:size],
+        :backing_volume => volume_hash
+      }
+
+      if (disk = disks.detect { |d| d.location == dev })
+        # Disk exists: save id.
+        disk_hash[:id] = disk.id
+      else
+        # New disk.
+        disk_hash[:hardware_id]     = hardware.id
+        disk_hash[:device_name]     = dev
+        disk_hash[:device_type]     = "disk"
+        disk_hash[:controller_type] = "EBS Volume"
+        disk_hash[:location]        = dev
+      end
+
+      backing_links << disk_hash
+    end
+  end
+
+  def backing_links
+    @data[:backing_links] ||= []
   end
 
   class << self
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
@@ -29,7 +29,7 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParserInventoryOb
   def parse_volume(volume)
     uid = volume['volume_id']
 
-    {
+    volume_hash = {
       :type                  => self.class.volume_type,
       :ext_management_system => ems,
       :ems_ref               => uid,
@@ -41,6 +41,10 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::RefreshParserInventoryOb
       :base_snapshot         => inventory_collections[:cloud_volume_snapshots].lazy_find(volume['snapshot_id']),
       :availability_zone     => inventory_collections[:availability_zones].lazy_find(volume['availability_zone'])
     }
+
+    link_volume_to_disk(volume_hash, volume['attachments'])
+
+    volume_hash
   end
 
   def parse_snapshot(snap)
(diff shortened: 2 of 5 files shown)


app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
```diff
@@ -41,6 +41,10 @@ def parse_volume(volume)
       :base_snapshot         => inventory_collections[:cloud_volume_snapshots].lazy_find(volume['snapshot_id']),
       :availability_zone     => inventory_collections[:availability_zones].lazy_find(volume['availability_zone'])
     }
+
+    link_volume_to_disk(volume_hash, volume['attachments'])
```
gberginc: I am not sure whether the linking should happen here or at some other place (for example, should I only store the attachment here and implement linking at some other place?)

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
```diff
@@ -59,6 +63,53 @@ def parse_snapshot(snap)
...
+
+      dev = File.basename(a['device'])
+
+      vm = inventory_collections[:vms].find(a['instance_id']).data
+      unless vm
+        _log.warn "VM referenced by attachment (#{a['instance_id']} not found."
+        next
+      end
+
+      hardware = vm.hardware
```
gberginc: I get something for the VM, but definitely not the correct thing. Not sure how to get there. You can see in the `refresh_parser` what's needed: https://github.com/ManageIQ/manageiq-providers-amazon/pull/125/files#diff-9ed63d88e4d31be7b6b921a61efec9fdR100

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -860,4 +862,51 @@ def assert_specific_vm_in_other_region
   def assert_relationship_tree
     expect(@ems.descendants_arranged).to match_relationship_tree({})
   end
+
+  def assert_specific_snapshot
```
gberginc: These specs here are just used to demonstrate what I plan to test once everything works. The legacy parser passes both specs.

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
Ladas: right, so looking at the OpenStack and a current implementation, seems like ephemeral disks are filled on the CloudManager side and Volumes are added to that on the StorageManager side.

We could probably set it all from the CloudManager side. Or not sure if the AWS even supports ephemeral disks? So local VM disks, if not, we could move it all to storage manager.

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
```diff
@@ -77,6 +79,57 @@ def link_storage_associations
...
+
+      vm = @ems.parent_manager.vms.detect { |v| v.ems_ref == a['instance_id'] }
+      unless vm
+        _log.warn "VM referenced by attachment (#{a['instance_id']} not found."
+        next
+      end
+
+      hardware = vm.hardware
+      disks = hardware.disks
+      unless disks
```
Ladas: this is strange condition, not sure what this means for OpenStack, but if I have only volumes, disks are always empty

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
Ladas: this would do query per each disk. You can rather load Hardware InventoryCollection, hardware uses the same id so then you can just do

hardware = inventory_collections[:hardwares].find(a['instance_id'])

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
```diff
@@ -59,6 +63,53 @@ def parse_snapshot(snap)
...
+      disks = hardware.disks
+      unless disks
```
Ladas: hm, commented this in some other revision, I have 'disks' always empty, when I have only volumes on a VM

so I think this whole code block should not be needed

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
```diff
@@ -59,6 +63,53 @@ def parse_snapshot(snap)
...
+        :size           => volume_hash[:size],
+        :backing_volume => volume_hash
+      }
+
+      if (disk = disks.detect { |d| d.location == dev })
+          # Disk exists: save id.
+        disk_hash[:id] = disk.id
+      else
+        # New disk.
+        disk_hash[:hardware_id]     = hardware.id
```
Ladas: change this to 

 disk_hash[:hardware] = hardware

since the unique key on disk is defined as [:hardware, :location]

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
```diff
@@ -59,6 +63,53 @@ def parse_snapshot(snap)
...
+        _log.warn "#{log_header}:   EMS: #{@ems.name}, Instance: #{a['instance_id']}"
+        next
+      end
+
+      disk_hash = {
+        :size           => volume_hash[:size],
+        :backing_volume => volume_hash
+      }
+
+      if (disk = disks.detect { |d| d.location == dev })
```
Ladas: I this whole branch is not needed, :id is ignored when saving

app/models/manageiq/providers/amazon/inventory/inventory_collection_default_init_data.rb
```diff
@@ -309,4 +309,12 @@ def cloud_object_store_objects_init_data(extra_attributes = {})
 
     init_data(::CloudObjectStoreObject, attributes, extra_attributes)
   end
+
+  def backing_links_init_data(extra_attributes = {})
```
Ladas: I think we don't need a special definition here, just use disks_init_data

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
gberginc: I think we can keep this now because we are storing disks for attached volumes in the cloud manager as well.

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser_inventory_object.rb
gberginc: Because storage manager is not aware of the ephemeral disks, I kept it in the cloud manager (basically, ephemeral disks are only created because the flavor itself specifies how many local disks there are).

gberginc:
@miq-bot remove_label wip

@Ladas I am removing wip as it looks like we now have both parsers working. Thanks for your great support!

app/models/manageiq/providers/amazon/storage_manager/ebs/refresh_parser.rb
```diff
@@ -77,6 +79,57 @@ def link_storage_associations
...
+    attachments.each do |a|
+      if a['device'].blank?
+        _log.warn "#{log_header}: Volume: #{uid}, is missing a mountpoint, skipping the volume processing"
+        _log.warn "#{log_header}:   EMS: #{@ems.name}, Instance: #{a['instance_id']}"
+        next
+      end
+
+      dev = File.basename(a['device'])
+
+      vm = @ems.parent_manager.vms.detect { |v| v.ems_ref == a['instance_id'] }
```
Ladas: @ems.parent_manager.vms.find_by(:ems_ref => a['instance_id'])

this will do 1 small query to the db, otherwise you are doing fetch me all vms and then O(n) search in Ruby. :-)

Ladas:
@gberginc awesome :+1:

PR #127, opened 2017-02-01
Collect block device mappings in AWS cloud manager

gberginc:
Besides collecting ephemeral storage in the AWS cloud manager refresh
parser, we are adding inventory of attached block devices. This is
necessary because with the introduction of the EBS manager we are adding
disks there as well. If the cloud manager does not collect attached
block devices, those disks collected by the EBS manager will be removed
from the inventory.

The suggested change does not require any change in the core repository.

### DIFF
--- a/app/models/manageiq/providers/amazon/refresh_helper_methods.rb
+++ b/app/models/manageiq/providers/amazon/refresh_helper_methods.rb
@@ -41,6 +41,17 @@ module ManageIQ::Providers::Amazon::RefreshHelperMethods
     super(disks, size, name, location, "amazon")
   end
 
+  def add_block_device_disk(disks, name, location)
+    disk = {
+      :device_name     => name,
+      :device_type     => "disk",
+      :controller_type => "amazon",
+      :location        => location,
+    }
+    disks << disk
+    disk
+  end
+
   # Compose an ems_ref combining some existing keys
   def compose_ems_ref(*keys)
     keys.join('_')
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
@@ -271,6 +271,12 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
       end
     end
 
+    instance.block_device_mappings.each do |blk_map|
+      disks = new_result[:hardware][:disks]
+      device = File.basename(blk_map.device_name)
+      add_block_device_disk(disks, device, device)
+    end
+
     return uid, new_result
   end
 
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser_inventory_object.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser_inventory_object.rb
@@ -135,6 +135,13 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParserInventoryObject < 
       end
     end
 
+    if instance.key?("block_device_mappings")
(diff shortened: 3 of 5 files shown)


app/models/manageiq/providers/amazon/refresh_helper_methods.rb
```diff
@@ -41,6 +41,17 @@ def add_instance_disk(disks, size, name, location)
...
+  def add_block_device_disk(disks, name, location)
+    disk = {
+      :device_name     => name,
+      :device_type     => "disk",
+      :controller_type => "amazon",
+      :location        => location,
+    }
+    disks << disk
+    return disk
+  end
```
bronaghs: return is not needed here.

app/models/manageiq/providers/amazon/refresh_helper_methods.rb
```diff
@@ -41,6 +41,17 @@ def add_instance_disk(disks, size, name, location)
...
   end
 
+  def add_block_device_disk(disks, name, location)
+    disk = {
+      :device_name     => name,
+      :device_type     => "disk",
+      :controller_type => "amazon",
+      :location        => location,
+    }
+    disks << disk
```
bronaghs: @gberginc - is it possible to get any more information about the block devices, maybe size and free space? 
Can device_type be any string? If so, can we distinguish between instance store volume or EBS volume here?

app/models/manageiq/providers/amazon/refresh_helper_methods.rb
```diff
@@ -41,6 +41,17 @@ def add_instance_disk(disks, size, name, location)
...
+    return disk
```
bronaghs: @gberginc - ```return``` isn't needed here.

bronaghs: @gberginc - ```return``` is not needed here.

app/models/manageiq/providers/amazon/refresh_helper_methods.rb
bronaghs: @gberginc - 
- will the block devices always be EBS volumes and not an instance store volume? Can this be reflected in the ```device_type``` attribute assuming any string value will do.
- Is there any more information available here eg size, free space? 
- Should this link back to the EBS volume object?

gberginc: @bronaghs 
* I would also prefer `EBS volume` or something similar, so I'd be happy to use that instead of `amazon`. Honestly, instead of `amazon` for instance storage, I'd also prefer type that would explicitly say it's ephemeral storage, but I don't know currently what this would mean for other parts of the code.
* From the list of mappings I can only get the device (and the volume_id). Other data will be parsed by the EBS storage manager. #125 will populate the remaining data. This PR is basically preparing for the other because currently, cloud_manager deletes all links storage manager creates :(.
* linking is done in #125

app/models/manageiq/providers/amazon/refresh_helper_methods.rb
gberginc: @bronaghs I had a dillema :-). The code is similar to the one in the core repo: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/refresh_parser.rb#L51. The reason for introducing a new function in the AWS provider is because we'd like not to mess something up in the core refresh parser for other managers (discussed with @Ladas).

One thing I will inspect, though, is whether OpenStack suffers from the same problem as AWS does. The code is very similar so I think it could, but I need a test.

Ladas:
@gberginc awesome

@bronaghs yes, the link to volume will be done in the storage manager parser

PR #129, opened 2017-02-02
Moving to use only kwargs for InventoryCollection init

Ladas:
Moving to use only kwargs for InventoryCollection init.
Unifying format, preparing for the abstraction of the
InventoryCollection definitions.

Depends on https://github.com/ManageIQ/manageiq/pull/13739

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/inventory_collection_default_init_data.rb
+++ b/app/models/manageiq/providers/amazon/inventory/inventory_collection_default_init_data.rb
@@ -5,7 +5,7 @@ module ManageIQ::Providers::Amazon::Inventory::InventoryCollectionDefaultInitDat
     }
 
     init_data.merge!(attributes).merge!(extra_attributes)
-    return init_data
+    init_data
   end
 
   def vms_init_data(extra_attributes = {})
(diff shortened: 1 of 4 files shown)


Ladas:
I will send rubocop -a for the whole manager_refresh as a next PR, once @durandom has his enhancements merged

durandom:
makes master green again

chrisarcand:
Minor note (not trying to be pedantic, I was actually genuinely confused for a bit):

These aren't keyword arguments :)

durandom:
@chrisarcand I like how pedantic people always start sentences with *not trying to be...* 😛 
But you have a point, and I also [pointed that out](https://github.com/ManageIQ/manageiq/pull/13743#pullrequestreview-21244757).
But those hashes are later used to fill keyword args https://github.com/ManageIQ/manageiq/pull/13739/files#diff-fb299216c1dbe0f44b0f5c2352b4b6e5R12

If there is still a subtle difference, please continue to be pedantic :)

Ladas:
It acts like kwargs, that is enough for me :-D

chrisarcand:
❤️

Issue #126, opened 2017-01-31
Make subscribing to AWS event services configurable


bronaghs:
Define a blacklist of AWS services we do not wish to capture events from.
Pivotal story:https://www.pivotaltracker.com/story/show/138951121

 A new setting would be added to config/settings.yml, some like this:
    ```:ems_amazon:```
   ``` :disabled_event_services: ["AWS_Config", "CloudTrail", "EC2", "ELB", "ESB" ]```

Events from a service included in ```:disabled_event_services ``` will not be collected on AWS.

ManageIQ currently receives events from the following three AWS services: 

- AWS_Config,
- CloudTrail,
- CloudWatch.

The process for enabling or disabling event collection from each is not the same; AWS_Config and CloudWatch have an ON/OFF switch that can be flipped. CloudWatch is more complicated; it contains a set of rules where a rule applies to a single service eg EC2, ELB, EBS. A rule will be disabled if the service it applies to has been marked as disabled in ```disabled_event_services```. For example:

```:ems_amazon:```
``` :disabled_event_services: ["EC2", "ELB" ]```

 CloudWatch event rules that apply to the EC2 and ELB services will be disabled on AWS preventing events for ELB and EC2 from being emitted, all other CloudWatch event rules will remain enabled.

####    Consideration(s):

- Maybe not directly relevant to this GitHub issue, but during CloudTrail setup, a trail can be applied to all regions, see screenshot below. In ManageIQ, an EMS is tied to a specific region which means each EMS will receive the same CloudTrail events. An excerpt from the [AWS documentation](http://docs.aws.amazon.com/awscloudtrail/latest/userguide/configure-cloudtrail-to-send-notifications.html):

_"When you configure notifications for a trail that applies to all regions, notifications from all regions are sent to the Amazon SNS topic that you specify. If you have one or more region-specific trails, you must create a separate topic for each region and subscribe to each individually."_

![screen shot 2017-02-03 at 1 13 12 pm 2](https://cloud.githubusercontent.com/assets/4379675/22602777/9ff3f082-ea12-11e6-95bd-5f5046b0d474.png)

bronaghs:
@blomquisg -  I bet you are thinking that we can rename the topic to something more generic like MIQ_AWS_Events are something. That will break the existing AWS_Config configuration and will require some logic be implemented to fix it. We should mull over it given that AWS_Config might be deprecated leaving behind an irrelevant topic name.

durandom:
[...]

PR #130, opened 2017-02-02
Creates the AWS SNS topic if one isn't already present.

bronaghs:
Currently, if the AWS_Config SNS topic does not exist on AWS a warning is written to the aws.log file and events are not collected.
This PR will first try to get the AWS_Config topic, and if it does not exist one will be created.

https://github.com/ManageIQ/manageiq-providers-amazon/issues/126
https://www.pivotaltracker.com/story/show/138951121

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
@@ -110,17 +110,21 @@ class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Stream
     end
   end
 
-  # @return [Aws::SNS::Topic] the found topic
-  # @raise [ProviderUnreachable] in case the topic is not found
   def sns_topic
     @ems.with_provider_connection(:service => :SNS) do |sns|
-      sns.topics.detect { |t| t.arn.split(/:/)[-1] == @topic_name }
-    end || begin
-      $aws_log.warn("#{log_header} Unable to find the AWS Config Topic '#{@topic_name}'. " \
-      "Cannot collect Amazon events for AWS Access Key ID #{@ems.authentication_userid}")
-      $aws_log.warn("#{log_header} Contact Amazon to create the AWS Config service and topic for Amazon events.")
-      raise ProviderUnreachable, "Unable to find the AWS Config Topic '#{@topic_name}'"
+      get_topic(sns) || create_topic(sns)
     end
+  end
+
+  def get_topic(sns)
+    sns.topics.detect { |t| t.arn.split(/:/)[-1] == @topic_name }
+  end
+
+  def create_topic(sns)
+    sns.create_topic(:name => @topic_name)
+    $aws_log.info("Created SNS topic #{@topic_name}")
+  rescue Aws::SNS::Errors::ServiceError => err
+    $aws_log.error("Cannot create SNS topic #{@topic_name}, #{err.class.name}, Message=#{err.message}")
   end
 
   # @param [Aws::SNS::Topic] topic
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
```diff
@@ -110,19 +110,23 @@ def sqs_get_queue_url(queue_name)
...
 
+  def get_topic(sns)
+    sns.topics.detect { |t| t.arn.split(/:/)[-1] == @topic_name }
+  end
+
+  def create_topic(sns)
+    sns.create_topic(:name => @topic_name)
+    $aws_log.info("Created SNS topic #{@topic_name}")
+  rescue Aws::SNS::Errors::ServiceError => err
+    $aws_log.error("Cannot create SNS topic #{@topic_name}, #{err.class.name}, Message=#{err.message}")
```
Ladas: I wonder if we should still raise ProviderUnreachable, I see we do that in Ansible and Google, but the exception is not being processed anywhere, @durandom your thoughts ?

Ladas:
@bronaghs this looks great :+1: , just one concern about the exception, pinged @durandom there

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
durandom: +1 for raising, I think it's used to throttle the startup of that worker. But as always room for improvement and propagate that to the UI

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
```diff
@@ -110,19 +110,23 @@ def sqs_get_queue_url(queue_name)
     end
   end
 
-  # @return [Aws::SNS::Topic] the found topic
-  # @raise [ProviderUnreachable] in case the topic is not found
```
durandom: can we keep the yard docs?
I know we rarely use them, but I still like them 😄

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
```diff
@@ -110,19 +110,23 @@ def sqs_get_queue_url(queue_name)
...
+      get_topic(sns) || create_topic(sns)
     end
   end
 
+  def get_topic(sns)
+    sns.topics.detect { |t| t.arn.split(/:/)[-1] == @topic_name }
+  end
+
+  def create_topic(sns)
+    sns.create_topic(:name => @topic_name)
```
durandom: So thats awesome, that you can create the topic. 

But what if aws config is not enabled for that account? Wouldnt that result in a false positive? As in topic found, but nobody sends something to it?

bronaghs: @durandom - Wouldn't that mean the end user does not want event support? In which case the event role would be disabled in ManageIQ.
How would I programmatically determine that CloudWatch and AWS Config are disabled? I looked at the API documentation but didn't see anything to that end.

bronaghs:
@Ladas @durandom - feedback addressed.

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/stream.rb
durandom: to enable aws config, you have to log into the aws console and enable it.

Now what happens, if the use has AWS Config turned off? Or not initially turned it on.
Before he could see in the logs, that he should enable config.
Now he sees nothing, because we just listen to a SQS queue that is subscribed to a SNS topic, that nobody sends to.

Is that right? Or am I missing something?

screen shot 2017-02-07 at 17 51 14

durandom: But yeah, that might mean that he does not want events...

durandom:
@bronaghs thanks. Do you think this requires a change in docs?

http://manageiq.org/docs/guides/providers/amazon_aws_config

bronaghs:
@durandom - thanks for reminding me, yes we need to update the documentation.
@dayleparker - can you update the documentation? If you are not the right person maybe you can direct me to someone who is  - thank you.

PR #133, opened 2017-02-06
[WIP] Decouple S3 Storage Manager from Amazon CloudProvider

miha-plesko:
This commit introduces S3 Storage Manager as a standalone management system so that multiple CloudManagers can share it.

When first Amazon Cloud Manager is created, it creates a new S3 Manager giving it a copy of its own authentication data. Then when another Amazon Cloud Manager is created using same WS_ACCESS_KEY_ID, no new S3 Manager is created - the existing one is shared.

At the moment there is no GUI for manually editing S3 credentials, but when credentials for Amazon Cloud Manager are updated, change is reflected to S3 as well. When all Amazon Cloud Managers are removed,
the S3 manager is also removed.

How does MIQ know which S3 manager belongs to a given Cloud Manager?
-> A new field `manager_group` is added to the ExtManagementSystem model. Cloud Managers with identic AWS_ACCESS_KEY_ID are given equal manager_group. S3 Manager is given same manager_group.

This commit depends on [introduction of `manager_group` field](https://github.com/ManageIQ/manageiq/pull/13771) in the main repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 /config/environments/*.local.yml
 
 /spec/manageiq
+.idea/*
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -22,6 +22,7 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
   OrchestrationTemplateCfn.register_eligible_manager(self)
 
   include ManageIQ::Providers::Amazon::ManagerMixin
+  include ManageIQ::Providers::Amazon::StandaloneS3Mixin
 
   has_one :network_manager,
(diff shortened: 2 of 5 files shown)


miha-plesko:
@miq-bot add_label providers/amazon

miha-plesko:
@miq-bot add_label wip

miha-plesko:
@blomquisg this commit is the very first attempt to address a situation where S3 Storage Manager is shared within multiple Cloud Managers. The main guideline at this point was "do it with as little changes as possible".

I'm pretty sure this is not the final version of it :innocent: so I'm looking forward to be hearing your opinion!

/cc @gberginc

miha-plesko:
Let me modify this to make use of Provider instead of `manager_group` attribute.

sergio-ocon:
Is there another PR solving this?

PR #138, opened 2017-02-14
[WIP] add amazon events

durandom:
I'm trying to add events to be displayed in timelines

If I do it this way, it will overwrite the `:critical` keys from manageiq core settings, because thats how the settings work. They overwrite - they don't merge.

1. Change settings processing to merge arrays instead of overwrite
2. Change [EmsEvent.event_groups](https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_event.rb#L28) to also look at other keys, eg. `:ems -> <provider_name> -> :event_handling ...`
3. Overhaul the event_groups and put them into own files? in `config/settings/events/<any_file>.yml`
4. Overhaul the event_groups and put them into `manageiq-content`
5. make `EmsEvent.event_groups` configurable / injectable from outside. But that would require bootstrapping from the provider. 

I tend to lean towards 3 or 4. While 3 has the benefit of being in the provider repo.
But I guess we also need to able to tune the events via advanced settings in the UI.
This is a plus for 2. But people will probably overlook stuff under different keys than `:event_handling`, so this speaks for 1. 

Well, well.

Maybe move it all out of settings and give it a proper UI and model it in the DB via seeding?

https://bugzilla.redhat.com/show_bug.cgi?id=1392391

@Fryguy @bdunne @blomquisg  ?

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,4 +1,17 @@
 ---
+:amazon:
+  :event_handling:
+    :event_groups:
+      :addition:
+        :critical:
+        - AWS_EC2_Instance_CREATE
+        - AWS_EC2_Instance_UPDATE
+        - AWS_EC2_Instance_DELETE
+      :power:
+        :critical:
+        - AWS_EC2_Instance_running
+        - AWS_EC2_Instance_shutting-down
+        - AWS_EC2_Instance_stopped
 :ems:
   :ems_amazon:
     :disabled_regions: []


Fryguy:
1. :-1: Merge is actually the default in config gem, and we specifically added overwrite to solve certain other problems.  One of the biggest issues with merge is when a customer wants to *remove* something, and then they need to knockout inherited values.  The knockout interface is clunky for Arrays, and would make it difficult for customers.

2. :+1:

3. :-1:  That's how this all started and it was a pain in the UI with multiple separate tabs.  You would also need an entire separate instance of Settings for it, which is also more pain.

4. :-1: manageiq-content will eventually be split into provider specific domains, and that includes events.  However, this code is really about UI presentation and not about the automate model, so I'm slightly against this idea.

5. This seems a lot like 2, no?  Or do you mean via some vmdb_plugin helper method?  I think that 2 can solve this in a much more elegant way.

durandom:
closing in favour of https://github.com/ManageIQ/manageiq-providers-amazon/pull/163

Issue #140, opened 2017-02-14
Event filtering through configuration


bronaghs:
This issue identifies how event filtering is done today under the covers and explores a way to make it configurable to the end user.

### Option 1 - Blacklist table
There is a table called the "blacklisted_events" that stores event types that will be filtered out for a specific provider type. The only way to populate this table today is by:

1. Hardcoding the event types in the CloudManager.rb file of the providers, eg:
https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager.rb#L105:111

2. Adding rows through the rails console, like this:
`BlacklistedEvent.create(:event_name => "filter_me", :provider_model => "ManageIQ::Providers::Amazon::CloudManager")`

### Option 2 - Event storms
Duplicate events that are received en mass within a short timeframe are considered an event storm and are de-duplicated for the VMware provider.
I am not sure this functionality can be leveraged for this purpose; configuring the handling of event storms is not available to the end user and it addresses a different use case.

Given that the `BlacklistedEvents` class and the corresponding table already exist for the purposes of event filtering, I think it makes sense to leverage it. One place to configure this blacklist is the settings.yml file, this file already has an event_handling section but it is enormous and looks like it needs to be moved out to its own file. I think a new events_setting.yml file should be created for each provider as follows:
crawl - it would support a blacklist of event types to be filtered
walk  - move over the entire `event_handling` section

bronaghs:
Pivotal story:
https://www.pivotaltracker.com/story/show/139861047

bascar:
This sounds like a good path forward for a crawl phase.

bronaghs:
@bascar - great, thanks

bronaghs:
Actually, amazon got its own settings file in December past, no need to create one then.

durandom:
>  One place to configure this blacklist is the settings.yml file, this file already has an event_handling section but it is enormous and looks like it needs to be moved out to its own file.

See the reasoning in https://github.com/ManageIQ/manageiq-providers-amazon/pull/138
Looks like "events_setting.yml" is not really the way to go, because it would require its own UI and bad (see jasons replies)

But once the events are put under their own key, I can see also moving the blacklisted events close to them.

I'm not sure though why the Blacklisted events are ActiveRecord objects in the DB and if they are needed to be in the DB - cause this would require seeding then.

bronaghs:
@durandom 
[...]

PR #141, opened 2017-02-16
Filter out events in Settings instead of CloudManager

bronaghs:
There is currently no way to filter out events using a configuration setting. There already is a blacklisted_events table but the only way to populate it is through hard coding event names in the provider cloud manager or to create rows through the rails console.

This PR creates a configuration setting for Amazon in the config/settings.yml file where event names to be filtered out can be added.

This change can be applied to each provider individually.

https://www.pivotaltracker.com/story/show/139861047
https://github.com/ManageIQ/manageiq-providers-amazon/issues/140

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -103,11 +103,7 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
   end
 
   def self.default_blacklisted_event_names
-    %w(
-      ConfigurationSnapshotDeliveryCompleted
-      ConfigurationSnapshotDeliveryStarted
-      ConfigurationSnapshotDeliveryFailed
-    )
+    Settings.ems.ems_amazon.blacklisted_event_names
   end
 
   #
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,6 +1,10 @@
 ---
 :ems:
   :ems_amazon:
+    :blacklisted_event_names:
+      - ConfigurationSnapshotDeliveryCompleted
+      - ConfigurationSnapshotDeliveryStarted
+      - ConfigurationSnapshotDeliveryFailed
     :disabled_regions: []
     # disable regions by their keys found in app/models/manageiq/providers/amazon/regions.rb e.g.
     # - us-gov-west-1


djberg96:
👍 LGTM

config/settings.yml
```diff
@@ -45,3 +45,7 @@
         :ems_refresh_worker_amazon_network: {}
         :ems_refresh_worker_amazon_ebs_storage: {}
         :ems_refresh_worker_amazon_s3: {}
+:blacklisted_event_names:
```
durandom: The problem with root level keys is, that they will be merged into the main settings.yml - which is editable via the UI 

I'm trying to get https://github.com/ManageIQ/manageiq/pull/13942 in and then we could have them under e.g. the `:amazon:` key

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -103,11 +103,7 @@ def self.hostname_required?
   end
 
   def self.default_blacklisted_event_names
-    %w(
-      ConfigurationSnapshotDeliveryCompleted
-      ConfigurationSnapshotDeliveryStarted
-      ConfigurationSnapshotDeliveryFailed
-    )
+    Settings.blacklisted_event_names
```
durandom: When the event names are at `:amazon:` this will of course change.
I also see in the core repo `::Settings`.

Also, how about making this change in the base class?
Or do you want to prove it first here?

durandom:
Do you happen to know if the blacklisted events in the db `BlacklistedEvent` could also move to here? Maybe we can even get away without the DB table?!

app/models/manageiq/providers/amazon/cloud_manager.rb
bronaghs: @durandom - I had ```::Settings``` and it made no difference when executed.
I think it would be best to get this working on a small scale, i.e. on AWS, then we can think about moving it to the base class and updating the other providers in a separate PR.

config/settings.yml
bronaghs: @durandom  - is the merge of ManageIQ/manageiq#13942 imminent?
Why would we not want this editable?

bronaghs:
Waiting on ManageIQ/manageiq#13942  to be merged so I can take advantage of the new keys.

config/settings.yml
durandom: @bronaghs sorry for the confusion - of course we want this editable. But the toplevel key `:blacklisted_event_names:` would be overwritten or overwrite the same key in the core repo or other providers - because the arrays under the key are not merged, but overwritten.

Thats why I think it should be under the `:amazon:` top level key namespace

durandom:
@bronaghs see https://github.com/ManageIQ/manageiq/pull/13942#pullrequestreview-23314723

Basically we decided against a toplevel `:amazon` key and went for top level `:ems`.
Under this is already `:ems_amazon` so you can put it under there

Although I was opting for `:ems -> :amazon`. But this is part of another PR

bronaghs:
@miq-bot remove_label wip

bronaghs:
@durandom - please review, thanks.

PR #143, opened 2017-02-17
remove validate_timeline

durandom:
since https://github.com/ManageIQ/manageiq/pull/10588 got merged (a long time ago) these methods are no more called anyway - remove cruft


Ladas:
:+1:

PR #145, opened 2017-02-20
set power state to shutting_down when rebooting an instance

durandom:
Sets the same power state `shutting_down` when rebooting an instance as in the [stop operation](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager/vm/operations/power.rb#L20)

`pending` is basically an intermediate state ([aws api](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceState.html)) - which we translate to [suspended](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager/vm.rb#L10). But we want to show it in the UI as `shutting down`

In case the operation was not carried out correct via aws, then the next refresh will set it to the real state.

@mkanoor I'm not sure if this has an impact on automate. See this PR https://github.com/ManageIQ/manageiq/pull/7477 - do you now if clicking a power button in the UI triggers a statemachine that could depend on the `pending` / `suspended` power state?

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1401487

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/vm/operations/guest.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/vm/operations/guest.rb
@@ -11,6 +11,6 @@ module ManageIQ::Providers::Amazon::CloudManager::Vm::Operations::Guest
   def raw_reboot_guest
     with_provider_object(&:reboot)
     # Temporarily update state for quick UI response until refresh comes along
-    self.update_attributes!(:raw_power_state => "pending") # show state as suspended
+    self.update_attributes!(:raw_power_state => "shutting_down")
   end
 end


bronaghs:
This is the same approach we have taken on the Azure integration. LGTM.

Ladas:
looks great :+1:

gberginc:
I only have one nitpick: I think the comment should also be changed because it still says "suspended".

This commit got me thinking about these states, because I am not updating them for cloud volumes while performing operations on them in #151.

durandom:
@gberginc thanks, a great example of the dreaded stale comment

tbh, I think setting this is not the responsibility of the provider, but should be the platform. So I would not bother with it for cloud volumes...

tinaafitz:
@durandom Would the power state be shutting_down in this case? If so, we would have to modify the retirement state machine to add the new power_state. Our retirement Automate method checks for the power states of "off suspended terminated unknown".

durandom:
@tinaafitz I'm not sure __when__ the statemachine kicks in. 

1. When the user clicks on the button? 
2. When a refresh reports updates the power state? 
3. When an `Event` in amazon is triggered via the EventCatcher?

This here is only for 1. 
And the change is only temporary, as `shutting_down` is no final state. It will be `running` again, once reboot has finished.

blomquisg:
@tinaafitz yeah, since this is just a reboot I don't think that retirement plays a role here.

PR #146, opened 2017-02-21
Renamed refresh strategies

Ladas:
Renamed refresh strategies


durandom:
doesnt this need a PR in core? Or have the strategies renamed there already?

Ladas:
@durandom it does, but there are no Specs for this in AWS, only specs are in core :-)

durandom:
requires changes from https://github.com/ManageIQ/manageiq/pull/13995

PR #147, opened 2017-02-22
Catch cloudwatch alarms

Ladas:
Catch cloudwatch alarm events + specs

### DIFF
--- a/spec/models/manageiq/providers/amazon/cloud_manager/event_parser_spec.rb
+++ b/spec/models/manageiq/providers/amazon/cloud_manager/event_parser_spec.rb
@@ -51,6 +51,16 @@ describe ManageIQ::Providers::Amazon::CloudManager::EventParser do
     end
   end
 
+  context "AWS CloudWatch Alarm" do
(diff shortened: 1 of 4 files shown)


spec/models/manageiq/providers/amazon/cloud_manager/event_parser_spec.rb
```diff
@@ -51,6 +51,14 @@
     end
   end
 
+  context "AWS CloudWatch Alarm" do
+    it "parses AWS_ALARM_awselb-EmSRefreshSpecVPCELB-Unhealthy-Hosts alarm event" do
+      event = parse_event("cloud_watch/AWS_ALARM_awselb-EmSRefreshSpecVPCELB-Unhealthy-Hosts.json")
+
+      expect(described_class.event_to_hash(event, nil)).to include(:vm_ems_ref => nil)
```
durandom: how about adding something you actually parse? like 'AlarmName' and 'time' ?

Ladas: added

PR #149, opened 2017-02-23
Pass collector to persister to make targeted refresh work

Ladas:
A temporary solution to make Targeted refresh work. We need to figure out a right place to parse all references that relate to a Target and passing them to Colletor and Persister


app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -1,5 +1,5 @@
 class ManageIQ::Providers::Amazon::Inventory::Collector::TargetCollection < ManageIQ::Providers::Amazon::Inventory::Collector
-  def initialize(_ems, _options, _target)
+  def initialize(_manager, _target)
```
durandom: would be great to move the whole amazon refresh to use the stock initializers :)

durandom:
Isnt there a spec for this? Or how do you test it?

Ladas:
@durandom specs are on the way :-)  I need to first figure out what the right target is :-)

PR #151, opened 2017-02-23
Cloud volume operations

gberginc:
This patch adds support for the following operations on the cloud volume: create new, attach, detach, delete and update name. Stubbed specs are provided for these operations.

@miq-bot assign @blomquisg

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
@@ -1,2 +1,83 @@
 class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVolume
+  supports :create
+
+  def self.validate_create_volume(ext_management_system)
+    validate_volume(ext_management_system)
+  end
+
+  def self.raw_create_volume(ext_management_system, options)
+    volume_name = options.delete(:name)
+    volume = nil
+
+    ext_management_system.with_provider_connection do |service|
+      # Create the volume using provided options.
+      volume = service.client.create_volume(options)
+      # Also name the volume using tags.
+      service.client.create_tags(
+        :resources => [volume.volume_id],
+        :tags      => [{ :key => "Name", :value => volume_name }]
+      )
+    end
+    {:ems_ref => volume.volume_id, :status => volume.state, :name => volume_name}
+  rescue => e
+    _log.error "volume=[#{volume_name}], error: #{e}"
+    raise MiqException::MiqVolumeCreateError, e.to_s, e.backtrace
+  end
+
+  def validate_update_volume
+    validate_volume
+  end
+
+  def raw_update_volume(options)
+    # Only name update is currently supported
+    if options.key?(:name)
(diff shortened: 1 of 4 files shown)


spec/factories/ext_management_system.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
```
gberginc: Honestly, I don't know if that's correct. While working on these changes, I added the `ems_amazon_ebs` in the core repo, but then decided to try adding new factory in the Amazon provider because the EBS manager is really specific to the Amazon provider itself.

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -151,4 +152,7 @@ def orchestration_template_validate(template)
     _log.error "template=[#{template.name}], error: #{err}"
     raise MiqException::MiqOrchestrationValidationError, err.to_s, err.backtrace
   end
+
+  def vm_detach_volume(vm, volume_id)
```
gberginc: Oh, missed this one. Will remove it in the update

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -53,6 +53,7 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
 
   delegate :cloud_volumes,
            :cloud_volume_snapshots,
+           :supports_create,
```
durandom: I don't think you'll want this. This would mean `Amazon::CloudManger.supports_create? #=> true`

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -1,2 +1,75 @@
 class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVolume
+  include_concern 'Operations'
+
+  include SupportsFeatureMixin
```
durandom: no need to include again, it's already on the base class, I guess?

spec/factories/ext_management_system.rb
durandom: yes, factories in the amazon repo are correct.

gberginc:
@durandom this PR now only adds support for cloud volume operations. Your comments have been addressed (removal of `supports` features). I'll make a separate PR for snapshots.

gberginc:
@durandom I was following the way other providers are doing the operations, yes. We've done the same while working on VMware cloud provider, so I assumed this is the way to go.

It hurt me while I was working on these operations because some are part of "operations", others are part of cloud volume/cloud volume snapshot. I am fine with merging all ops into `cloud_volume` but this will affect code consistency, which I cannot decide on my own if it's worth at this point.

durandom:
It hurts me too :) And we are now in our own repo, so merge it and reduce the amount of files

gberginc:
@durandom Done, `operations`-no-more :smile:. Should also pass all tests. When this gets merged I'll rebase the snapshots PR (but that one doesn't have operations).

PR #152, opened 2017-02-27
Support deletion of CloudObjectStoreObject

miha-plesko:
Support deletion of CloudObjectStoreObject for Amazon provider.

Depends on (from other repos):
* https://github.com/ManageIQ/manageiq/pull/14080 (merged)
* https://github.com/ManageIQ/manageiq/pull/14189 (must be merged in one take with this PR) 👈 ‼️ 

Will enable merge of:
* https://github.com/ManageIQ/manageiq-ui-classic/pull/497

PR #153, opened 2017-02-27
Support operation "clear" on CloudObjectStoreContainer

miha-plesko:
With this commit we implement raw method that clears bucket on the S3. We also render it supported so that GUI will display button. To `clear` bucket means to remove all objects within it.

Depends on (from other repos):
* https://github.com/ManageIQ/manageiq/pull/14082 (merged)

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
@@ -7,6 +7,21 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
     end
   end
 
+  supports :cloud_object_store_container_clear do
+    unless ext_management_system
+      unsupported_reason_add(
+        :cloud_object_store_container_clear,
+        _("The Storage Container is not connected to an active %{table}") % {
+          :table => ui_lookup(:table => "ext_management_systems")
+        }
+      )
+    end
+
+    unless cloud_object_store_objects.count.positive?
+      unsupported_reason_add(:cloud_object_store_container_clear, _("The Storage Container is already empty"))
+    end
+  end
+
   def provider_object(connection = nil)
     connection ||= ext_management_system.connect
     connection.bucket(ems_ref)
@@ -14,5 +29,9 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
 
   def raw_delete
     with_provider_object(&:delete!)
+  end
+
+  def raw_cloud_object_store_container_clear
+    with_provider_object(&:clear!)
   end
 end
--- a/spec/models/manageiq/providers/amazon/storage_manager/s3/stubbed_refresher_spec.rb
+++ b/spec/models/manageiq/providers/amazon/storage_manager/s3/stubbed_refresher_spec.rb
@@ -130,6 +130,42 @@ describe ManageIQ::Providers::Amazon::StorageManager::S3::Refresher do
         expect(bucket.supports?(:delete)).to be_falsey
       end
     end
+
+    it "clear bucket (trigger)" do
+      options = {:ids => [bucket.id], :task => "cloud_object_store_container_clear", :userid => "admin"}
+
+      expect { CloudObjectStoreContainer.process_tasks(options) }.to change { MiqQueue.count }.by(1)
+    end
+
+    it "clear bucket (process)" do
+      with_aws_stubbed(stub_responses) do
+        # should not remove from MIQ database, we rather rely on refresh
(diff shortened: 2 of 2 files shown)


miha-plesko:
@miq-bot add_label enhancement,wip

.gitignore
```diff
@@ -14,3 +14,5 @@
 /config/environments/*.local.yml
 
 /spec/manageiq
+
+.idea/*
```
bronaghs: @miha-plesko - what the idea directory for?

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container/operations.rb
```diff
@@ -0,0 +1,15 @@
+module ManageIQ::Providers::Amazon::CloudManager::CloudObjectStoreContainer::Operations
+  extend ActiveSupport::Concern
+
+  def raw_destroy
+    raise "CloudObjectStoreContainer has no #{ui_lookup(:table => "ext_management_systems")}, unable to destroy it" unless ext_management_system
```
bronaghs: same comment as https://github.com/ManageIQ/manageiq-providers-amazon/pull/152 about ```raise```

bronaghs: Do you need a ```:supports....``` statement for destroy?

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container/operations.rb
```diff
@@ -0,0 +1,15 @@
...
+    with_provider_object(&:delete!)
+    destroy
+  end
+
+  def raw_empty
+    raise "CloudObjectStoreContainer has no #{ui_lookup(:table => "ext_management_systems")}, unable to empty it" unless ext_management_system
```
bronaghs: Same here, I think it needs a ```:supports``` statement.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container/operations.rb
```diff
@@ -0,0 +1,15 @@
+module ManageIQ::Providers::Amazon::CloudManager::CloudObjectStoreContainer::Operations
+  extend ActiveSupport::Concern
+
+  def raw_destroy
+    raise "CloudObjectStoreContainer has no #{ui_lookup(:table => "ext_management_systems")}, unable to destroy it" unless ext_management_system
+    with_provider_object(&:delete!)
+    destroy
```
bronaghs: @blomquisg @gmcculloug
Should creating a compliance policy for Amazon S3 bucket deletion be a consideration here?

gmcculloug: You are referring to the policy checks to prevent actions like: [app/models/vm/operations/guest.rb#L23](https://github.com/ManageIQ/manageiq/blob/1261a4636668854e446fe1e603c8cc4bc4a6e626/app/models/vm/operations/guest.rb#L23)

That would be a good idea, but it should be in base method that calls the `raw_*` method, like the example above.

bronaghs: @gmcculloug - so the call to ```destroy``` here should be removed then I imagine?

gmcculloug: I am saying the caller to `raw_destroy` should be checking policy before calling this provider level method.  That way the policy check is performed for all providers from one spot.

bronaghs: Ah I see. Thanks.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  supports :empty do
+    unless ext_management_system
+      unsupported_reason_add(
+        :empty,
+        "CloudObjectStoreContainer has no #{ui_lookup(:table => "ext_management_systems")}, unable to empty it"
+      )
+    end
+  end
```
bronaghs: @miha-plesko - is this the only reason ```empty``` would be unsupported? If the bucket is empty should the menu item in the UI still be enabled?

bronaghs:
@miha-plesko - the test failure indicates the SupportsFeatureMixin isn't included, I think. ```supports``` isnt recognized. That doesn't seem right since none of this would work without it.

bronaghs:
@miha-plesko - can you rebase please?

miha-plesko:
@bronaghs Rebased. Please take a look.

@miq-bot assign @blomquisg

PR #154, opened 2017-02-27
[WIP] Save remote AWS tags locally

djberg96:
This is a first stab at converting remote key/value pairs on AWS tags into something we can store and re-use locally. This is complicated by the fact that:

* Currently tags, and their values, are split between two tables - classifications and tags.
* The classification description serves as the key's value. But, it must be unique within the table.
* The classification (or category subclass) is a self referential table.
* The tag and parent are then referenced within the classifications table for each value.
* The `taggings` table actually stores the reference to the tagged resource by type, e.g. VmOrTemplate.
* There are custom validations and methods on the `Tag`, `Classification`, `Category tables`.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
@@ -11,6 +11,9 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
     @data_index          = {}
     @known_flavors       = Set.new
     @options             = options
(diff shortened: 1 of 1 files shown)


djberg96:
@miq-bot add_label wip

bronaghs:
@djberg96 
I think the instantiation of Tags and Classification objects belong in post processing, having this logic in the refresh parser seems out of scope to me.
I think we need to create a ```:tag``` hash key for each object type in the various ```parse_*``` methods which will store the key / value pair. Then in post processing, together with the resource ID, this can be used to create the Tag and Classification objects.

djberg96:
@bronaghs Agreed.

PR #156, opened 2017-02-28
Add cloud volume snapshot operations for EBS manager

gberginc:
Create and delete snapshot operations are added to the cloud volume snapshot. Queued versions of these operations are provided allowing background creation and deletion of volume snapshots.

This currently depends on #151 adding cloud volumes support as it extends the cloud volume with the operation to create a new snapshot from the volume.

@miq-bot assign @durandom

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
@@ -76,6 +76,14 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
     end
   end
 
+  def create_volume_snapshot(options)
+    ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolumeSnapshot.create_snapshot(self, options)
+  end
+
+  def create_volume_snapshot_queue(userid, options)
+    ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolumeSnapshot.create_snapshot_queue(userid, self, options)
+  end
+
   def provider_object(connection = nil)
     connection ||= ext_management_system.connect
     connection.volume(ems_ref)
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -1,2 +1,73 @@
...
+  end
+
+  def raw_delete_volume
+    with_provider_object(&:delete)
+  rescue => e
+    _log.error "volume=[#{name}], error: #{e}"
+    raise MiqException::MiqVolumeDeleteError, e.to_s, e.backtrace
+  end
+
+  def create_volume_snapshot(options)
```
gberginc: This and the next function is required by this PR - remaining code in `cloud_volume` is taken from #151.

gberginc:
@miq-bot add_label wip

Adding a wip to ensure it gets merged after #151.

gberginc:
@miq-bot remove_label wip

@durandom Since #151 has been merged, I am removing wip here.

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
```diff
@@ -1,2 +1,100 @@
 class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolumeSnapshot < ::CloudVolumeSnapshot
+  include SupportsFeatureMixin
```
durandom: no need to include it again, its already in the base class

gberginc:
I've removed the unnecessary `include`.

Regarding the queuing comment: is there anything I should/could do here? Contrary to `CloudVolume` that has this in the [base model](https://github.com/manageiq/manageiq/blob/master/app/models/cloud_volume.rb#L32), `CloudVolumeSnapshot` has [nothing](https://github.com/manageiq/manageiq/blob/master/app/models/cloud_volume_snapshot.rb).

durandom:
> Regarding the queuing comment: is there anything I should/could do here? 

nope, it's not your fault - I mean you can clean the stuff up and refactor it, if you want... but :)

> Contrary to CloudVolume that has this in the base model, CloudVolumeSnapshot has nothing.

IIRC they were in the middle of a refactoring and probably did not finish it..

PR #157, opened 2017-02-28
[WIP] Add tagging support for Amazon

djberg96:
This is the beginning of a PR that imitates the Kubernetes code to support tagging for Amazon resources.


bronaghs:
@djberg96 -
I like this approach a lot. 
Couple of things:
- The ContainerLabel terminology is terribly distracting, we might have to discuss that
- Does ```ContainerLabelTagMapping.cache``` mean anything in AWS?
I had a look at the `ContainerLabelTagMapping.map_name_type_value` method and noticed it uses the cache so we might get stuck there. Did you past this point?

djberg96:
Closing since I don't think this is really the approach we want after all.

PR #158, opened 2017-03-02
Add options for graph refresh

Ladas:
Add options for graph refresh, by default graph refresh is switched off

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
@@ -46,7 +46,7 @@ class ManageIQ::Providers::Amazon::CloudManager::Refresher < ManageIQ::Providers
       all_targets, sub_ems_targets = targets.partition { |x| x.kind_of?(ExtManagementSystem) }
 
       unless sub_ems_targets.blank?
-        if refresher_options.try(:[], :event_targeted_refresh)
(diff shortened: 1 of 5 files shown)


Ladas:
@miq-bot assign @durandom

Ladas:
@blomquisg @durandom time has come to switch the graph refresh as default, it's also a nice time to rename the settings, inventory_object_refresh sounds weird

spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
```diff
@@ -9,6 +9,17 @@
     @ems.update_authentication(:default => {:userid => "0123456789ABCDEFGHIJ", :password => "ABCDEFGHIJKLMNO1234567890abcdefghijklmno"})
   end
 
+  before(:each) do
+    settings                                  = OpenStruct.new
```
durandom: I've been told we should stub settings like this https://github.com/ManageIQ/manageiq-providers-amazon/pull/158/files

Ladas: ok, let's do that once we have the helpers, leaving this way for now

Ladas:
@durandom @blomquisg changing the graph refresh as off by default, but adding the settings, so we can easily turn it on.

Are you fine about the naming? inventory_object_refresh sounds a bit weird

PR #159, opened 2017-03-03
Fix logger in Amazon EBS refresh parser

gberginc:
Amazon EBS refresh parser is using an inexistent logger (`_log`) resulting in errors due to a missing symbol. Patch fixes this by using the AWS-specific logger.

Thanks @miha-plesko for this find!

@miq-bot add_label bug
@Ladas would you be willing to review this one?


Ladas:
:+1: looks good

PR #160, opened 2017-03-03
Event parser can parse new format of target

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/14249

Event parser can parse new format of target + specs


Ladas:
@durandom @kbrock @agrare @blomquisg @Fryguy so this is how the new target format could look like, so we don;t require a Target being created in our DB

{"Vm" => {:manager_id => ems_id, :manager_ref => Set.new({:ems_ref => xy}, {:ems_ref => zz}) }}

It would be probably worth to pack this to ManagerRefresh::Target, where .serialize would return the ^ hash, so it can be queued to the refresh and we could then get it back with ManagerRefresh::Target.new(class, manager_id:, manager_ref: , ..)

then you could always get the records connected to the Target with:
 ``` target.manager_ref.map { |ref| target.class.find_by(ref.merge(:ems_id => target.manager_id)) } ```

spec/models/manageiq/providers/amazon/cloud_manager/event_parser_spec.rb
```diff
@@ -1,4 +1,12 @@
 describe ManageIQ::Providers::Amazon::CloudManager::EventParser do
+  before :each do
+    _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
+    @ems                 = FactoryGirl.create(:ems_amazon, :zone => zone)
+
+    allow_any_instance_of(EmsEvent).to receive(:handle_event)
```
durandom: `allow_any_instance_of` is the root of all evil. Chris and JoeR try hard to get eliminate it. Can do without it? Maybe inject an instance of EmsEvent into somewhere?

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -30,33 +59,132 @@ def self.parse_cloud_watch_alarm_event!(event, event_hash)
...
+  def self.parse_availability_zone_ref(event)
+    event.fetch_path("configurationItem", "availabilityZone")
+  end
+
+  def self.parse_ems_event_targets(event)
+    available_targets = self.available_targets || {}
+
+    case event.full_data["event_source"]
+    when :cloud_watch_api
+      collect_cloudwatch_api_references!(available_targets, event.full_data.fetch_path("detail", "requestParameters") || {})
```
durandom: all those bang methods. It's hard to tell what they actually modify.
This screams for extraction of an object 🛠

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -1,4 +1,33 @@
...
+      :source     => "AMAZON",
+      :full_data  => event,
+      :ems_id     => ems_id
+    }
+
+    unless %i(config cloud_watch_api cloud_watch_ec2 cloud_watch_alarm).include?(event["event_source"])
+      raise "Unsupported event source #{event["event_source"]}"
+    end
+
+    send("parse_#{event["event_source"]}_event!", event, event_hash)
```
durandom: chilly bang bang. What is modified by this method?
It would be much easier to follow if you return the modified version and assign it.

```ruby
event_hash = send("parse_#{event["event_source"]}_event", event)
```

durandom: but I just saw, this is already in the code :/

durandom:
I like the `.serialize` idea 👍 

Could you explain a bit, when and how this logic is triggered? 
So, when is a refresh scheduled for such a non existent target?

spec/models/manageiq/providers/amazon/cloud_manager/event_parser_spec.rb
Ladas: sure we can get rid of that if we will rewrite the code. The EmsEvent.add needs to have submethods, now it's all entangled together.

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
Ladas: yeah this would be probably nicer

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
Ladas: well they do build available_targets

Ladas:
@durandom so this logic is not trigered yet, but the workflow should be

1) parse event so we can store EmsEvent to the DB

2) if there is a targeted refresh, queue for refresh all targets we get from parser_class.parse_targets(ems_event) and optionally limit the targets parsed
- so the new automate method would be something like:
2.1) /System/event_handlers/graph_refresh # queues all available targets for refresh
2.2) /System/event_handlers/graph_refresh?limit_targets=Vm # queues only target of type Vm, if available
2.3) /System/event_handlers/graph_refresh?with_relations=Vm # queues Vm targeted refresh, with options to scan the Vm and queue also all it's relations for refresh, so it's a Vm full refresh

3) Save EmsEvent with all the targets as M:N relations, this needs to be invoked after the refresh, so all the targets are in the DB

app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
kbrock: sigh

durandom:
@Ladas 🤘 for creating all in new classes

spec/models/manageiq/providers/amazon/cloud_manager/event_target_parser_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
+    @ems                 = FactoryGirl.create(:ems_amazon, :zone => zone)
+
+    allow_any_instance_of(EmsEvent).to receive(:handle_event)
+    allow(EmsEvent).to receive(:create_completed_event)
+  end
+
+  context "AWS Config Event" do
+    it "parses vm_ems_ref into event" do
+      ems_event = create_ems_event("sqs_message.json")
```
durandom: Still, I would create an `EmsEvent` via a factory here and not test the whole EmsEvent Stream stack.

But I get your integration spec needs. But shouldnt those parts of the code already be tested somewhere else?

app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb
```diff
@@ -0,0 +1,139 @@
+class ManageIQ::Providers::Amazon::CloudManager::EventTargetParser
```
durandom: 👍 for this class

Now extract another class, e.g. `EventTarget` which holds `add_reference` and the `available_targets` data.  Then you have fully encapsulated that data you want to transport into a class and moved away from hashes.

spec/models/manageiq/providers/amazon/cloud_manager/event_target_parser_spec.rb
Ladas: I really want to test this on real events, so using the actual jsons of the event. I don't really want to stub the middlelayers

app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb
Ladas: done, ManagerRefresh::Target and ManagerRefresh::TargetCollection are being introduced as a dependency

app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb
```diff
@@ -0,0 +1,122 @@
...
+
+  def collect_cloudwatch_ec2_references!(target_collection, event_data)
+    instance_id = event_data.fetch_path("detail", "instance-id")
+    add_target!(target_collection, :vms, instance_id) if instance_id
+  end
+
+  def collect_config_references!(target_collection, event_data)
+    resource_type = event_data.fetch_path("configurationItem", "resourceType")
+    resource_id   = event_data.fetch_path("configurationItem", "resourceId")
+    target_class  = case resource_type
```
kbrock: would it make sense to do a hash lookup here?

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -131,72 +137,45 @@ def stack_template(stack_name)
...
+
+      vm.network_ports.collect(&:ems_ref).compact.each do |ems_ref|
+        target.add_target!(:association => :network_ports, :manager_ref => {:ems_ref => ems_ref})
+      end
+
+      vm.floating_ips.collect(&:ems_ref).compact.each do |ems_ref|
+        target.add_target!(:association => :floating_ips, :manager_ref => {:ems_ref => ems_ref})
+      end
     end
+    byebug
```
kbrock: heh

app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb
```diff
@@ -0,0 +1,122 @@
...
+    collect_cloudwatch_api_references!(target_collection, event_data["networkInterface"], depth + 1) if event_data["networkInterface"]
+
+    (event_data.fetch_path("groupSet", "items") || []).each do |x|
+      collect_cloudwatch_api_references!(target_collection, x, depth + 1)
+    end
+    (event_data.fetch_path("instancesSet", "items") || []).each do |x|
+      collect_cloudwatch_api_references!(target_collection, x, depth + 1)
+    end
+    (event_data.fetch_path("instances") || []).each do |x|
+      collect_cloudwatch_api_references!(target_collection, x, depth + 1)
```
kbrock: would it make sense to expand this api? `collect_cloud_watch_api_referenceses!`

of course, naming is hard. Then again, do you ever `collect_cloudwatch_api_references` with a single array?
```
collect_cloud_watch_api_referenceses!(target_collection, event_data.fetch_path("instances") || [], depth + 1)
```

app/models/manageiq/providers/amazon/inventory/collector.rb
```diff
@@ -30,49 +30,21 @@ def initialize(_manager, _target)
 
   def initialize_inventory_sources
     @instances                   = []
-    @instances_refs              = Set.new
```
kbrock: yay.

I'm assuming this doesn't add any O(N) lookups

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -131,72 +137,45 @@ def stack_template(stack_name)
...
   end
 
   def infer_related_ems_refs!
     # We have a list of instances_refs collected from events. Now we want to look into our DB and API, and collect
     # ems_refs of every related object. Now this is not very nice fro ma design point of view, but we really want
     # to see changes in VM's associated objects, so the VM view is always consistent and have fresh data. The partial
     # reason for this is, that AWS doesn't send all the objects state change,
-    changed_vms = manager.vms.where(:ems_ref => instances_refs.to_a).includes(:key_pairs, :network_ports, :floating_ips,
-                                                                          :orchestration_stack)
+    byebug
```
kbrock: heh

app/models/manageiq/providers/amazon/inventory/persister/target_collection.rb
```diff
@@ -24,6 +24,10 @@ def initialize_inventory_collections
 
   private
 
+  def references(collection)
```
kbrock: is this a generic method?
Can you get define it in somewhere like `ManagerRefresh::Inventory::Collector` or highter?

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
Ladas: hehe

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
Ladas: yaay :-)

app/models/manageiq/providers/amazon/inventory/collector.rb
Ladas: the refs were moved to ManagerRefresh::TargetCollection

app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb
Ladas: no usually the cloudwatch events have a nested arrays that contain references

app/models/manageiq/providers/amazon/inventory/persister/target_collection.rb
Ladas: right, it's not really generic since it states :ems_ref, but maybe I could pack part of it to the TargetCollection

app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb
Ladas: possibly, but maybe the case is enough? Should be still pretty quick

app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb
```diff
@@ -0,0 +1,122 @@
...
+  # @param ems_event [EmsEvent] EmsEvent object
+  def initialize(ems_event)
+    @ems_event = ems_event
+  end
+
+  # Parses all targets that are present in the EmsEvent given in the initializer
+  #
+  # @return [Array] Array of ManagerRefresh::Target objects
+  def parse
+    parse_ems_event_targets(ems_event)
```
durandom: Now that you add `ems_event` as a state (instance variable), I think you can also create `target_collection` as an instance var and let `collect_cloudwatch_api_references` operate on that.

app/models/manageiq/providers/amazon/inventory/collector.rb
```diff
@@ -3,22 +3,22 @@ class ManageIQ::Providers::Amazon::Inventory::Collector < ManagerRefresh::Invent
...
-  attr_reader :public_images, :public_images_refs, :public_images_deleted
-  attr_reader :cloud_networks, :cloud_networks_refs, :cloud_networks_deleted
-  attr_reader :cloud_subnets, :cloud_subnets_refs, :cloud_subnets_deleted
-  attr_reader :security_groups, :security_groups_refs, :security_groups_deleted
-  attr_reader :floating_ips, :floating_ips_refs, :floating_ips_deleted
-  attr_reader :network_ports, :network_ports_refs, :network_ports_deleted
-  attr_reader :load_balancers, :load_balancers_refs, :load_balancers_deleted
-  attr_reader :stacks, :stacks_refs, :stacks_deleted
-  attr_reader :cloud_volumes, :cloud_volumes_refs
-  attr_reader :cloud_volume_snapshots, :cloud_volume_snapshots_refs
```
durandom: 👏 for less `attr_reader`s

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -3,24 +3,30 @@ def initialize(_manager, _target)
     super
     parse_targets!
     infer_related_ems_refs!
+
+    target.manager_refs_by_association_reset
```
durandom: This would be a perfect `!` method

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -131,72 +137,43 @@ def stack_template(stack_name)
...
-
-    (hash.fetch_path("groupSet", "items") || []).each { |x| collect_references!(x) }
-    (hash.fetch_path("instancesSet", "items") || []).each { |x| collect_references!(x) }
-    (hash.fetch_path("instances") || []).each { |x| collect_references!(x) }
-    (hash.fetch_path("networkInterfaceSet", "items") || []).each { |x| collect_references!(x) }
-  end
-
   def parse_vm_target!(t)
-    instances_refs << t.ems_ref if t.ems_ref
+    target.add_target(:association => :vms, :manager_ref => {:ems_ref => t.ems_ref}) if t.ems_ref
```
durandom: If  this is mostly always `target.add_target`, you might also allow `target.add` syntax

durandom: How about renaming this to `targets` or `target_collection`

app/models/manageiq/providers/amazon/inventory/collector.rb
kbrock: Ooh, then that is very cool

PR #164, opened 2017-03-06
List VMs from the same availability zone as volume

gberginc:
This patch implements a method from the base cloud volume that provides
the list of VMs (instances) that the volume can be attached to. In case
of Amazon, the volume can be attached to any instance residing in the
same availability zone as the volume itself.

Depends on: ManageIQ/manageiq#14058

@miq-bot add_label enhancement,pending core


gberginc:
The changes have been merged in [core](https://github.com/ManageIQ/manageiq/pull/14058); UI being under [review](https://github.com/ManageIQ/manageiq-ui-classic/pull/601).

@miq-bot remove_label pending core

Ladas:
looks great :+1:

PR #165, opened 2017-03-07
Set default Container and Object class when collecting inventory for the 1st time

miha-plesko:
When user adds AWS cloud manager S3 storage manager is added as well and refresh is triggered for it. However, during this first refresh, CloudObjectStoreContainers and CloudObjectStoreObjects were assigned empty type. This prevented S3 storage manager from showing them as its children. When user triggered refresh on S3 manually, S3 manager assumed that no object is in database yet resulting in every object
being fetched once again while keeping orphaned ones in database. And again. And again... The database was getting bigger and bigger with every new refresh.

Fixed with this commit - turns out we have to specify default class for Containers and Objects and the orphaned database entries wont appear.

@miq-bot add_label bug
@miq-bot assign @Ladas

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/parser/storage_manager/s3.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/storage_manager/s3.rb
@@ -28,7 +28,6 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::StorageManager::S3 < Manag
   def parse_container(bucket)
     uid = bucket['name']
     {
-      :type                  => self.class.container_type,
       :ext_management_system => ems,
       :ems_ref               => uid,
       :key                   => bucket['name']
@@ -65,7 +64,6 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::StorageManager::S3 < Manag
   def parse_object(object, bucket)
     uid = object['key']
     {
-      :type                         => self.class.object_type,
       :ext_management_system        => ems,
       :ems_ref                      => "#{bucket.ems_ref}_#{uid}",
       :etag                         => object['etag'],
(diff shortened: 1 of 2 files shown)


miha-plesko:
After @Ladas noticed that `require_nested` is missing (see https://github.com/ManageIQ/manageiq-providers-amazon/pull/166), I'm not sure that this is still needed.

Ladas:
This is preferable to adding the type in the parser. We will change the syntax of the above and then probably do a 1 big cleanup.

The difference is that saving code is doing:

::ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreObject.create

vs CloudObjectStoreObject.create(:type =>"::ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreObject" )

so calling .create directly on the class is cleaner. But :type can still be used if there are sub STIs in the parser.

----

that being said, I think we want this merged, so we can remove the :type from the parser.

miha-plesko:
@Ladas thanks for explanation, I shall repush with type removed, gimme a minute.

miha-plesko:
@Ladas probably I should only remove in new parser
https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/inventory/parser/storage_manager/s3.rb#L68
 and leave it be in old one
https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/storage_manager/s3/refresh_parser.rb#L97
 right?

Ladas:
@miha-plesko yes, only the new parser supports this

Ladas:
looks great :+1:

Issue #161, opened 2017-03-06
CloudObjectStoreContainer warning and MissingRegionError


djberg96:
Running on latest master with a refreshed database.

I spotted an error and a warning in the evm.log during an AWS refresh that may be of concern. Neither appears to affect the refresh, but they should probably be addressed.

The warning is: `warning: toplevel constant CloudObjectStoreContainer referenced by ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer`

The error is: `[Aws::Errors::MissingRegionError]: missing region; use :region option or export region name to ENV['AWS_REGION']  Method:[rescue in block in refresh]`

To duplicate, just add or refresh an AWS provider that contains resources. The full backtrace that I'm seeing is below:

```
[----] I, [2017-03-06T09:10:10.300053 #79505:3fc7d783fa0c]  INFO -- : MIQ(ManageIQ::Providers::Amazon::StorageManager::S3::Refresher#refresh_targets_for_ems) EMS: [amazon S3 Storage Manager], id: [8] Refreshing target ManageIQ::Providers::Amazon::StorageManager::S3 [amazon S3 Storage Manager] id [8]...
/Users/dberger/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/manageiq-providers-amazon-b65b8993449b/app/models/manageiq/providers/amazon/inventory/parser/storage_manager/s3.rb:80: warning: toplevel constant CloudObjectStoreContainer referenced by ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
/Users/dberger/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/manageiq-providers-amazon-b65b8993449b/app/models/manageiq/providers/amazon/inventory/parser/storage_manager/s3.rb:80: warning: toplevel constant CloudObjectStoreContainer referenced by ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
[----] E, [2017-03-06T09:10:11.343181 #79505:3fc7d783fa0c] ERROR -- : MIQ(ManageIQ::Providers::Amazon::StorageManager::S3::Refresher#refresh) EMS: [amazon S3 Storage Manager], id: [8] Refresh failed
[...]

PR #166, opened 2017-03-07
Suppress warning "toplevel constant referenced"

miha-plesko:
As reported in https://github.com/ManageIQ/manageiq-providers-amazon/issues/161, calling
```ruby
ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer.name
```
in refresh parser resulted in warning "warning: toplevel constant CloudObjectStoreContainer referenced by ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer" being printed to evm.log. Fixing this by hard-coding `type` name as string. Other refresh parsers have it hard-coded as well, but I wanted to be smarter... :D

@miq-bot add_label bug
@miq-bot assign @Ladas

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/s3.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/s3.rb
@@ -1,4 +1,6 @@
 class ManageIQ::Providers::Amazon::StorageManager::S3 < ManageIQ::Providers::StorageManager
+  require_nested :CloudObjectStoreContainer
+  require_nested :CloudObjectStoreObject
   require_nested :RefreshParser
   require_nested :RefreshWorker
   require_nested :Refresher


Ladas:
@miha-plesko hello, the actual fix for this should be require the STI models, like this https://github.com/Ladas/manageiq-providers-amazon/blob/a5ed18d77e7661b58791c2789445f384c3f6e983/app/models/manageiq/providers/amazon/storage_manager/ebs.rb#L2-L2

Then the warning should go away. The warning actually shows that the class was not loaded correctly.

miha-plesko:
@Ladas thanks for telling me this! Also, by adding `require_nested` stanzas, we seem to have resolved this https://github.com/ManageIQ/manageiq-providers-amazon/pull/165.

Q: Is it any good to be doing [THIS](https://github.com/ManageIQ/manageiq-providers-amazon/pull/165/files#diff-3cd026a24503ab2b6944f71ceb6dca75) in addition to `require_nested`?

Ladas:
@miha-plesko commented in  https://github.com/ManageIQ/manageiq-providers-amazon/pull/165/

Ladas:
:+1:

PR #168, opened 2017-03-08
AWS SDK call `get_bucket_location` returns empty string

miha-plesko:
For buckets in US East (N. Virginia) region, the SDK returns empty string "" instead of "us-east-1". But when connecting to S3, we need to specify "us-east-1" explicitly. With this commit we map empty string to region name to prevent error.

@miq-bot assign @Ladas 


djberg96:
Yep, weird API decision:

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETlocation.html

Thanks for fixing.

👍

bronaghs:
That is the oddest API design decision ever. Thanks Miha.

app/models/manageiq/providers/amazon/inventory/parser/storage_manager/s3.rb
```diff
@@ -37,6 +37,7 @@ def parse_container(bucket)
   def process_objects(bucket_id, bucket_object)
     # S3 bucket accessible only for API client with same region
     region = collector.aws_s3.client.get_bucket_location(:bucket => bucket_id).location_constraint
+    region = "us-east-1" if region.empty? # SDK returns empty string for default region
```
Ladas: good catch, this probably the failure I was seeing

PR #169, opened 2017-03-09
Connect to correct region with S3

miha-plesko:
Current code always connects to S3's region that of cloud manager. This works fine when buckets that we're accessing are within that region, but results in error when accessing bucket from any other region.
    
With this commit we update provider_object for CloudObjectStoreContainer and CloudObjectStoreObject so that it now connects to appropriate region.

@miq-bot assign @bronaghs

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
@@ -23,8 +23,14 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
   end
 
   def provider_object(connection = nil)
+    connect(connection).bucket(ems_ref)
+  end
+
+  def connect(connection = nil)
     connection ||= ext_management_system.connect
-    connection.bucket(ems_ref)
+    region = connection.client.get_bucket_location(:bucket => ems_ref).location_constraint
+    region = "us-east-1" if region.empty? # SDK returns empty string for default region
+    ext_management_system.connect(:region => region)
   end
 
   def raw_delete
--- a/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
@@ -14,8 +14,7 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreObject < 
   end
 
   def provider_object(connection = nil)
-    connection ||= ext_management_system.connect
-    connection.bucket(cloud_object_store_container.ems_ref).object(key)
+    cloud_object_store_container.provider_object(connection).object(key)
   end
 
   def raw_delete
--- a/spec/models/manageiq/providers/amazon/storage_manager/s3/stubbed_refresher_spec.rb
+++ b/spec/models/manageiq/providers/amazon/storage_manager/s3/stubbed_refresher_spec.rb
@@ -85,10 +85,12 @@ describe ManageIQ::Providers::Amazon::StorageManager::S3::Refresher do
     end
 
     it "bucket's provider_object is of expected type" do
-      provider_obj = bucket.provider_object
+      with_aws_stubbed(stub_responses) do
+        provider_obj = bucket.provider_object
 
-      expect(provider_obj).to_not be_nil
-      expect(provider_obj.class).to eq(Aws::S3::Bucket)
+        expect(provider_obj).to_not be_nil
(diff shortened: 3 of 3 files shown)


app/models/manageiq/providers/amazon/storage_manager/s3.rb
```diff
@@ -38,4 +38,9 @@ def connect(options = {})
     options[:service] = :S3
     parent_manager.connect options
   end
+
+  def connect_regional(region)
+    @regional_connection ||= {}
+    @regional_connection[region] ||= connect(:region => region)
```
Fryguy: Not sure I like that we are caching the connection.  Is there a reason for that?

Additionally, this feels like it should just be a parameter to connect method above instead of a whole method.  That is, the caller can just do `connect(:region => region)`

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -23,7 +23,7 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
   end
 
   def provider_object(connection = nil)
-    connection ||= ext_management_system.connect
+    connection = ext_management_system.connect_regional(provider_region)
```
Fryguy: The ||= should stay, because it allows the caller to override the connection used by passing one in (pretty sure this was a dependency injection for testing purposes).

miha-plesko:
@blomquisg @Fryguy this is a substitute to https://github.com/ManageIQ/manageiq/pull/14242. We query bucket region on-the-fly so that we don't need to have additional column in the database. This should not be a preformance problem since it's only used when user clicks button on the GUI.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -24,7 +24,11 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
 
   def provider_object(connection = nil)
     connection ||= ext_management_system.connect
-    connection.bucket(ems_ref)
+    region = connection.client.get_bucket_location(:bucket => ems_ref).location_constraint
+    region = "us-east-1" if region.empty? # SDK returns empty string for default region
```
gberginc: Can you move this (L27-28) to a helper (private) method?

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -24,7 +24,11 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
 
   def provider_object(connection = nil)
     connection ||= ext_management_system.connect
-    connection.bucket(ems_ref)
```
gberginc: What if `connection` parameter is given? Should it fail in case it is not correct for the corresponding bucket?

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
```diff
@@ -14,8 +14,8 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreObject <
   end
 
   def provider_object(connection = nil)
-    connection ||= ext_management_system.connect
-    connection.bucket(cloud_object_store_container.ems_ref).object(key)
+    bucket = cloud_object_store_container.provider_object(connection)
```
gberginc: Similar question as above: if connection is given, should it fail?

gberginc: Any reason not to merge L17 and L18 (as it was before).

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
gberginc: Alternatively (perhaps even more clean to me), you could have a `def connect` in the `cloud_object_store_container` that would do 

```
    region = connection.client.get_bucket_location(:bucket => ems_ref).location_constraint
    region = "us-east-1" if region.empty? # SDK returns empty string for default region
    ext_management_system.connect(:region => region)
```

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
gberginc: If you go with the `connect` proposal above, simplify the code here.

Ladas:
looks great :+1:

PR #170, opened 2017-03-09
Maintain instance types

durandom:
This takes data from 
* https://raw.githubusercontent.com/powdahound/ec2instances.info/master/www/instances.json
* https://aws.amazon.com/ec2/instance-types/#instance-type-matrix
* https://aws.amazon.com/ec2/previous-generation/#Previous_Generation_Instance_Details_and_Pricing_

and creates a new `instance_types.rb` file to be copied over `app/models/manageiq/providers/amazon/instance_types.rb` 

It can be run via a rake task

```bash
❯ bin/rails providers:amazon:instance_types
Created ./instance_types.rb with 21 new, 4 newly deprecated and 0 new
discontinued instances.
Now copy it to the existing one:

mv instance_types.rb
app/models/manageiq/providers/amazon/instance_types.rb
```

It maintains order of the current instances.

**NOTE:** I havent found a better place than a rake task for this. I started out with a spec, but specs should not create files. Then a `tools/` script, but I can't `rails runner` in an engine. Now it's a task. 
For this, I had to load a engine tasks in the Rakefile. I opted to put them under the `providers:amazon` namespace, because they are also loaded in the core repo

If you want I could split this into 2 PRs

1. change Rakefile
2. add rake task

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1427966

cc @bronaghs @Fryguy @blomquisg


durandom:
includes https://github.com/ManageIQ/manageiq-providers-amazon/pull/171 
because new instances is euwe/yes and this is euwe/no

PR #174, opened 2017-03-13
Do vm targeted full refresh by default

Ladas:
If VM was scheduled for refresh, always refresh the references that
we can obtain by scanning the Vm in the API and our DB. This way, we
should always get a Vm that has all references filled.

The alternative for this would be a multi-phase refresh, where each
refresh can be planning another refresh for missing references.

In this case, we output a fully refreshed Vms in one step, which
will be later configurable, as this is handy for integration
with external tools. E.g. when we need to send refreshed data
to ansible, we need Vm completely refreshed after 1 refresh, not
possible many incremental refreshes.

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
@@ -149,7 +149,7 @@ class ManageIQ::Providers::Amazon::Inventory::Collector::TargetCollection < Mana
   end
 
   def parse_vm_target!(t)
-    target.add_target(:association => :vms, :manager_ref => {:ems_ref => t.ems_ref}) if t.ems_ref
+    add_simple_target!(:vms, t.ems_ref)
   end
 
   def infer_related_ems_refs!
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -12,7 +13,8 @@ def references(collection)
   end
 
   def instances
-    hash_collection.new(
+    return @instances unless @instances.blank?
```
durandom: stay positive: `if @instances.any?`

app/models/manageiq/providers/amazon/inventory/parser/storage_manager/ebs.rb
```diff
@@ -76,14 +76,11 @@ def link_volume_to_disk(volume_hash, attachments)
 
       dev = File.basename(a['device'])
 
-      disk_hash = {
```
durandom: this `disk` part seems unrelated to the rest - can you split that into a new PR?

Ladas: ah this is related to the targeted refresh, this code doesn't work for targeted without the find_or_build_by

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
Ladas: the ! of .blank? is .present?, but the unless here should be ok :-)

durandom: ? :) !

Yeah, any? present?. I'm just saying `return if` is easier to grok than `return unless`. 

Always try to use a positive version...

app/models/manageiq/providers/amazon/inventory/parser/storage_manager/ebs.rb
durandom: so, the changes here in `storage_manager/ebs.rb ` and `inventory_collection_default/cloud_manager.rb` are required to make the changes in `collector/target_collection.rb` to work?

Ladas: yes, the parser is loading existing InventoryObject and modifies them, so we need the find_or_build for that. It's not an issue when the parser runs alone, then it build it.

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
Ladas: changed this in a followup with specs :-)

PR #175, opened 2017-03-13
Relative path for the VCRs is allowing to run specs from the core

Ladas:
Relative path for the VCRs is allowing to run specs from the core

### DIFF
--- a/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
+++ b/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
@@ -1,4 +1,13 @@
 module AwsRefresherSpecCommon
+  extend ActiveSupport::Concern
+
+  included do
+    # We need a relative path so the specs are executable from the core
+    VCR.configure do |c|
+      c.cassette_library_dir = File.join(File.dirname(__FILE__), "../" * 4 + "vcr_cassettes")
+    end
+  end
+
   def assert_common
     assert_table_counts
     assert_ems


durandom:
This is only needed if you run the specs via rspec from the manageiq core checkout (as the current working directory). Because then the `spec_helper.rb` from the amazon is not loaded and hence does not set the vcr paths.

This has the same effect as this PR
```
rspec --require ./plugins/manageiq-providers-amazon/spec/spec_helper.rb ./plugins/manageiq-providers-amazon/spec/models/manageiq/providers/amazon/cloud_manager/refresher_spec.rb
```

Ladas:
@durandom hm, I will need to tinker with this more it need it as simple as:

<max 3 short words, like bundle exec rspec> <copy paste the path to spec>

so maybe alias for the rails test, we need something like that ^ and put to to the docs :-)

durandom:
@chrisarcand is there an 'easy' way to include any spec_helper in directories up from the path of the specs?
Or maybe add `.rspec` files that are up from the path?

chrisarcand:
> This is only needed if you run the specs via rspec from the manageiq core checkout (as the current working directory).

Yeah, wat? I don't see the reason for this change. The configuration here should be for running the specs...here. It should be up to the dev on the CLI as demonstrated if they really want to run specs in the example app that this plugin is included in (aka core), otherwise one should expect to run them from the top level of the gem as normal. 

In any case, the other issue is that you're setting a singleton setting on a gem in an `included` hook on a module, which is... 😬

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -1,4 +1,13 @@
 module AwsRefresherSpecCommon
+  extend ActiveSupport::Concern
+
+  included do
+    # We need a relative path so the specs are executable from the core
+    VCR.configure do |c|
+      c.cassette_library_dir = File.join(File.dirname(__FILE__), "../" * 4 + "vcr_cassettes")
```
bdunne: Why not use `ManageIQ::Providers::Amazon::Engine.root.join(...)`

bdunne:
I agree with @chrisarcand 
It is not common practice to run engine specs from within the "dummy app".  In this engine we are currently using ManageIQ as the "dummy" app, but I expect that may change as we further split our application.

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
Ladas: indeed, that is the right question :-)

Ladas:
@bdunne @chrisarcand well I do run all the specs within MIQ core, because usually I change core+aws or more gems at once. But yeah, I suppose it's a matter of having a nice alias for some more complex rspec call.

bdunne:
@Ladas If I need to work on a gem and core at the same time, I symlink `spec/manageiq` to my clone of manageiq and either update the ManageIQ Gemfile or add an `override_gem` and reference the gem by path.  I run the ManageIQ specs from the manageiq directory and the gem specs from the gem directory.

Ladas:
@bdunne I'll try that, thanks. (I see it's in the developer setup docs :-))

PR #176, opened 2017-03-13
Increase minimum AWS version for modify_volume

gberginc:
In order to enable updating EBS volumes on the fly, the SDK needs to support the [#modify_volume](http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html#modify_volume-instance_method). This has been introduced in version [2.7.8](https://github.com/aws/aws-sdk-ruby/commit/9e0558fdfb0eec9ed03f500d552edd3b36cdaa9f#diff-be5c280d1f1bc5bd07cf35078448e1caR1668) of aws-sdk gem.

This patch thus increases the minimum supported version to v2.7.8.

### DIFF
--- a/manageiq-providers-amazon.gemspec
+++ b/manageiq-providers-amazon.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config.lib}/**/*"]
 
-  s.add_dependency("aws-sdk", ["~>2.6.14"])
+  s.add_dependency("aws-sdk", ["~>2.8.7"])
 
   s.add_development_dependency "codeclimate-test-reporter", "~> 1.0.0"
   s.add_development_dependency "simplecov"


gberginc:
I am adding wip label as I'd first like to check if any of the tests will fail.

@miq-bot add_label wip

gberginc:

@miq-bot assign @blomquisg 

@blomquisg As discussed today, I've checked the changelog. It's huge, but based on the docs and skimming through the code, I couldn't spot anything that could break the provider (I thought moving from 2.6.x to 2.7.x would introduce more changes but it doesn't seem the case).

durandom:
@gberginc any reason not to move to the latest? 2.8.7 https://rubygems.org/gems/aws-sdk/versions

durandom:
@Ladas @bronaghs dont we also need CloudWatchEvents https://github.com/aws/aws-sdk-ruby/blob/master/CHANGELOG.md - they have been added in 2.8.7

gberginc:
@durandom the only reason for me was that I wanted to make sure the minimum version is set to the one introducing `#modify_volume`. I am fine changing to the latest.

Ladas:
I am fine with going to the latest to see if something breaks :-) @bronaghs you might use the cloudwatch API for the automated rules?

bronaghs:
@Ladas - are you referring to the event switchboard? If so, then, yes, that will need looked into. However, the last time I rebooted an instance I only got an AWS Config event, maybe that was fixed in this release.. or maybe that was operator error :-)  I will look into it.

I can foresee no problems upgrading, lets give it a go.

gberginc:
@durandom looks like 2.8.7 also passed all tests :-).

PR #177, opened 2017-03-13
Enable dynamic cloud volume modifications

gberginc:
Recently Amazon introduced a capability to modify existing cloud volumes
on the fly. Depending on the volume type size, volume type and IOPS can
be modified without detaching the volume first.

This patch enhances the existing `update_volume` method that previously
allowed changing only the name of the volume by using the new method
`#modify_volume` from the aws-sdk gem. The patch also adds a set of
tests verifying different possible options passed to the
`update_volume`.

Depends on #176 (requires newer aws-sdk gem).

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
@@ -33,11 +33,13 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
   end
 
   def raw_update_volume(options)
-    # Only name update is currently supported
-    if options.key?(:name)
-      with_provider_object do |volume|
-        volume.create_tags(:tags => [{:key => "Name", :value => options[:name]}])
-      end
+    with_provider_object do |volume|
+      # Update the name in case it was provided in the options.
+      volume.create_tags(:tags => [{:key => "Name", :value => options[:name]}]) if options.key?(:name)
+
+      # Mofiy volume configuration based on the given options.
+      modify_opts = modify_volume_options(options)
+      volume.client.modify_volume(modify_opts.merge(:volume_id => ems_ref)) unless modify_opts.empty?
     end
   rescue => e
     _log.error "volume=[#{name}], error: #{e}"
@@ -91,5 +93,18 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
   def provider_object(connection = nil)
     connection ||= ext_management_system.connect
     connection.volume(ems_ref)
+  end
+
+  private
+
+  def modify_volume_options(options = {})
+    modify_opts = {}
+    if volume_type != 'standard'
+      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != 'standard'
+      modify_opts[:size]        = Integer(options[:size]) if options[:size] && Integer(options[:size]).gigabytes != size
+      modify_opts[:iops]        = options[:iops] if (options[:volume_type] == "io1" || volume_type == 'io1') && options[:iops]
+    end
+
+    modify_opts
   end
 end
--- a/spec/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_spec.rb
+++ b/spec/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_spec.rb
@@ -4,7 +4,7 @@ describe ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume do
   let(:ems_cloud) { FactoryGirl.create(:ems_amazon_with_authentication) }
   let(:ebs) { FactoryGirl.create(:ems_amazon_ebs, :parent_ems_id => ems_cloud.id) }
   let(:availability_zone) { FactoryGirl.create(:availability_zone_amazon) }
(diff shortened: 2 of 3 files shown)


manageiq-providers-amazon.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config.lib}/**/*"]
 
-  s.add_dependency("aws-sdk", ["~>2.6.14"])
+  s.add_dependency("aws-sdk", ["~>2.7.8"])
```
durandom: can you put this in a separate PR - for backporting reasons...

durandom: oh, it's already in #176  👍

gberginc:
Since #176 was just merged, I rebased it and removed the gem version bump from this one.

spec/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_spec.rb
```diff
@@ -35,6 +35,139 @@
...
     end
 
+    context "#update_volume" do
+      let(:modify_volume_stub) { { :modify_volume => {} } }
+
+      context "updating volume name" do
+        it "is allowed when :name is provided" do
+          # Name is set via AWS tags so stub the create_tags.
+          expect_any_instance_of(Aws::EC2::Volume).to receive(:create_tags).with(
+            :tags => [{:key => "Name", :value => "new_name"}]
```
durandom: I think you can also stub this. add the api calls to `with_aws_stubbed` below

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -33,12 +33,17 @@ def validate_update_volume
   end
 
   def raw_update_volume(options)
-    # Only name update is currently supported
-    if options.key?(:name)
-      with_provider_object do |volume|
-        volume.create_tags(:tags => [{:key => "Name", :value => options[:name]}])
-      end
+    res = nil
```
durandom: does any caller really care about the return value?

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -92,4 +97,17 @@ def provider_object(connection = nil)
...
     connection.volume(ems_ref)
   end
+
+  private
+
+  def modify_volume_options(options = {})
+    modify_opts = {}
+    if volume_type != 'standard'
+      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != 'standard'
+      modify_opts[:size]        = options[:size] if options[:size] && options[:size] != (size / 1_073_741_824)
```
durandom: Is this something like `1.gigabyte` ?

spec/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_spec.rb
gberginc: Possible, but only to some degree. I wanted to ensure that specific arguments are used as well as check whether the call is made or not (if name is given, it must be called, if it's not given, it should not be called).

What I could do is ignore input parameters to `create_tags` and then use an exception in the stub to ensure it doesn't get called. I'll have a look.

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
gberginc: I am not completely sure. Perhaps I just misunderstood what's needed.

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
gberginc: true, alternative check would be `... && options[:size].gigabyte == size`. Database stores the size in bytes, but the `:size` in options must be in gigabytes.

gberginc:
@durandom I've updated the change:
 * use stubs for all tests: in these stubs I have explicitly set exceptions for the calls for which I'd like to ensure they are not called (for example, when `:name` is not given, it should not call `create_tags`, so I set the stubbed response to `failed` to ensure exception is thrown in case the call is made)
 * use `gigabyte` instead of that ugly number
 * ignore return value of `update_volume` - it really doesn't seem like it's used somewhere

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -92,4 +94,17 @@ def provider_object(connection = nil)
...
     connection.volume(ems_ref)
   end
+
+  private
+
+  def modify_volume_options(options = {})
+    modify_opts = {}
+    if volume_type != 'standard'
+      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != 'standard'
+      modify_opts[:size]        = options[:size] if options[:size] && options[:size].gigabytes != size
```
durandom: where does `options[:size]` come from? Maybe do `Integer(options[:size])` ?

gberginc: @durandom Thanks for spotting this. So far value of `options[:size]` came from the UI where I tested only with valid values, however it could also come from API or automate which means that it could in fact be invalid. I added two more tests to ensure that valid and invalid strings work properly with the updated function.

Ladas:
looks great, very nice specs :+1:

PR #180, opened 2017-03-16
Use secrets for VCR based specs

Ladas:
Use secrets for VCR based specs

### DIFF
--- a/spec/factories/ext_management_system.rb
+++ b/spec/factories/ext_management_system.rb
@@ -6,14 +6,14 @@ FactoryGirl.define do
     provider_region "us-east-1"
   end
 
-  factory :ems_amazon_with_vcr_authentication, :parent  => :ems_amazon do
+  factory :ems_amazon_with_vcr_authentication, :parent => :ems_amazon do
     zone do
       _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
       zone
     end
     after(:create) do |ems|
-      client_id         = Rails.application.secrets.amazon.try(:[], 'client_id') || 'AMAZON_CLIENT_ID'
-      client_key        = Rails.application.secrets.amazon.try(:[], 'client_secret') || 'AMAZON_CLIENT_SECRET'
+      client_id  = Rails.application.secrets.amazon.try(:[], 'client_id') || 'AMAZON_CLIENT_ID'
+      client_key = Rails.application.secrets.amazon.try(:[], 'client_secret') || 'AMAZON_CLIENT_SECRET'
 
       cred = {
         :userid   => client_id,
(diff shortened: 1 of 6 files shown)


spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -48,7 +48,7 @@ def expected_table_counts
       :floating_ip                   => 12,
       :guest_device                  => 0,
       :hardware                      => 46,
-      :miq_queue                     => 50,
```
durandom: huh? Why does this change?

Ladas: I was asking myself the same question :-)

Ladas:
@durandom packed into Factory girl

spec/models/manageiq/providers/amazon/regions_spec.rb
```diff
@@ -1,11 +1,7 @@
 describe ManageIQ::Providers::Amazon::Regions do
   it "has all the regions" do
-    EvmSpecHelper.create_guid_miq_server_zone
-    ems = FactoryGirl.create(:ems_amazon_with_authentication)
-    ems.update_authentication(:default => {
-                                :userid   => "0123456789ABCDEFGHIJ",
-                                :password => "ABCDEFGHIJKLMNO1234567890abcdefghijklmno"
-                              })
+    _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
```
durandom: Do we always need a zone? Maybe we can move this into the factory too?

Ladas: not sure, I know it was failing somewhere deep without it

spec/factories/ext_management_system.rb
```diff
@@ -5,4 +5,22 @@
           :parent  => :ems_storage do
     provider_region "us-east-1"
   end
+
+  factory :ems_amazon_with_vcr_authentication, :parent => :ems_amazon do
+    zone do
```
durandom: this is great!!!

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
durandom: ok. can you dig some deeper? I dont think we should merge it without knowing why it changed... after all these numbers are specs!

Ladas:
@durandom I would ignore the MiqQueue, several things can be writing there and that number changes a lot

durandom:
Ok, so, actually the `miq_queue` numbers dont have anything to do with the tests..
Let's remove them ✂️

durandom:
> Ok, so, actually the miq_queue numbers dont have anything to do with the tests..

@Ladas says he does this in a follow up

PR #181, opened 2017-03-16
Aws targetted refresh specs for ec2 classic vm and lb

Ladas:
Aws targetted refresh specs for ec2 classic vm and lb


spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -36,35 +35,35 @@ def assert_common
...
-      :floating_ip                   => 12,
+      :floating_ip                   => 15,
       :guest_device                  => 0,
-      :hardware                      => 46,
-      :miq_queue                     => 48,
-      :miq_template                  => 20,
-      :network                       => 14,
-      :network_port                  => 32,
+      :hardware                      => 44,
+      :miq_queue                     => 46,
```
durandom: didnt you want to remove `miq_queue`?

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -36,35 +35,35 @@ def assert_common
 
   def expected_table_counts
     {
-      :auth_private_key              => 12,
```
durandom: I'm starting to think if those numbers make sense at all.
Did you check if they are correct? So they dont provide any value.

Can we create them by running an api call?
For `:vm` it should be just `client.describe_instances.count`

PR #183, opened 2017-03-16
Add tags for VMs and Images

djberg96:
Depends on:
- [ ] https://github.com/ManageIQ/manageiq/pull/14436

This adds tags for VM's and images.

The other half of this will happen in the core repo on the save_xxx side.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
@@ -11,6 +11,7 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
     @data_index          = {}
     @known_flavors       = Set.new
     @options             = options
+    @label_tag_mapping   = ContainerLabelTagMapping.cache
   end
 
   def ems_inv_to_hashes
@@ -34,6 +35,10 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
   end
 
   private
+
+  def map_labels(model_name, labels)
+    ContainerLabelTagMapping.map_labels(@label_tag_mapping, model_name, labels)
+  end
 
   def get_flavors
     process_collection(ManageIQ::Providers::Amazon::InstanceTypes.all, :flavors) { |flavor| parse_flavor(flavor) }
@@ -176,6 +181,8 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
     name ||= $1 if location =~ /^(.+?)(\.(image|img))?\.manifest\.xml$/
     name ||= uid
 
+    labels = parse_labels(image.tags)
+
     new_result = {
       :type               => ManageIQ::Providers::Amazon::CloudManager::Template.name,
       :uid_ems            => uid,
@@ -185,7 +192,8 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
       :vendor             => "amazon",
       :raw_power_state    => "never",
       :template           => true,
-      :labels             => parse_labels(image.tags),
+      :labels             => labels,
+      :tags               => map_labels('Image', labels),
       # the is_public flag here avoids having to make an additional API call
       # per image, since we already know whether it's a public image
       :publicly_available => is_public,
@@ -224,6 +232,8 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
       :hostname  => instance.public_dns_name.presence
     }.delete_nils
 
+    labels = parse_labels(instance.tags)
+
     new_result = {
       :type                => ManageIQ::Providers::Amazon::CloudManager::Vm.name,
       :uid_ems             => uid,
@@ -232,7 +242,8 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
       :vendor              => "amazon",
       :raw_power_state     => status,
       :boot_time           => instance.launch_time,
-      :labels              => parse_labels(instance.tags),
+      :labels              => labels,
+      :tags                => map_labels('Vm', labels),
       :hardware            => {
         :bitness              => architecture_to_bitness(instance.architecture),
         :virtualization_type  => instance.virtualization_type,
--- a/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
+++ b/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
@@ -33,6 +33,7 @@ module AwsRefresherSpecCommon
     assert_specific_orchestration_template
     assert_specific_orchestration_stack
     assert_relationship_tree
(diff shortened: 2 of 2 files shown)


djberg96:
@cben, How's it look?

cben:
Looks fine, modulo "Amazon::".
Do you have other half?

djberg96:
@cben Not yet. Hope to have it in tomorrow.

djberg96:
Bah, was hoping I could get away with the mapping in the refresh parser itself. I'll have to undo this latest commit. The equivalent code just has to go in the save_inventory.rb file somewhere.

djberg96:
@cben I updated the PR and added a couple of super basic specs. Since most of the magic happens on the core repo's side, I'm not sure what more I can do here for now.

@bronaghs @blomquisg Look alright?

djberg96:
@Ladas Or do you think I should just move this logic into `save_tags_inventory`?

app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
```diff
@@ -35,6 +36,10 @@ def ems_inv_to_hashes
 
   private
 
+  def map_labels(model_name, labels)
```
cben: Consider calling this `map_aws_tag_to_miq_tags` or something clear like that.

app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
```diff
@@ -176,6 +181,8 @@ def parse_image(image, is_public)
     name ||= $1 if location =~ /^(.+?)(\.(image|img))?\.manifest\.xml$/
     name ||= uid
 
+    labels = parse_labels(image.tags)
```
cben: You're storing to custom_attributes with "section" = "labels".
IIUC the real AWS term is "tags" (right?), so perhaps you'd prefer another section name?

(everything about ContainerLabelTagMapping could use a rename, I understand you don't want big changes now, but perhaps naming the section right is better done early.  I'm pretty sure ContainerLabelTagMapping code doesn't care what section you use.)

And whatever the 'section' perhaps the `parse_labels` function should be called `parse_tags`.  Is `parse_*` convention to follow incoming API terminology or how it'll end up in DB terminology?

djberg96: Good catch on the section name, I'll update that. The parse_* convention is just an ems refresher convention that I've noticed. I don't think there's any direct connection to the DB.

app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
djberg96: I think that might be confusing, given that the method called internally is `.map_labels`, and none of the *actual* mapping happens until post.

djberg96:
@cben, ok, made some updates.

app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
```diff
@@ -192,8 +190,8 @@ def parse_image(image, is_public)
       :vendor             => "amazon",
       :raw_power_state    => "never",
       :template           => true,
-      :labels             => labels,
-      :tags               => map_labels('Image', labels),
+      :labels             => parse_labels(image.tags),
+      :tags               => map_labels('Image', parse_tags(image.tags)),
```
cben: Hmm, not sure, AFAICT it's equivalent to previous version which IMO was clearer.
`:labels => parse_labels(...)` will eventually be passed to save_labels_inventory, and will result in saving into "labels" section.  `parse_tags` here feeds same info with different :section into `map_labels()`, but `map_labels` doesn't look at section at all.

What I meant is that maybe it makes more sense to actually save AWS tags into custom_attributes with a different section e.g. "aws_tags" or "provider_tags" or whatever.  *IF* you want that, you can do it (map_labels doesn't care).

cben: BTW, my original intent for "types" that ContainerLabelTagMapping deals with was model (base) names, essentially matching the custom_attribute `resource_type` polymorphic link.
And AFAICT there is no actual model called `Image`?

In practice though you can pass an arbitrary string to `map_labels()`, it doesn't even see the custom_attribute record, it will work as long as you defined mappings expecting this arbitrary string (which you do with `Amazon::Image` in MAPPABLE_ENTITIES).

djberg96:
@cben Ok, I've returned the code to the way it was.

Regarding the `section` attribute for custom_attributes, it might help if I knew what it does exactly. Does the UI hook into it somehow? It was @bronaghs that set that, so I don't want to trample on it if she put it there for a reason.

cben:
My understand is `section` (and `source`) are just arbitrary scoping for different data to co-exist in custom_attributes table.
Ah, yes, if you change it you'd have to change the `:labels, -> { where(:section => "labels") }` scope (https://github.com/ManageIQ/manageiq/pull/14202/files)
and any UI that shows/edits `vm.labels`...

LGTM.

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -1101,4 +1103,9 @@ def assert_specific_vm_in_other_region
   def assert_relationship_tree
     expect(@ems.descendants_arranged).to match_relationship_tree({})
   end
+
+  def assert_specific_tags_on_vm
+    vm = ManageIQ::Providers::Amazon::CloudManager::Vm.where(:name => "EmsRefreshSpec-PoweredOn-Basic3").first
+    expect(vm.tags).to be_empty
+  end
```
bronaghs: This isn't really testing anything, can you pick a VM with tags? Given the feature freeze, I am fine with merging as is if a more robust test is implemented in a follow up PR.

Ladas: we should use the Vms that are testing labels

Ladas: but the specs won't pass unless it's implemented also for the graph refresh :-)

Ladas: gah, I think the fact it passes now is because we don't have the core patch merged. Then this will break since it will be blank only for the graph refresh. @djberg96 @bronaghs so we need to comment out the specs?

bronaghs: @Ladas - so this  PR needs ported to the new refresh? right? Oh...

Ladas: @bronaghs yeah, so if you want to merge it today, I would go without the refresh specs, then we would do them together with the graph refresh. But with that, I would like to evaluate performance impact too :-)

bronaghs: I agree. @djberg96 - do you agree?

djberg96: @bronaghs @Ladas Ok, commented out for now.

PR #187, opened 2017-03-17
Tasks private

durandom:
# Keep your friends close

Rails loads all engine / railtie tasks that are in `lib/tasks` - but we dont want to leak certain tasks to the main app (ManageIQ that is). E.g. `rake instance_types` creates an updates instance_types.rb just for this gem. No need to have it in ManageIQ. [Here](https://github.com/ManageIQ/manageiq-content/pull/7#issuecomment-268354339) is an example from manageiq-content

This moves tasks that are supposed to be private to the gem into `lib/tasks_private`
All tasks in there are loaded in the `Rakefile`

* Engine tasks from `lib/tasks` are __not__ loaded

## Running the specs from core?

I also moved the specs to the private tasks. First I thought it would be good to have them avail from the manageiq app - but I guess it adds more options confusion to the game. See this [discussion](https://github.com/ManageIQ/manageiq-providers-amazon/pull/175). There should be only __one__ canonical way to run the specs.

cc @Ladas @bdunne @Fryguy


Fryguy:
I like this change.  lib/tasks_private seems like a good directory name as its' clear and lives right next to the other tasks method so there's no ambiguity where something should live.  @bdunne Thoughts?

Fryguy:
@durandom I think we should still keep the `ManageIQ::Providers::Amazon::Engine.load_tasks` call, for consistency across the repos, if we *do* want to have a method that we need to expose to manageiq for some reason.  Perhaps it won't make sense in this repo, but I could see it happening in other repos (like maybe manageiq-schema or manageiq-content).  What do you think?

durandom:
> I think we should still keep the ManageIQ::Providers::Amazon::Engine.load_tasks

you mean in the `Rakefile`? Actually I would not, because of consistency. Then you have a task appearing under `app:` and at the toplevel. I guess the tasks are always either for gem use or for use in the application. If they are for use in the application, fine, they are available in the `app:` namespace - and you are remembered that they are an app task

Fryguy:
I think you are missing what I am saying.  What I would expect, as a provider author, is the following to all be true:

- All tasks that I place in `lib/tasks` will be **public** rake tasks, available from the ManageIQ application.  For example, I create a helper task to purge something in the database, meant for the field to use, but it's only applicable to the amazon provider.  I am fine if we require that provider authors namespace their rake tasks with the provider name (i.e. from manageiq directory: `rake manageiq:providers:amazon:purge_something`)
- All tasks that I place in `lib/tasks_private` will be **private** rake tasks, available only from the provider plugin.  These tasks would mostly be developer and test based tasks, such as the spec task, but also things like the create_yaml_files from manageiq-content.  We should not require provider authors to namespace their private rake tasks.

From what I can tell, bullet 1 was removed by removing the `ManageIQ::Providers::Amazon::Engine.load_tasks` call, but perhaps I am wrong?

durandom:
> From what I can tell, bullet 1 was removed by removing the ManageIQ::Providers::Amazon::Engine.load_tasks call, but perhaps I am wrong?

Nope, bullet 1 still works, because [Rails](https://github.com/rails/rails/blob/master/railties/lib/rails/engine.rb#L648) loads the engine tasks - not the engines Rakefile. 

So both bullets are true:

### lib/taks
* in engine via `rake app:<task_name>`
* in miq via `rake <task_name>

### lib/tasks_private
* in engine via `rake <task_name>
* in miq unavail

durandom:
> we require that provider authors namespace their rake tasks with the provider name

we could look into enforcing that, or provide an empty namespace in the generator

Fryguy:
> Nope, bullet 1 still works, because Rails loads the engine tasks

Awesome!

> we could look into enforcing that, or provide an empty namespace in the generator

When you create a new rails engine it does create a commented out file like that, so that would be a good idea.  Strangely, the Rails default one isn't namespaced but I think we should, and we should use the full namespace. So I'm thinking, create a file named `lib/tasks/provider_tasks.rake` with the default content of:

```ruby
# desc "Explaining what the task does"
# namespace :manageiq do
#   namespace :providers do
#     namespace :amazon do
#       # Task goes here
#     end
#   end
# end
```

(The above was modeled on the default Rails generated task, but with full namespacing)

PR #188, opened 2017-03-20
Counts specs mixin

Ladas:
Counts specs mixin, extracting specs for count to a new mixin.
Instead of hardcoded numbers, that doesn't test much, we compare
our DB to the raw data we get from the API.

### DIFF
--- a/spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
+++ b/spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
@@ -1,0 +1,273 @@
+module AwsRefresherSpecCounts
+  extend ActiveSupport::Concern
+
+  def assert_counts(expected_table_counts)
+    assert_table_counts(expected_table_counts)
+    assert_ems(expected_table_counts)
+  end
+
+  def table_counts_from_api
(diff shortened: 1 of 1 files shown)


spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
```diff
@@ -0,0 +1,273 @@
...
+  def stacks
+    hash_collection.new(aws_cloud_formation.client.describe_stacks[:stacks])
+  end
+
+  def stacks_resources
+    resources = []
+    stacks.each do |stack|
+      resources << hash_collection.new(aws_cloud_formation.client.list_stack_resources(
+        :stack_name => stack["stack_name"]
+).try(:stack_resource_summaries)
```
durandom: wow, awesome formatting :)

Ladas: I bow to the robot overlords :-)

hopefully rubocop -a will fix it later, might be an issue of the rubocop-git

durandom:
@Ladas 👏 this is going to be great... no more numbers..

PR #189, opened 2017-03-20
Refactoring of counts specs

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/188

Using new counts mixin to compare DB counts to war API data counts. This way is more accurate than comparing to fixed numbers and doesn't require constant changes when re-record the VCR.


app/models/manageiq/providers/amazon/inventory/hash_collection.rb
```diff
@@ -13,8 +13,11 @@ def all
     collection.each_with_object([]) { |item, obj| obj << transform(item) }
   end
 
-  private
+  delegate :count, to: :all
+  alias size count
 
+  # rubocop:disable all
```
durandom: why this?

PR #191, opened 2017-03-21
Rubocop -a for all Amazon specs

Ladas:
Rubocop -a for all Amazon specs

### DIFF
--- a/spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
+++ b/spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
@@ -23,16 +23,16 @@ describe ManageIQ::Providers::Amazon::CloudManager::MetricsCapture do
       }
       with_aws_stubbed(stubbed_responses) do
         expect(vm.perf_collect_metrics('realtime')).to eq([
-          {"amazon-perf-vm" => described_class::VIM_STYLE_COUNTERS},
-          {"amazon-perf-vm" => {}}
-        ])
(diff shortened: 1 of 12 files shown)


spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
```diff
@@ -23,16 +23,16 @@
       }
       with_aws_stubbed(stubbed_responses) do
         expect(vm.perf_collect_metrics('realtime')).to eq([
-          {"amazon-perf-vm" => described_class::VIM_STYLE_COUNTERS},
```
durandom: if you move the `[` to the next line - does it look nicer?

spec/models/manageiq/providers/amazon/cloud_manager/provision_workflow_spec.rb
```diff
@@ -395,7 +395,8 @@
       stub_get_next_vm_name
 
       # the dialogs populate this
-      values.merge!(:src_vm_id => template.id, :vm_tags => [])
+      values[:src_vm_id] = template.id
```
durandom: that was rubocop?

Ladas: yes

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
Ladas: can we keep this pure rubocop -a ?

PR #192, opened 2017-03-21
Adjust attribute set of newly created bucket

miha-plesko:
This PR was originally designed so that CloudObjectStoreContainer model stores `provider_region`. But a decision was made in https://github.com/ManageIQ/manageiq/pull/14242 to omit this field for now and instead query bucket region dynamically when user triggers operation. Hence this tiny adjustment.

@miq-bot assign @Ladas


Ladas:
sounds good :+1:

PR #193, opened 2017-03-22
Targeted refresh spec for vm with public ip and template

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/191

Targeted refresh spec for vm with public ip and template


spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -228,6 +228,63 @@ def assert_specific_template
...
+        :cpu_reserve           => nil,
+        :cpu_reserve_expand    => nil,
+        :cpu_limit             => nil,
+        :cpu_shares            => nil,
+        :cpu_shares_level      => nil
+      )
+    )
+
+    expect(@template2.ext_management_system).to eq(@ems)
+    expect(@template2.operating_system).to       be_nil # TODO: This should probably not be nil
```
durandom: rubocop is sometimes weird

PR #195, opened 2017-03-23
Fetch all referenced templates regardles of what we allow for collection

Ladas:
Fetch all referenced templates regardless of what we allow for collection. Meaning the template relation of a VM and therefore a VM's operating system name should never be blank. 

Implements:
https://www.pivotaltracker.com/story/show/142318659

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
@@ -21,7 +21,8 @@ class ManageIQ::Providers::Amazon::Inventory::Collector::CloudManager < ManageIQ
     @private_images_hashes ||= hash_collection.new(
       aws_ec2.client.describe_images(:owners  => [:self],
                                      :filters => [{:name   => "image-type",
-                                                   :values => ["machine"]}]).images).all
+                                                   :values => ["machine"]}]).images
+    ).all
   end
 
   def shared_images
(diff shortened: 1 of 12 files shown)


Ladas:
@miq-bot assign @durandom

app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
```diff
@@ -12,25 +12,41 @@ def availability_zones
...
+    @public_images_hashes ||= hash_collection.new(
+      aws_ec2.client.describe_images(:executable_users => [:all],
+                                     :filters          => options.public_images_filters).images).all
+  end
+
+  def referenced_images
+    return [] if extra_image_references.blank?
+
+    hash_collection.new(
+      aws_ec2.client.describe_images(:filters => [{:name => 'image-id', :values => extra_image_references}]).images
```
durandom: you if there is an api limit on how large `:values` can be?
Maybe (in the future) this will need a `.batch`ed version

app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
```diff
@@ -46,4 +65,19 @@ def stack_resources(stack_name)
   def stack_template(stack_name)
     aws_cloud_formation.client.get_template(:stack_name => stack_name).template_body
   end
+
+  private
```
durandom: yeah, like rubocop said, this is useless in a module

spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
```diff
@@ -269,4 +300,18 @@ def cloud_volumes
   def cloud_volume_snapshots
     hash_collection.new(aws_ec2.client.describe_snapshots(:owner_ids => [:self])[:snapshots])
   end
+
+  private
```
durandom: here, this is useless private

durandom:
this is great stuff

app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
Ladas: Hm, probably not, it's sent via POST. I could try how it performs for 70k refs though. :-)

PR #196, opened 2017-03-23
Add support snapshot_create to EBS cloud volume

gberginc:
The support in the backend was already there but the missing `supports`
prevented UI from showing the corresponding menu item.

This screen shows the menu item for EBS volume

screen shot 2017-03-23 at 17 30 02

@miq-bot add_label bug


Ladas:
looks good, will merge when green :+1:

PR #197, opened 2017-03-24
Use public flag we get from data rather than sending it explicitly

Ladas:
Use public flag we get from data rather than sending it explicitly,
since we don't always use the explicit way, e.g. for the targeted
refresh.


app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -247,7 +247,7 @@ def parse_image(image, is_public)
       :template              => true,
       # the is_public flag here avoids having to make an additional API call
       # per image, since we already know whether it's a public image
-      :publicly_available    => is_public,
+      :publicly_available    => image['public'],
```
durandom: can you have that tested somewhere? I don't trust booleans from external code - this could as well be a String.new("false")

Ladas: I wanted to answer, of course this is tested in the refresh and it passes. But this attribute is not tested in the refresh :-D Let me add it.

durandom:
great stuff again... amazon is going to be amazing

PR #198, opened 2017-03-24
Remap `:name` parameter into `:bucket` parameter when creating new bucket

miha-plesko:
Current implementation assumes that frontend composes all the options that backend then just sends to AWS. But there is an exception with :name option. While it is needed on frontend (to display flash), it is rejected by AWS.

Initial idea was to just remove the :name option here in backend, but turns out (see [this conversation](https://github.com/ManageIQ/manageiq-ui-classic/pull/651#discussion_r107285748)) that renaming it to :bucket works better. Hence this commit.

@miq-bot assign @Ladas

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
@@ -42,7 +42,9 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
   end
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
-    options.except!(:name) # name is part of general options, but S3 doesn't like it
+    # frontend stores name as :name, but amazon expects it as :bucket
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -42,7 +42,9 @@ def raw_cloud_object_store_container_clear
   end
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
-    options.except!(:name) # name is part of general options, but S3 doesn't like it
+    # frontend stores name as :name, but amazon expects it as :bucket
+    options[:bucket] = options.delete(:name) unless options[:name].nil?
```
Ladas: hm, it's strange it expects it as :bucket but returns bucket.name

Ladas:
@miha-plesko looks good, it's a weird naming on the aws side though :-)

miha-plesko:
Yeah, I'm not smart enough to understand it :D

PR #199, opened 2017-03-24
move amazon authenticator from core to amazon repo

durandom:
This was the only code still using `aws-sdk` directly. Moving over here

https://github.com/ManageIQ/manageiq/pull/14492

cc @chessbyte


durandom:
added specs for .validate_config in last commit. This moves the amazon spec from https://github.com/ManageIQ/manageiq/pull/14492 over here

durandom:
@blomquisg merge this one too? To get amazon auth back in...

PR #213, opened 2017-04-17
Make S3 a prototype feature

blomquisg:
Add a configuration to the Settings.yml file that allows administrators to enable or disable Amazon S3 support.

The feature will be disabled by default.

Disabling S3 will disable the S3 manager, which will prevent the S3 refresh worker from starting.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -69,3 +69,6 @@
         :ems_refresh_worker_amazon_network: {}
         :ems_refresh_worker_amazon_ebs_storage: {}
         :ems_refresh_worker_amazon_s3: {}
+:prototype:
+  :amazon:
+    :s3: false
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -80,8 +80,10 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
     build_ebs_storage_manager unless ebs_storage_manager
     ebs_storage_manager.name = "#{name} EBS Storage Manager"
 
-    build_s3_storage_manager unless s3_storage_manager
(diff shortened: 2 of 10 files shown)


Ladas:
@blomquisg hm, this will prevent it from being created. Problem is that we can't add it then from the UI, so that is probably an RFE for later.

blomquisg:
Looks like I have to add the Settings check to all of the tests, too...

Ladas:
@blomquisg hm, you can probably just change the count of :ext_management_systems and only place where we need the setting enabled is https://github.com/Ladas/manageiq-providers-amazon/blob/05c2d067bbe36fba395211f5a9dd27bae7afdeec/spec/models/manageiq/providers/amazon/storage_manager/s3/stubbed_refresher_spec.rb

config/settings.yml
```diff
@@ -69,3 +69,6 @@
         :ems_refresh_worker_amazon_network: {}
         :ems_refresh_worker_amazon_ebs_storage: {}
         :ems_refresh_worker_amazon_s3: {}
+:prototype:
```
durandom: note that settings will be merged with the core configs...
is there already a top level prototype key?
otherwise I'd put it under`ems -> ems_amazon -> s3_prototype` e.g.

blomquisg: Yeah, there's already a prototype key in core settings.

blomquisg:
@Ladas cops updated, and tests passing.

simaishi:
Fine backport details:

```
$ git log -1
commit 0a8f3b8ac870a4c383a427d646975eb5ac672528
Date:   Tue Apr 25 13:04:41 2017 +0200

    Merge pull request #213 from blomquisg/amazon_s3_prototype
    
    Make S3 a prototype feature
    (cherry picked from commit 2b92588b6de01b14521ebb6d7b02f669feba82c1)
```

PR #226, opened 2017-04-24
Support deletion of S3 folders

miha-plesko:
On S3, folders are nothing but regular objects. However, deleting folder fails silently if there are any objects in it. To resolve this issue, we perform a batch delete of all objects within that folder.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1444079

@miq-bot assign @bronaghs 
@miq-bot add_label bug

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
@@ -18,6 +18,10 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreObject < 
   end
 
   def raw_delete
-    with_provider_object(&:delete)
+    if key.end_with? "/" # delete object with subobjects (aka. folder)
+      cloud_object_store_container.provider_object.objects(:prefix => key).batch_delete!
+    else # delete single object
+      with_provider_object(&:delete)
+    end
   end
 end
--- a/spec/models/manageiq/providers/amazon/storage_manager/s3/stubbed_refresher_spec.rb
+++ b/spec/models/manageiq/providers/amazon/storage_manager/s3/stubbed_refresher_spec.rb
@@ -255,6 +255,62 @@ describe ManageIQ::Providers::Amazon::StorageManager::S3::Refresher do
         expect(object.supports?(:delete)).to be_falsey
       end
     end
+
+    describe "delete folder" do
+      before do
+        allow(object.cloud_object_store_container).to receive(:provider_object)
+          .and_return(provider_object_for_container)
+      end
+
+      let :provider_object_for_container do
+        container = double("provider_object_for_container")
+        allow(container).to receive(:objects).and_return(objects_batch)
+        allow(container).to receive(:object).and_return(single_object)
+        container
+      end
+
+      let :objects_batch do
+        objects_batch = double("objects_batch")
+        allow(objects_batch).to receive(:batch_delete!)
+        objects_batch
+      end
+
+      let :single_object do
+        single_object = double("single_object")
+        allow(single_object).to receive(:delete)
+        single_object
+      end
+
+      it "delete folder #1" do
+        object.key = "myfolder/"
+        expect(objects_batch).to receive(:batch_delete!)
+        expect(single_object).not_to receive(:delete)
+
+        with_aws_stubbed(stub_responses) do
+          object.delete_cloud_object_store_object
+        end
+      end
+
+      it "delete folder #2" do
+        object.key = "myfolder/subfolder/"
+        expect(objects_batch).to receive(:batch_delete!)
+        expect(single_object).not_to receive(:delete)
+
+        with_aws_stubbed(stub_responses) do
+          object.delete_cloud_object_store_object
+        end
+      end
+
+      it "delete regular object" do
+        object.key = "file.txt"
+        expect(objects_batch).not_to receive(:batch_delete!)
+        expect(single_object).to receive(:delete)
+
+        with_aws_stubbed(stub_responses) do
+          object.delete_cloud_object_store_object
+        end
+      end
+    end
   end
 
   def refresh_spec


bronaghs:
@miha-plesko - is it possible to ask specs for this?

Ladas:
@miha-plesko could we add at least a small spec for this?

Ladas:
@bronaghs hehe, didn't notice your comment :-)

bronaghs:
@ladas - yeah, I realized this could do with some specs the second I hit the approve button which was a bit misleading. I changed my review status.

miha-plesko:
@bronaghs @Ladas thanks for complaining about the missing tests. I did see your message in the morning already but took me some time to figure how to implement them. Thanks you @gberginc for informing me about `double` structure that came really handy here!

@bronaghs I'm looking forward to be hearing your opinion on the tests - is this the way to go in such case?

Ladas:
specs look great 👍

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
```diff
@@ -18,6 +18,10 @@ def provider_object(connection = nil)
   end
 
   def raw_delete
-    with_provider_object(&:delete)
+    if key.end_with? "/" # delete object with subobjects (aka. folder)
```
gberginc: Although `/` is probably the most common one, I think this is not completely correct. To my knowledge S3 does not have a concept of directories - the fact that the console supports "creating" them and shows them hierarchically is merely a presentation detail simplifying the workflow.

Experiment I just did. In the S3 console I created a "folder" named `testing` and uploaded a file into it. Listing the buckets using S3 client returns (I truncated irrelevant info)

```ruby
Aws> s3.list_objects({bucket: "miha-oregon"})
[Aws::S3::Client 200 0.943381 0 retries] list_objects(bucket:"miha-oregon",encoding_type:"url")
=> #<struct Aws::S3::Types::ListObjectsOutput
 contents=
  [#<struct Aws::S3::Types::Object
    key="testing/",
    storage_class="STANDARD",
   #<struct Aws::S3::Types::Object
    key="testing/test.rtf",
    storage_class="STANDARD",
```

There are two objects. The fact that the first one is named `testing/` does not make it a parent of `testing/test.rtf`. If I now delete `testing/`

```ruby
Aws> s3.delete_object({bucket: "miha-oregon", key: "testing/"})
[Aws::S3::Client 204 0.879845 0 retries] delete_object(bucket:"miha-oregon",key:"testing/")
=> #<struct Aws::S3::Types::DeleteObjectOutput delete_marker=nil, version_id=nil, request_charged=nil>
```

and list the objects again:

```ruby
Aws> s3.list_objects({bucket: "miha-oregon"})
[Aws::S3::Client 200 1.142258 0 retries] list_objects(bucket:"miha-oregon",encoding_type:"url")
=> #<struct Aws::S3::Types::ListObjectsOutput
 contents=
  [#<struct Aws::S3::Types::Object
    key="testing/test.rtf",
    storage_class="STANDARD",
```

Notice that `testing/` is in fact gone, but `testing/test.rtf` is still there. If you go into the S3 console, you will still see the "folder" and the file within.

This is also the reason why `list_objects` has an additional attribute `delimiter` [doc](http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html#list_objects-instance_method).

Honestly, I don't know if this is relevant for this function, but it's important to understand this.

bronaghs:
@miha-plesko - thanks @miha-plesko  for adding the specs. I am just waiting to hear if I can merge this or if we are already in the 'blockers only' phase.

bronaghs:
Ignore my comment above:

> " I am just waiting to hear if I can merge this or if we are already in the 'blockers only' phase."

I had a pang of confusion, non blockers can be merged to master but not the Fine branch.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
miha-plesko: Thank you for opening this debase @gberginc. Please see my answer below - bottomline is that I would still merge this PR.

Reading the user documentation here [Listing Keys Hierarchically Using a Prefix and Delimiter](http://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html) I can agree to what you've found out @gberginc, objects are generally not in parent-children relation. On the other hand I also agree to Matouš who has reported the bug on Bugzilla: if S3 console mimics the parent-children behavior assuming `/` separator, why would we do it any different in MiQ?

But then based on [Working with Folders](http://docs.aws.amazon.com/AmazonS3/latest/UG/FolderOperations.html) I woud say that it is perfectly safe to merge this PR. The link says that object whose name ends with `/` actually **is** different from other objects:
```
An object named with a trailing "/" displays as a folder in the Amazon S3 console.
The Amazon S3 console does not display the content and metadata for such an object.
When copying an object named with a trailing "/" by using the Amazon S3 console, a
new folder is created in the destination location but the object's data and metadata
are not copied. 
```

@bronaghs, @Ladas may I ask for your opinion?

miha-plesko:
@bronaghs don't worry, we all have our moments 😄

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_object.rb
gberginc: You are correct about this. When looking from the perspective of UI, it makes perfect sense. I was coming more from the provider perspective. For example, if this is called from the automate or miq API, what will the user expect - to delete the object itself (as S3 API does) or the entire hierarchy (as the UI does)?

That being said, I also don't mind merging this, but keep this in mind for future reference.

simaishi:
Fine backport details:

```
$ git log -1
commit 51c9e182a38088c4a3aa805c02addcecedf63347
Date:   Tue May 2 09:30:52 2017 -0400

    Merge pull request #226 from miha-plesko/delete_s3_folder
    
    Support deletion of S3 folders
    (cherry picked from commit 5cb3108a20eb8a2f775dab75e07c8797152a9a55)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1459258
```

PR #227, opened 2017-04-24
Rewrite S3 parser to a new format

Ladas:
Rewrite S3 parser to a new format


Ladas:
@gberginc new format of the graph refresh parser

gberginc:
@durandom Based on the PR below I assume, this should be fine/yes. It will be disabled by default, but present in this release.

https://github.com/ManageIQ/manageiq-providers-amazon/pull/213

/cc @miha-plesko

Ladas:
@gberginc only blockers can got to Fine now :-( So Fine has the new format of the parser only for the CloudManager, I've almost made it all.

PR #228, opened 2017-04-25
Rewrite EBS parser to a new format

Ladas:
Rewrite EBS parser to a new format


Ladas:
cc @gberginc @miha-plesko

PR #236, opened 2017-04-25
Changelog 59 Fine Beta-2

mfeifer:
Update RC-2 to Beta-2

[skip ci]


mfeifer:
@miq-bot add_label documentation

Ladas:
@mfeifer should this go directly to a Fine, or we want to backport it? I am asking since it doesn't look like the normal changelog :-)

mfeifer:
@Ladas good question.  Once we start releasing, the items move to the roll up UNLESS the things are not backported. So, there are two that look pretty much the same one for master and one for fine because all the items that go into the changelog are winding up in fine beta.

PR #242, opened 2017-04-26
Cleanup storage managers parsers

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/239

Cleanup storage managers parsers


Ladas:
cc @miha-plesko @gberginc and even more storage parsers cleanup (last commit)

PR #243, opened 2017-05-02
Wrap the validation Settings.ems call in a lambda

jrafanie:
Regions.names calls regions which calls

`Settings.ems.ems_amazon.try!(:additional_regions)).stringify_keys`

This Settings call was occurring at require time. We should not hit
Settings by just requiring files and that is what happens
previously.  This seems to cause a sporadic test failure [1]
perhaps because resetting the global constant Settings is very time
dependent and if we access it at the wrong time, it could be nil/not
initialized.

We can prevent this by delaying the Settings call until the validation
is called by wrapping the code in a lambda.

```
  1) Metric as vmware with enabled and disabled targets executing perf_capture_timer should queue up enabled targets
     Failure/Error: if (_klass = name.safe_constantize) # this triggers the load

     NoMethodError:
       undefined method `ems_amazon' for nil:NilClass
     # ./lib/extensions/descendant_loader.rb:234:in `load_subclasses'
     # ./lib/extensions/descendant_loader.rb:254:in `descendants'
     # ./lib/extensions/ar_virtual.rb:593:in `preloaders_for_one'
     # ./lib/miq_preloader.rb:4:in `preload'
     # ./app/models/metric/targets.rb:68:in `load_infra_targets_data'
     # ./app/models/metric/targets.rb:12:in `capture_infra_targets'
     # ./app/models/metric/targets.rb:135:in `capture_targets'
     # ./app/models/metric/capture.rb:46:in `perf_capture_timer'
     # ./spec/models/metric_spec.rb:61:in `block (5 levels) in <top (required)>'
```

### DIFF
--- a/app/models/manageiq/providers/amazon/manager_mixin.rb
+++ b/app/models/manageiq/providers/amazon/manager_mixin.rb
@@ -2,7 +2,7 @@ module ManageIQ::Providers::Amazon::ManagerMixin
   extend ActiveSupport::Concern
 
   included do
-    validates :provider_region, :inclusion => {:in => ManageIQ::Providers::Amazon::Regions.names}
+    validates :provider_region, :inclusion => {:in => ->(_region) { ManageIQ::Providers::Amazon::Regions.names }}
   end
 
   def description


durandom:
Is this only for sporadic spec failures or is this a general race condition?

jrafanie:
> Is this only for sporadic spec failures or is this a general race condition?

@durandom it happens for a specific test sequence so it's definitely repeatable but since tests run randomly it manifests as a sporadic test failure

durandom:
I think the error lies in [stub_settings](https://github.com/ManageIQ/manageiq/blob/acb8c764c6a27ba53984ea4165624096a7aac19a/spec/support/settings_helper.rb#L1)

This removes all other settings whereas [stub_settings_merge](https://github.com/ManageIQ/manageiq/blob/acb8c764c6a27ba53984ea4165624096a7aac19a/spec/support/settings_helper.rb#L7) augments the settings.

durandom:
https://github.com/ManageIQ/manageiq/pull/14974 is another attempt to fix this

durandom:
@jrafanie wdyt about https://github.com/ManageIQ/manageiq/pull/14974 ? 
While your fix here somehow works around this issue, it also will hit `Settings` for every validation of the manager. Which happens at least for refresh. 

OTOH it also removes the need to reboot the server when adding additional regions via settings -
 which is another hidden bug 😄 

I would merge both, but this here for making regions not static at boot time.

jrafanie:
I think we should merge both @durandom.  My fix just prevents the bug from happening, it doesn't fix the problem.

Fryguy:
Even so, let's say there was another test that modified the disabled_regions list, and then wanted to verify it.  In that case, since the Settings were one-time-set at file require time, the test would still fail.  I think that this PR is still worthy.

> While your fix here somehow works around this issue, it also will hit Settings for every validation of the manager

That's not very expensive at all.

### Before

```ruby
Benchmark.ips do |x|
  x.report("before") { ManageIQ::Providers::Amazon::CloudManager.new.valid? }
end
Warming up --------------------------------------
              before    16.000  i/100ms
Calculating -------------------------------------
              before    227.269  (± 4.4%) i/s -      1.136k in   5.009078s
```

### After

```ruby
Benchmark.ips do |x|
  x.report("after") { ManageIQ::Providers::Amazon::CloudManager.new.valid? }
end
Warming up --------------------------------------
              after    16.000  i/100ms
Calculating -------------------------------------
              after    226.792  (± 4.9%) i/s -      1.136k in   5.021210s
```

### Settings.* performance

```ruby
Benchmark.ips do |x|
  x.report("settings") { Settings.ems.ems_amazon }
end
Warming up --------------------------------------
            settings   187.796k i/100ms
Calculating -------------------------------------
            settings      3.557M (± 7.4%) i/s -     17.841M in   5.043957s
```

durandom:
😄 for the benchmarks

PR #244, opened 2017-05-03
Refresh port and floating ip for a lb

Ladas:
Refresh port and floating ip for a LB(both old and new refresh) + specs and new VCRs

Related to BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1387616


simaishi:
@Ladas this doesn't need to be `fine/yes` for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1387616?

Ladas:
@simaishi the BZ says Azure. So we would need BZ for AWS?

simaishi:
@Ladas Ah, that's true. I'd think a separate BZ would make it easier to track changes.

PR #245, opened 2017-05-03
bump ruby to 2.3.3 and add 2.4.1

durandom:
cc @jrafanie 
same as https://github.com/ManageIQ/manageiq-providers-lenovo/pull/36


jrafanie:
@durandom @Fryguy I don't merge rights here but LGTM

PR #246, opened 2017-05-05
Parse timestamp for aws alarm events

Ladas:
Parse correct timestamp for AWS Alarm event

Test that we parse a timestamp for each event type, timestamp is needed for being able to display event in the timelines, also an event without timestamp can break timelines, if it's first or last event of the provider.


durandom:
@Ladas I think we need a BZ for this

Ladas:
@simaishi could you create a BZ for this?

simaishi:
@Ladas We ask the developers to create a BZ so BZ will have accurate and detailed information that I will not be able to provide 😄

simaishi:
@Ladas ping, is this still needed in Fine branch?

Ladas:
@simaishi I am not sure, @bronaghs what do you think? Looks like I should have created a BZ long time ago. :-)

bronaghs:
I think it should be backported.

PR #247, opened 2017-05-05
Add CloudWatch events for VM to a timeline

Ladas:
Add CloudWatch events for VM to a timeline

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -8,11 +8,21 @@
           - AWS_EC2_Instance_CREATE
           - AWS_EC2_Instance_UPDATE
           - AWS_EC2_Instance_DELETE
+          - TerminateInstances
+          - RunInstances
         :power:
           :critical:
           - AWS_EC2_Instance_running
           - AWS_EC2_Instance_shutting-down
           - AWS_EC2_Instance_stopped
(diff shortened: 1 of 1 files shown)


durandom:
@Ladas @bronaghs  @blomquisg I'm having a hard time extracting all events in settings.yml into the provider repos, because the event names are generic and there is no way to attribute them to a certain provider.
I like how the event are prefixed with `AWS_EC2_` - should'nt we do this with the cloudwatch and other sources as well?

bronaghs:
@durandom - I see what you mean. The thing it, once extracted to each provider repo there is no need for the prefix so it becomes an exercise to help with the extraction process only. Is that right? Or is it helpful in the timelines too?

Ladas:
@durandom @bronaghs make sense, is there still need for the prefix when we have it extracted? The prefix might be needed for other processing though, but all events have prefix except the API based events, so it should be good.

durandom:
yep, you are right. I thought people view events in the timeline across providers, but I think thats not even possible, it's always in the context of an EMS or other object. So, cool and sorry for the confusion 🎊

Ladas:
@durandom in the context of more providers, there is :source attribute on EmsEvent, to distinct a provider.

PR #249, opened 2017-05-09
Prevent volume update with no modifications

gberginc:
Amazon EBS API will throw an error in case `modify_volume` is called
with the options that do not differ from the current volume
configuration. This patch ensures that data is properly tested before
it's submitted to the API (if at all). Specs needed minor changes due to
this change; one more test was added to ensure that "no modifications
update" do not invoke the API.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1449237

@miq-bot add_label bug

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
@@ -100,8 +100,9 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
 
   def modify_volume_options(options = {})
     modify_opts = {}
+
     if volume_type != 'standard'
-      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != 'standard'
+      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != volume_type
       modify_opts[:size]        = Integer(options[:size]) if options[:size] && Integer(options[:size]).gigabytes != size
       modify_opts[:iops]        = options[:iops] if (options[:volume_type] == "io1" || volume_type == 'io1') && options[:iops]
     end
--- a/spec/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_spec.rb
+++ b/spec/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_spec.rb
@@ -75,7 +75,7 @@ describe ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume do
 
         with_aws_stubbed(stubbed_responses) do
           expect do
-            cloud_volume.update_volume(:volume_type => 'gp2')
+            cloud_volume.update_volume(:volume_type => 'gp2', :size => 4)
           end.to raise_error(MiqException::MiqVolumeUpdateError)
         end
       end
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -100,8 +100,9 @@ def provider_object(connection = nil)
 
   def modify_volume_options(options = {})
     modify_opts = {}
+
     if volume_type != 'standard'
-      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != 'standard'
+      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != volume_type
       modify_opts[:size]        = Integer(options[:size]) if options[:size] && Integer(options[:size]).gigabytes != size
```
bronaghs: @gberginc -Looks like this will allow non-standard volume types to be updated with a 'standard' volume type which wasn't allowed before, is that right?

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -100,8 +100,9 @@ def provider_object(connection = nil)
 
   def modify_volume_options(options = {})
     modify_opts = {}
+
     if volume_type != 'standard'
-      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != 'standard'
+      modify_opts[:volume_type] = options[:volume_type] if options[:volume_type] && options[:volume_type] != volume_type
```
bronaghs: @gberginc - Looks like ``` options[:volume_type] if options[:volume_type]```  can be simplified by using the 'try' method

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
gberginc: You are correct; I wondered about this check if it makes sense or not here. The UI prevents setting standard while editing. On the other hand, when updating the volume directly through the provider, it may be better to actually get an exception if you tried to set `standard` (and it wasn't `standard` before).

For example, if I have a `gp2` volume and call `update_volume` with new `options[:volume_type] = 'standard'`, it may be better to let this through and let AWS exception.

What do you think? I can add the check for `standard` again.

bronaghs: This code can be called from other places in addition to the UI eg the REST API so I think we should handle the exception AWS throws. Looks like you are already handling this [here](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb#L47). Is that right?

gberginc: There I wrap the exception received from AWS API, yes. Prior to this patch, the API call would not even be made when `standard` volume type is requested. This means that the update would actually success (from MIQ perspective) because AWS API was not made.

Other providers do not actually care what data is being passed to the method - it's more or less verbatim passed to the actual provider API. For example, [here](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/cloud_volume.rb#L36-L39) is the update for OpenStack.

bronaghs: Given that you are gracefully handling the exception, I think the code is fine as it is.

gberginc: ok, I'll replace the `try` above and update the PR

gberginc:
We agreed in the Gitter that we do not change to `try` as it brings other complications.

If there's anything else regarding this PR, let me know.

simaishi:
Fine backport details:

```
$ git log -1
commit b0d8314ac89484931b06f7556b34c3d141a7b4be
Date:   Wed May 10 15:39:36 2017 -0400

    Merge pull request #249 from gberginc/fix_cloud_volume_update
    
    Prevent volume update with no modifications
    (cherry picked from commit e3579a00d7dd04df3a275d71788cb6eeb715c499)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1459940
```

PR #253, opened 2017-05-19
Leverage builder params to provide ems id and defaults

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/15142

Leverage builder params to provide ems_id and the defaults. That will simplify parsers logic and allow us to create placeholder objects using just ems_ref.


durandom:
@Ladas is there a PR on core needed? I tried cleaning the travis caches - no luck. Maybe `git commit --amend` ?

PR #256, opened 2017-05-26
Require rails/engine and action_controller/railtie

jrafanie:
Now, we don't require anything and I get to delete code.

~~Because we use isolate_namespace, we need to require action_controller/railtie. A blank rails engine generated by rails doesn't require anything...[1] Other engines do the sensible thing: require what they need [2] We shouldn't make assumptions about the rails features the application using the engine will want to use.~~

~~[1]
https://github.com/rails/rails/blob/0e453e9150f0ac9fa2bdff4511c3c806a3e45f49/railties/lib/rails/generators/rails/plugin/templates/lib/%25namespaced_name%25/engine.rb~~

~~[2] https://github.com/rails/coffee-rails/blob/db0659924c82f24e2665d799b1ab342078f2501d/lib/coffee/rails/engine.rb#L1~~

Note, I tested this by running the following:
```
05:50:14 ~/Code/manageiq-providers-amazon (dont_require_all_of_rails) (2.4.1) - be rake -T | grep instance_types
rake instance_types                                              # Create new instance_types.rb
```

I tested this before this change, dropped the rails/all and added back things until I could load the rake task.

EDIT: MORE 🔥 🚽 ✂️

### DIFF
--- a/lib/manageiq/providers/amazon/engine.rb
+++ b/lib/manageiq/providers/amazon/engine.rb
@@ -1,5 +1,3 @@
-require 'rails/all'
-
 module ManageIQ
   module Providers
     module Amazon
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,5 @@
 require 'bundler/setup'
 require 'bundler/gem_tasks'
-require 'manageiq-providers-amazon'
 
 begin
   require 'rspec/core/rake_task'


jrafanie:
@durandom please review

lib/manageiq/providers/amazon/engine.rb
```diff
@@ -1,4 +1,5 @@
-require 'rails/all'
+require 'rails/engine'
+require 'action_controller/railtie' # needed for isolate_namespace below
```
durandom: How does https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/manageiq/automation_engine/engine.rb get away without requiring that railtie?

lib/manageiq/providers/amazon/engine.rb
```diff
@@ -1,4 +1,5 @@
-require 'rails/all'
```
durandom: This was introduced in https://github.com/ManageIQ/manageiq-providers-amazon/pull/170/commits/69db2fcd84644b12125c2e680214e959bc323f77
The original and generated engine.rb does not require `rails/all`

durandom:
```diff
diff --git a/Rakefile b/Rakefile
index 8a48792..29e5b6b 100644
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,5 @@
 require 'bundler/setup'
 require 'bundler/gem_tasks'
-require 'manageiq-providers-amazon'
 
 begin
   require 'rspec/core/rake_task'
diff --git a/lib/manageiq/providers/amazon/engine.rb b/lib/manageiq/providers/amazon/engine.rb
index 75be854..6dfcdbd 100644
--- a/lib/manageiq/providers/amazon/engine.rb
+++ b/lib/manageiq/providers/amazon/engine.rb
@@ -1,5 +1,3 @@
-require 'rails/all'
-
 module ManageIQ
   module Providers
     module Amazon
```

durandom:
@jrafanie or do you want to `require 'rails/engine'` because the rails app that loads the engine might not have `rails/engine` loaded? Can that even be?

lib/manageiq/providers/amazon/engine.rb
jrafanie: @durandom probably because other engines require rails/all

durandom: ah, sounds like a dependency race condition is lurking there...

as said, I dont think we need to require anything here, thats how we started off at least

jrafanie:
> I think we should get away with requiring nothing at all, by removing https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/Rakefile#L3
> 
> See the original commit that introduced it.
> 
> Now we're loading rake tasks via load in the Rakefile

@durandom are you saying we don't need to add the instance_types rake task?

durandom:
> are you saying we don't need to add the instance_types rake task?

@jrafanie we still need the task. And with the [patch above](https://github.com/ManageIQ/manageiq-providers-amazon/pull/256#issuecomment-304615889) applied the task is still there.

So, unless you want to specifically require `rails/engine`, you could cut down to the requires to :zero:

jrafanie:
Ok, updated @durandom

lib/manageiq/providers/amazon/engine.rb
```diff
@@ -1,4 +1,4 @@
...
+require 'action_controller/railtie' # needed for isolate_namespace below
```
chrisarcand: @jrafanie ??? `isolate_namespace` isn't ActionController; you're just getting it as a side effect.

This should be `rails/engine`, though I note you already mention that above sorta. Huh? I don't get the reasoning for not doing that.

jrafanie: I'll double check but `isolate_namespace` was undefined without action_controller/railtie.  

Also, `rails/engine` is autoloaded by lots of gems and even the scaffold generator for 'engine' [here](https://github.com/rails/rails/blob/0e453e9150f0ac9fa2bdff4511c3c806a3e45f49/railties/lib/rails/generators/rails/plugin/templates/lib/%25namespaced_name%25/engine.rb).

No worries, I'll review it again and update this PR.

durandom:
@jrafanie as said in https://github.com/ManageIQ/manageiq-providers-amazon/pull/256#issuecomment-304615889 can't we do without requiring anything in `engine.rb`

chrisarcand:
> can't we do without requiring anything in engine.rb

I was thinking that it made sense in a purely Ruby angle that one should have to ensure the engine code is available (even though by now in a normal Rails boot it would be loaded) but it looks like Rails convention is that you just assume that. Shame on me for thinking you'd ever execute an engine standalone 🙄 

LGTM 👍

jrafanie:
> Shame on me for thinking you'd ever execute an engine standalone 🙄

LOL, yup

jrafanie:
ok, @durandom updated

PR #260, opened 2017-06-12
Add notification for snapshot delete failure

sasoc:
Deleting snapshot initiates the action and after there is no notification if it fails. This changes add notification in case of failure.

Recording of notification example: http://k00.fr/jbn3j

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1449243

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
@@ -81,7 +81,18 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolumeSnapshot < ::
   def delete_snapshot(_options = {})
     with_provider_object do |snapshot|
       if snapshot
-        snapshot.delete
+        begin
+          snapshot.delete
+        rescue => e
+          notification_options = {
+            :snapshot_op => 'delete',
+            :subject     => "[#{name}]",
+            :error       => e.to_s
+          }
+
+          Notification.create(:type => :vm_snapshot_failure, :options => notification_options)
+          raise MiqException::MiqVolumeSnapshotDeleteError, e.to_s
+        end
       else
         _log.warn "snapshot=[#{name}] already deleted"
       end


sasoc:
cc @gberginc

sasoc:
I used vm_snapshot_failure for notification_type and I'm not sure its the right one. It looks the best one from existing types. Adding new type requires changes in core project.

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
```diff
@@ -81,7 +81,17 @@ def delete_snapshot_queue(userid = "system", _options = {})
...
+        begin
+          snapshot.delete
+        rescue => e
+          notification_options = {
+            :snapshot_op => 'delete',
+            :subject     => "[#{name}]",
+            :error       => e.to_s
+          }
+
+          Notification.create(:type => :vm_snapshot_failure, :options => notification_options)
```
gberginc: I think the problem here is that you are actually hiding the exception from the caller. Previously, in case of an error, this was propagated up the call stack. Now, only the notification is thrown, and while this will resolve the issue we had in the UI, it does not handle the case when this function is called somewhere else.

Raise this error: https://github.com/ManageIQ/manageiq-gems-pending/blob/476478cb5aaec1139f052269a9756e0fcd7c4668/lib/gems/pending/util/miq-exception.rb#L130.

Here's an example I've shown and also raises the error.

gberginc:
@AparnaKerve We discussed this could be a way to resolve the problem of async snapshot removal. Was our interpretation correct?

I think in the long run, the [CloudVolumeSnapshot](https://github.com/ManageIQ/manageiq/blob/571fca19551c09fda90eae891fd067de0154cb3b/app/models/cloud_volume_snapshot.rb) should expose its operations like other models, like [CloudVolume](https://github.com/ManageIQ/manageiq/blob/571fca19551c09fda90eae891fd067de0154cb3b/app/models/cloud_volume.rb). Then we should have this kind of notifications in the core model (`delete_snapshot`) and do the actual delete in `raw_delete_snapshot`.

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
Ladas: yeah, we either need to raise exception or log an error. @gberginc is right, raising MiqVolumeSnapshotDeleteError will be the best.

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
```diff
@@ -81,7 +81,17 @@ def delete_snapshot_queue(userid = "system", _options = {})
   def delete_snapshot(_options = {})
     with_provider_object do |snapshot|
       if snapshot
-        snapshot.delete
+        begin
+          snapshot.delete
+        rescue => e
+          notification_options = {
+            :snapshot_op => 'delete',
+            :subject     => "[#{name}]",
```
Ladas: do:
:subject     => self,

AparnaKarve:
> @AparnaKarve We discussed this could be a way to resolve the problem of async snapshot removal. Was our interpretation correct?

Yes, @gberginc The current implementation with `Notification.create` is exactly what I meant.

> I think in the long run, the CloudVolumeSnapshot should expose its operations like other models, like CloudVolume. Then we should have this kind of notifications in the core model (delete_snapshot) and do the actual delete in raw_delete_snapshot.

Agree. If we could add the notifications in the core model, that would be better.
To expose operations for `CloudVolumeSnapshot` like we have for `CloudVolume`, would we need any migrations?

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
sasoc: This is what message looks like if I use self for subject: 
screen shot 2017-06-19 at 14 53 05

So for now I left  ``"[#{name}]"``

gberginc:
@Ladas can you check #277 and the corresponding PR in the core? I think those two will be more generic.

bronaghs:
Hi @gberginc - Is this PR dependent on https://github.com/ManageIQ/manageiq-providers-amazon/pull/277 and therefore a WIP? Or is this good to go?

gberginc:
@bronaghs this should actually be closed as now that we have the patch in core merged, #277 resolves this better.

I can’t close it. Can you or @sasoc close it, please?

PR #264, opened 2017-06-27
allow raw_connect to validate the connection and decrypt the password

jntullo:
The UI is moving towards validating credentials on the queue to allow for validating credentials in separate zones utilizing the `raw_connect` class method, rather than needing to create a temporary EMS for validation.

 For many of the providers, `raw_connect` ensures the credentials are correct. This updates `raw_connect` to not only return a connection but also be able to validate (it will *not* validate by default).

Because credentials are going to be stored on the queue, it also needs to be able to decrypt them. This uses `MiqPassword.try_decrypt` so that it can accept both unencrypted and encrypted credentials.

PR with greater detail: https://github.com/ManageIQ/manageiq-ui-classic/pull/1580


jntullo:
@miq-bot assign @blomquisg

PR #265, opened 2017-06-28
Refactor the regions_by_hostname method to be simpler.

bronaghs:
Refactor the regions_by_hostname method to be simpler.


djberg96:
The failures appear to be some sort of unrelated bundler issue.

PR #267, opened 2017-06-29
Handle SSA Response from AWS Scanning Instance

jerryk55:
Previously the request to perform SSA on AWS by the Scanning Instance was implemented.
We are now implementing the code to process the response from the the SSA scan upon
completion.

Note that much of the code to do this, in perform_metadata_sync,
was taken from the ScanningMixin method scan_via_miq_vm.  The intent will be to
mitigate duplication when possible moving forward but for now we wanted to get
this running at the very least.

Also note that for now this code will be reading the first available response from
SQS.  If more than one SSA request is issued simultaneously there is a good chance
that the responses will be handled out of order.  We intend to implement a worker
to perform this task in the near future to handle this issue.

@roliveri and @hsong-rh please review for relevance with regard to SSA.
@bronaghs please review and or assign appropriately to a Providers team member for merging
when satisfied.  Thanks.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/vm_or_template_shared/scanning.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/vm_or_template_shared/scanning.rb
@@ -25,11 +25,11 @@ module ManageIQ::Providers::Amazon::CloudManager::VmOrTemplateShared::Scanning
     connect_args[:service] = :S3
     @s3                    = ext_management_system.connect(connect_args)
     raise "Unable to obtain a new S3 resource" unless @s3
-    ssaq_args                 = {}
-    ssaq_args[:ssa_bucket]    = "evm-prototype"
-    ssaq_args[:region]        = ext_management_system.provider_region
-    ssaq_args[:sqs]           = @sqs
-    ssaq_args[:s3]            = @s3
+    ssaq_args              = {}
+    ssaq_args[:ssa_bucket] = "evm-prototype"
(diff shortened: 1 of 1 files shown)


jerryk55:
@roliveri @bronaghs can someone review and merge?  The Travis issue that was occurring before my PTO seems to have been fixed. Thanks much.

app/models/manageiq/providers/amazon/cloud_manager/vm_or_template_shared/scanning.rb
```diff
@@ -24,21 +26,130 @@ def perform_metadata_scan(ost)
     @s3                    = ext_management_system.connect(connect_args)
     raise "Unable to obtain a new S3 resource" unless @s3
     ssaq_args                 = {}
-    ssaq_args[:ssa_bucket]    = name
+    ssaq_args[:ssa_bucket]    = "evm-prototype"
```
roliveri: Is this a temporary bucket name?

jerryk55: Uh- yes.  I thought I actually removed that before pushing.   But we should probably come up with a game plan for the naming before going forward with the PR...

jerryk55:
@roliveri @hsong-rh bucket name changed as discussed.  One other fix added.  Ready for re-review.  Need providers review as well from @bronaghs or designated reviewer.

bronaghs:
LGTM

PR #268, opened 2017-06-30
Enable to save key_pair in Authentication

hsong-rh:
Enable to save key_pair information in authentication table. Smart state analysis will use it to transfer files to the agent.


hsong-rh:

app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
```diff
@@ -1,2 +1,37 @@
 class ManageIQ::Providers::Amazon::CloudManager::AuthKeyPair < ManageIQ::Providers::CloudManager::AuthKeyPair
+  def self.raw_create_key_pair(ext_management_system, create_options)
+    ec2 = ext_management_system.connect
+    kp = ec2.create_key_pair(create_options)
+    ost = OpenStruct.new
```
roliveri: Can we use a hash or Struct here?

hsong-rh: Changed it to use Struct.

bronaghs:
Hi @hsong-rh - can you set the labels, thanks.

PR #270, opened 2017-07-04
run provider generator to sync changes from core

durandom:
ran the generator again with

```bash
bundle exec rails g provider amazon --path ../
```

these are the changes.

cc @bdunne

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -2,3 +2,4 @@
 --require spec_helper
 --color
 --order random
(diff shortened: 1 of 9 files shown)


bdunne:
The test failures are unrelated... Anyone fixing those?

durandom:
spec have been fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/271

PR #272, opened 2017-07-14
Adjust test result to match new data_type

syncrou:
The security_groups key now defaults to a data_type of `:array_integer`.  The resulting value is an Array of Integers as opposed to an array with the first value being an integer and the second being a human readable value.  

This fix modifies the resulting hash to expect an array of Integers.

Dependent on: https://github.com/ManageIQ/manageiq/pull/15572


syncrou:
@miq-bot add_label bug

@miq-bot assign @durandom 

cc - @gmcculloug, @imtayadeway

durandom:
👏 thanks @syncrou and thanks @gmcculloug for merging 😸

PR #275, opened 2017-07-19
[WIP] Add agent manager module for Amazon SSA support

hsong-rh:
Introduce a module to add SSA agent management functionalities. The deploy_agent will create an instance in AWS as SSA agent with all necessary packages installed. This agent is running to wait for SSA requests.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
@@ -1,0 +1,326 @@
+require 'aws-sdk'
+require 'linux_admin'
+require 'amazon_ssa_support'
+
+module ManageIQ::Providers::Amazon::CloudManager::SsaAgentManager
+
+  MIQ_SSA = "MIQ_SSA"
+
+  #
+  # Intialize Amazon services
+  #
+
+  def ec2
+    @ec2 ||= ext_management_system.connect(:service => 'EC2')
+  end
+
+  def sqs
+    @sqs ||= ext_management_system.connect(:service => 'SQS')
+  end
+
+  def s3
+    @s3 ||= ext_management_system.connect(:service => 'S3')
+  end
+
+  def iam
+    @iam ||= ext_management_system.connect(:service => 'IAM')
+  end
+
+  def config
+    @config ||= VMDB::Config.new("vmdb").config
+  end
+
+  def region
+    ext_management_system.provider_region
+  end
+
+  def log_level
+    ll = config[:log][:level_aws] || AmazonSsaSupport::DEFAULT_LOG_LEVEL
+    ll.upcase
+  end
+
+  def heartbeat_prefix
+    config[:aws_ssa_heartbeat_prefix] || AmazonSsaSupport::DEFAULT_HEARTBEAT_PREFIX
+  end
+
+  def heartbeat_interval
+    config[:aws_ssa_heartbeat_interval] || AmazonSsaSupport::DEFAULT_HEARTBEAT_INTERVAL
+  end
+
+  def ssa_bucket
+    config[:aws_ssa_bucket] || AmazonSsaSupport::DEFAULT_BUCKET_PREFIX + '-' + ext_management_system.guid
+  end
+
+  def request_queue
+    config[:aws_ssa_request_queue] || AmazonSsaSupport::DEFAULT_REQUEST_QUEUE
+  end
+
+  def reply_queue
+    config[:aws_ssa_reply_queue] || AmazonSsaSupport::DEFAULT_REPLY_QUEUE
+  end
+
+  def reply_prefix
+    config[:aws_ssa_reply_prefix] || AmazonSsaSupport::DEFAULT_REPLY_PREFIX
+  end
+
+  def log_prefix
+    config[:aws_ssa_log_prefix] || AmazonSsaSupport::DEFAULT_LOG_PREFIX
+  end
+
+  def profile_name
+    "MIQ_SSA"
+  end
+
+  # check timestamp of heartbeat of agent_id, return true if the last beat time in
+  # in the time interval
+  def agent_alive?(agent_id, interval = 300)
+    bucket = s3.bucket(ssa_bucket)
+    return false unless bucket.exists?
+
+    obj_id = heartbeat_prefix + agent_id
+    obj = bucket.object(obj_id)
+    return false unless obj.exists?
+   
+    last_beat_stamp = YAML.load(obj.get.body.read, safe: true)
+    Time.now.utc - last_beat_stamp > interval ? false : true
+  end
+
+  def deploy_agent
+    kp = get_key_pair
+    security_group_id = create_security_group
+    data = create_user_data
+    zone_name = ec2.client.describe_availability_zones.availability_zones[0].zone_name
+    subnets = get_subnets(zone_name)
+    raise "No subnet_id is available for #{zone_name}!" if subnets.length == 0
+    create_profile
+
+    instance = ec2.create_instances({
+      image_id: get_agent_image_id,
+      min_count: 1,
+      max_count: 1,
+      key_name: kp.name,
+      security_group_ids: [security_group_id],
+      user_data: data,
+      instance_type: 't2.micro',
+      placement: {
+        availability_zone: zone_name
+      },
+      subnet_id: subnets[0].subnet_id,
+      iam_instance_profile: {
+        name: profile_name
+      },
+      tag_specifications: [{
+        resource_type: "instance",
+        tags: [{
+          key: "Name",
+          value: MIQ_SSA
+        }]
+      }]
+    })
+    ec2.client.wait_until(:instance_status_ok, {instance_ids: [instance[0].id]})
+
+    instance[0].id
+  end
+
+  private
+  def role_exists?(role_name)
+    begin
+      role = iam.role(role_name)
+      role.role_id
+      true
+    rescue ::Aws::IAM::Errors::NoSuchEntity
+      false
+    end
+  end
+
+  def create_role(role_name = MIQ_SSA)
+    return if role_exists?(role_name)
+
+    # Policy Generator:
+    policy_doc = {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Effect": "Allow",
+          "Principal": {"Service": "ec2.amazonaws.com"},
+          "Action": "sts:AssumeRole"
+        }
+      ]
+    }
+
+    role = iam.create_role({
+      role_name: role_name,
+      assume_role_policy_document: policy_doc.to_json
+    })
+    
+    # grant all priviledges
+    role.attach_policy({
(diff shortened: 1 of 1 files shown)


jerryk55:
So this one depends on https://github.com/ManageIQ/amazon_ssa_support/pull/21, right?

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+      security_group.authorize_ingress({
+        ip_permissions: [{
+          ip_protocol: 'tcp',
+          from_port: 443,
+          to_port: 443,
+          ip_ranges: [{
+            cidr_ip: '0.0.0.0/0'
+          }]}]
+      })
+
```
jerryk55: Instead of repeating these three calls to authorize_ingress explicitly can you set up a loop with a list of the 3 sets of port numbers and make the call inside the loop?  Less code.  Also define the port numbers outside the body of the method -
e.g. PORTS = [22, 80, 443]  ...

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+echo \'ssa_root=`bundle show amazon_ssa_support`\' >> start_agent.sh
+echo \'ssa_script="$ssa_root/tools/amazon_ssa_extract.rb"\' >> start_agent.sh
+echo \'ruby ${ssa_script}\' >> start_agent.sh
+chmod 755 start_agent.sh
+echo "Agent starts ...." >> /var/log/miq_ssa_deploy.log 2>&1
+./start_agent.sh
+'
+    Base64.encode64(script)
+  end
+
```
jerryk55: This seems like a lot of shell script code to have defined inside a ruby source file.
Two ideas - and the second one is better - 
1) The script should be defined via a "Here document" instead of just a really long string.
2) The script should be an external file defined in the source tree.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+    imgs = ec2.client.describe_images(
+      filters: [
+        {
+          name: "name",
+          values: [ "RHEL-7.3_HVM_GA*" ]
+        }
+      ]
+    ).images
+
+    imgs[0].image_id
```
jerryk55: Is the above value always correct?  Do we know that the first image with that prefix will always be the one we want?

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+  # Get Key Pair for SSH. Create a new one if not exists. 
+  def get_key_pair(pair_name = MIQ_SSA)
+    kp = Authentication.where(name: pair_name)
+    return kp[0] if kp.length > 0
+
+    ManageIQ::Providers::CloudManager::AuthKeyPair.create_key_pair(ext_management_system.id,
+      { :key_name => pair_name })
+  end
+
+  def create_pem_file(pair_name = MIQ_SSA)
```
jerryk55: Where is this called?

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+    kp = get_key_pair(pair_name)
+    pem_file_name = pair_name+".pem"
+    File.open(pem_file_name, 'w') {|f| f.write(kp.auth_key) }
+    File.chmod(0400, pem_file_name)
+    pem_file_name
+  end
+
+  def agent_startup_script(script_name = '/tools/amazon_ssa_extract.rb')
```
jerryk55: Where is this called too?

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+    kp = Authentication.where(name: pair_name)
+    return kp[0] if kp.length > 0
+
+    ManageIQ::Providers::CloudManager::AuthKeyPair.create_key_pair(ext_management_system.id,
+      { :key_name => pair_name })
+  end
+
+  def create_pem_file(pair_name = MIQ_SSA)
+    kp = get_key_pair(pair_name)
+    pem_file_name = pair_name+".pem"
```
jerryk55: Should be a space before and after the "+" (I know - RuboCop comment)  :-)

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+  def create_user_data
+    script = '#!/bin/bash
+yum -y update > /var/log/miq_ssa_deploy.log 2>&1
+yum -y install git-core zlib zlib-devel gcc-c++ patch readline readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison curl sqlite-devel postgresql-devel >> /var/log/miq_ssa_deploy.log 2>&1
+git clone https://github.com/rbenv/rbenv.git ~/.rbenv >> /var/log/miq_ssa_deploy.log 2>&1
+export HOME="/root" >> /var/log/miq_ssa_deploy.log 2>&1
+echo \'export PATH="$HOME/.rbenv/bin:$PATH"\' >> ~/.bash_profile
+echo \'eval "$(/root/.rbenv/bin/rbenv init -)"\' >> ~/.bash_profile
+source ~/.bash_profile >> /var/log/miq_ssa_deploy.log 2>&1
+git clone git://github.com/sstephenson/ruby-build.git ~/.rbenv/plugins/ruby-build >> /var/log/miq_ssa_deploy.log 2>&1
```
jerryk55: Is this path right?  Why are we using "sstephenson's" ruby-build instead of the one in rbenv like 5 lines above?

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+  def heartbeat_prefix
+    config[:aws_ssa_heartbeat_prefix] || AmazonSsaSupport::DEFAULT_HEARTBEAT_PREFIX
+  end
+
+  def heartbeat_interval
+    config[:aws_ssa_heartbeat_interval] || AmazonSsaSupport::DEFAULT_HEARTBEAT_INTERVAL
+  end
+
+  def ssa_bucket
+    config[:aws_ssa_bucket] || AmazonSsaSupport::DEFAULT_BUCKET_PREFIX + '-' + ext_management_system.guid
```
hsong-rh: Yes, Config should only has the prefix. Changed to append GUID for both cases.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -0,0 +1,326 @@
...
+  def reply_prefix
+    config[:aws_ssa_reply_prefix] || AmazonSsaSupport::DEFAULT_REPLY_PREFIX
+  end
+
+  def log_prefix
+    config[:aws_ssa_log_prefix] || AmazonSsaSupport::DEFAULT_LOG_PREFIX
+  end
+
+  def profile_name
+    "MIQ_SSA"
```
hsong-rh: Replaced with the global MIQ_SSA

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
hsong-rh: In AWS there is only one RH image with such a keyword in its name. Even there are more than one, the GA version of image should meet our requirement and we'll use the first one.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
hsong-rh: I'll define a separate authorize_ingress method with parameters of protocol, port numbers. There is no duplicate codes any more, and can support more possible settings.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
hsong-rh: The method can be called whenever we need to download the private key file for SSH/SCP purposes.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
hsong-rh: Added them.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
hsong-rh: This method is useless, removed.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -211,38 +211,24 @@ def create_security_group(group_name = MIQ_SSA)
...
-          ip_ranges: [{
-            cidr_ip: '0.0.0.0/0'
-          }]}]
-      })
+      security_group.group_id
+    end
+  end
 
-      security_group.authorize_ingress({
+  def sg_authorize_ingress(sg, protocal, from_port, to_port)
```
jerryk55: You misspelled the second argument as "protocal" instead of "protocol".  you spelled it correctly on line 225 so this needs to be changed.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
```diff
@@ -256,63 +242,87 @@ def get_key_pair(pair_name = MIQ_SSA)
...
     File.chmod(0400, pem_file_name)
     pem_file_name
   end
 
-  def agent_startup_script(script_name = '/tools/amazon_ssa_extract.rb')
-    ssa_root = Gem.loaded_specs['amazon_ssa_support'].full_gem_path
-    ssa_script = ssa_root + script_name
-    ssa_cmd = "ruby #{ssa_script}"
+  def setup_ruby(log = '/var/log/miq_ssa_deploy.log')
+    <<~HEREDOC
```
jerryk55: There's nothing actually wrong with naming your Here Doc "HEREDOC" but its kind of like naming a variable "VARIABLE".  You should probably name this and the ones in the other methods something more descriptive to the purpose of the script.

app/models/manageiq/providers/amazon/cloud_manager/ssa_agent_manager.rb
hsong-rh: Good catch! Will change it.

hsong-rh:
Moved these changes to manageiq repo. Closed this PR.

PR #280, opened 2017-08-03
Overriding queue name to generate an array of queue names

juliancheal:
this allows us to create an array of child managers for our queue.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresh_worker.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresh_worker.rb
@@ -1,3 +1,16 @@
 class ManageIQ::Providers::Amazon::CloudManager::RefreshWorker < ManageIQ::Providers::BaseManager::RefreshWorker
   require_nested :Runner
+
+  # overriding queue_name_for_ems so PerEmsWorkerMixin picks up *all* of the
+  # Amazon-manager types from here.
+  # This way, the refresher for Amazon's CloudManager will refresh *all*
+  # of the Amazon inventory across all managers.
+  def self.queue_name_for_ems(ems)
+    return ems unless ems.kind_of?(ExtManagementSystem)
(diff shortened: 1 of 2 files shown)


juliancheal:
This PR relies on https://github.com/ManageIQ/manageiq/pull/15711

juliancheal:
@miq-bot remove_label wip

app/models/manageiq/providers/amazon/cloud_manager/refresh_worker.rb
```diff
@@ -1,3 +1,16 @@
 class ManageIQ::Providers::Amazon::CloudManager::RefreshWorker < ManageIQ::Providers::BaseManager::RefreshWorker
   require_nested :Runner
+
+  # overriding queue_name_for_ems so PerEmsWorkerMixin picks up *all* of the
+  # Amazon-manager types from here.
+  # This way, the refresher for Amazon's CloudManager will refresh *all*
+  # of the Amazon inventory across all managers.
+  def self.queue_name_for_ems(ems)
+    return ems unless ems.kind_of?(ExtManagementSystem)
+    ["ems_#{ems.id}"] + ems.child_managers.collect { |manager| "ems_#{manager.id}" }
```
blomquisg: The parent implementation handles strange edge cases like if the input is a `Host`.

Maybe this should be:

```ruby
if ems.kind_of?(ExtManagementSystem)
  ["ems_#{ems.id}"] + ems.child_managers.collect { |manager| "ems_#{manager.id}" }
else
  super
end
```

Then, let `super` handle any of the edge cases.

juliancheal: Yeah agreed. I think I was too happy to have found my last bug!

juliancheal:
@durandom @blomquisg merge?

durandom:
Dont we need the dependent PR merged first?

juliancheal:
> Dont we need the dependent PR merged first?

To be honest, I don't know which PR needs merging first :(

PR #283, opened 2017-08-11
Lower the batch sizes to also test multiple batches being processed

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/15794
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/282
 
Lower the batch sizes to also test multiple batches being processed

### DIFF
--- a/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
+++ b/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
@@ -39,6 +39,11 @@ module AwsRefresherSpecCommon
   ].freeze
 
   def stub_refresh_settings(settings)
+    # TODO(lsmola) extract the batch sizes to the settings and stub the settings instead
+    # Lower batch sizes to test multiple batches for each collection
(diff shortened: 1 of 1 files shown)


spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -39,6 +39,11 @@ module AwsRefresherSpecCommon
   ].freeze
 
   def stub_refresh_settings(settings)
+    # TODO(lsmola) extract the batch sizes to the settings and stub the settings instead
+    # Lower batch sizes to test multiple batches for each collection
+    allow_any_instance_of(ManagerRefresh::InventoryCollection).to receive(:batch_size).and_return(4)
+    allow_any_instance_of(ManagerRefresh::InventoryCollection).to receive(:batch_size_pure_sql).and_return(4)
```
agrare: I dislike `allow_any_instance_of` this is a good reason why the batch size should be configurable

PR #285, opened 2017-08-11
Delete provision_requests_spec.rb

imtayadeway:
Moved here: https://github.com/ManageIQ/manageiq-api/pull/18

Should get the build back to 💚 

/cc @Ladas


Ladas:
@imtayadeway awesome, thank you :-)

PR #286, opened 2017-08-15
Move OrchestrationTemplateCfn to manageiq-providers-amazon

bzwei:
Prepare to move OrchestrationTemplate to each provider.

This is a step before the migration.

https://bugzilla.redhat.com/show_bug.cgi?id=1417021


Ladas:
@bzwei seems like a valid CI failure. Does this depend on some core PR?

bzwei:
@miq-bot add_label refactoring
@miq-bot assign @Ladas

Ladas:
@bzwei can we fix the rubocop warning? Or you need to call the load there?

bzwei:
@Ladas warning fixed.

PR #289, opened 2017-08-25
Support custom instance_types

jameswnl:
Allow user to add instance_types that MIQ doesn't include. These are types that added by AWS after MIQ's code freeze for release.

### DIFF
--- a/lib/tasks_private/instance_types.rake
+++ b/lib/tasks_private/instance_types.rake
@@ -4,9 +4,9 @@ task :instance_types => [:environment] do
   #
   # This task creates a instance_types.rb file in the current directory based upon the current
   # instance_types.rb file and data from
-  #  curl https://raw.githubusercontent.com/powdahound/ec2instances.info/master/www/instances.json > lib/tasks/instance_types_data/instances.json
-  #  https://aws.amazon.com/ec2/instance-types/#instance-type-matrix > lib/tasks/instance_types_data/instance_types.csv
-  #  https://aws.amazon.com/ec2/previous-generation/#Previous_Generation_Instance_Details_and_Pricing_ > lib/tasks/instance_types_data/instance_types_previous.csv
+  #  curl https://raw.githubusercontent.com/powdahound/ec2instances.info/master/www/instances.json > lib/tasks_private/instance_types_data/instances.json
+  #  https://aws.amazon.com/ec2/instance-types/#instance-type-matrix > lib/tasks_private/instance_types_data/instance_types.csv
+  #  https://aws.amazon.com/ec2/previous-generation/#Previous_Generation_Instance_Details_and_Pricing_ > lib/tasks_private/instance_types_data/instance_types_previous.csv
   #     converted to csv via http://www.convertcsv.com/html-table-to-csv.htm
   #
   # Other useful resources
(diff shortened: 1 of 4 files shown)


jameswnl:
@miq-bot add_labels enhancement

lib/tasks_private/instance_types.rake
```diff
@@ -1,12 +1,12 @@
-desc "Create new instance_types.rb"
-task :instance_types => [:environment] do
+desc("Create new instance_types.rb")
+task(:instance_types => [:environment]) do
```
Fryguy: Don't add the parens in...these are DSL methods.  If anything we should change the codeclimate.yml if that's the reason for this.  cc @bdunne

bdunne: A section for the rake gem under https://github.com/ManageIQ/guides/blob/master/.rubocop_base.yml#L75 would be great.  @jameswnl do you want to add that or should I?

jameswnl: @Fryguy updated!
@bdunne yes, in this [PR](https://github.com/ManageIQ/guides/pull/249)

lib/tasks_private/instance_types.rake
```diff
@@ -190,14 +190,20 @@ EOT
 
   new_instance_types_file << output_instances.map { |i| format_instance(i) }.join("\n")
   new_instance_types_file << <<EOT
-  }
+  }.freeze
+
+  def self.instance_types
```
bdunne: Is this copied from above?  Can we use that method instead of duplicating it?

jameswnl: You mean the `def self.instance_types` method in `instance_types.rb`?  
The `instance_types.rb` is the output of this rake task

Fryguy: I find it weird in general that this generates source code.  I feel this task should just generate a YAML file, which is processed on first access by the code.  Then this "duplication" wouldn't exist either.

jameswnl: I feel strongly the same as well.

My plan is to finish this task first for 4.6 and then revisit this.

bronaghs:
Going to merge this for now, but we will improve this in the near future to generate a JSON or YAML output file instead of a .rb file.

bronaghs:
@jameswnl - Can this be backported to FINE?

jameswnl:
@bronaghs yes, I think so.

jameswnl:
@miq-bot remove_labels fine/no
@miq-bot add_labels fine/yes

@simaishi this should be backportable. let me know if you need help. thanks

simaishi:
@bronaghs please add BZ link.

bronaghs:
@simaishi - https://bugzilla.redhat.com/show_bug.cgi?id=1557025

simaishi:
Fine backport details:

```
$ git log -1
commit 27eb32c874b4c877f200b81b0e141b73a2703287
Date:   Mon Sep 18 14:35:06 2017 -0400

    Merge pull request #289 from jameswnl/instancetypes
    
    Support custom instance_types
    (cherry picked from commit 5d3a5f8950878e75d9cbef6255213b9c04e0aaa9)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1557025
```

PR #290, opened 2017-09-01
Turn on graph refresh with targeted refresh as default

Ladas:
Turn on graph refresh with targeted refresh as default

Partially implements:
https://bugzilla.redhat.com/show_bug.cgi?id=1395356

----
related PRs for actually making targeted refresh work from automate event handlers

https://github.com/ManageIQ/manageiq-content/pull/178
https://github.com/ManageIQ/manageiq-content/pull/179
https://github.com/ManageIQ/manageiq-automation_engine/pull/67

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -49,14 +49,14 @@
         :values:
           - machine
     :ignore_terminated_instances: true
-    :inventory_object_refresh: false
-    :allow_targeted_refresh: false
+    :inventory_object_refresh: true
+    :allow_targeted_refresh: true
   :ec2_network:
-    :inventory_object_refresh: false
+    :inventory_object_refresh: true
   :s3:
     :inventory_object_refresh: true
   :ec2_ebs_storage:
-    :inventory_object_refresh: false
+    :inventory_object_refresh: true
 :http_proxy:
   :ec2:
     :host:


blomquisg:
@agrare @durandom @bronaghs Thoughts?

agrare:
@blomquisg we'll need to turn it on eventually I guess now is as good a time as any

bronaghs:
lets do it.

durandom:
🎉

cben:
I believe you too lost tag mapping when you switched this on — currently `retag_entity` is only called from save_inventory_container & save_inventory_cloud.

I'm [working](https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/126) on rewriting tag mapping to make it work in graph refresh too, for at least containers — and hopefully amazon but will need reviews/guidance for that.
first step:
https://github.com/ManageIQ/manageiq-providers-amazon/pull/316 
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/130
https://github.com/ManageIQ/manageiq/pull/16098

Ladas:
@bronaghs @cben hm, right, I think the specs for the tag mapping were commented out and we forgot about that :-) OK, lets do AWS once you are done @cben

PR #292, opened 2017-09-01
Turn on batch saving by default

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/290

Turn on batch saving by default

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -51,12 +51,20 @@
     :ignore_terminated_instances: true
     :inventory_object_refresh: true
     :allow_targeted_refresh: true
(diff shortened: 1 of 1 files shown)


Ladas:
@agrare well it's quicker :-) but yeah, I'd like to have more testing

blomquisg:
@agrare I'll let you make the call.  You know the cost/benefit better than I do.

Ladas:
@agrare we can merge now safely :-) We could backport to g-release

agrare:
Yes now we can merge it @Ladas :)

PR #293, opened 2017-09-04
Prefix API based events so we can identify them easily

Ladas:
Prefix API based events so we can identify them easily


Ladas:
cc @bronaghs, @durandom  something we talked about while ago, that all events look better when prefixed

Ladas:
@durandom yeah, those events are not tied to anything, except the definitions for timelines, I would just ignore those

bronaghs:
@Ladas 
I think the Aws API events are configured in the event switchboard, at least for EBS. Below is part of some example full text from an EBS CreateVolume event saved in our DB and [here](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Event/EmsEvent/Amazon.class/createvolume.yaml) is the corresponding event switchboard config.
We initially pursued CloudWatch EC2 events for EBS but learned these are only fired upon failure, see [here](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Event/EmsEvent/Amazon.class/createvolume.yaml)


version: '0'
id: f56b871d-ed8f-69a2-5423-790d6edfeeb6
detail-type: AWS API Call via CloudTrail
source: aws.ec2
time: '2017-09-08T20:44:15Z'
region: us-east-1
resources: []
detail:
  eventTime: '2017-09-08T20:44:15Z'
  eventSource: ec2.amazonaws.com
  eventName: CreateVolume
  awsRegion: us-east-1
  requestParameters:
    size: '1'
    zone: us-east-1e
    volumeType: gp2
    encrypted: false
    tagSpecificationSet:
      items:
      - resourceType: volume
        tags:
        - key: name
          value: bronagh_sept
  responseElements:
    size: '1'
    zone: us-east-1e
    status: creating
    createTime: 1504903455915
    volumeType: gp2
    tagSet:
      items:
      - key: name
        value: bronagh_sept
  eventType: AwsApiCall
eventType: CreateVolume
event_source: :cloud_watch_api

Ladas:
@bronaghs yes I am changing and expanding the event handlers here https://github.com/ManageIQ/manageiq-content/pull/183/files

bronaghs:
So they need to be merged at the same time then.
@gmcculloug - Once you approve or merge https://github.com/ManageIQ/manageiq-content/pull/183 then I will merge this PR.

PR #301, opened 2017-09-21
Improve instance_types.rake logic

jameswnl:
I have submitted [this PR](https://github.com/powdahound/ec2instances.info/pull/270). That PR will enhance the site's json output so that we can eliminate the [tedious manual step to curl and using 3rd party site to convert html to csv](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/lib/tasks_private/instance_types.rake#L8) 

~~Meanwhile we can use [my forked copy](https://github.com/jameswnl/ec2instances.info/blob/cpu/scrape.py) and run `python scrape.py` locally to generate the json file.~~
### update
The [PR](https://github.com/powdahound/ec2instances.info/pull/270) to http://ec2instances.info has been merged


jameswnl:
@durandom Would like to get your input on this before un-wip it.

@miq-bot add_label wip, enhancement

app/models/manageiq/providers/amazon/instance_types.rb
```diff
@@ -14,11 +14,11 @@ module ManageIQ::Providers::Amazon::InstanceTypes
       :ebs_only                => true,
       :instance_store_size     => 0,
       :instance_store_volumes  => 0,
-      :architecture            => [:x86_64],
+      :architecture            => [:i386, :x86_64],
       :virtualization_type     => [:hvm],
       :network_performance     => :low,
       :physical_processor      => "Intel Xeon family",
-      :processor_clock_speed   => 3.3,
+      :processor_clock_speed   => "up to 3.3",
```
durandom: is this a problem, that this is now a string?
is the db a string?
will it impact UI sorting?

app/models/manageiq/providers/amazon/instance_types.rb
```diff
@@ -1328,50 +1280,146 @@ module ManageIQ::Providers::Amazon::InstanceTypes
...
       :description             => "F1 Double Extra Large",
       :memory                  => 122.0.gigabytes,
       :vcpu                    => 8,
       :ebs_only                => false,
-      :instance_store_size     => 480.0.gigabytes,
+      :instance_store_size     => 470.0.gigabytes,
       :instance_store_volumes  => 1, # SSD
       :architecture            => [:x86_64],
-      :virtualization_type     => [],
-      :network_performance     => :high,
```
durandom: is `:network_performance` a free text? IIRC it should be rather a set of [:high, :low, etc]

lib/tasks_private/instance_types.rake
```diff
@@ -52,58 +49,55 @@ task :instance_types => [:environment] do
...
-    o[:instance_store_ssd] = i.fetch_path(:storage, :ssd)
-    o[:architecture] = i[:arch].map(&:to_sym).sort
-    o[:virtualization_type] = i[:linux_virtualization_types].map(&:downcase).map(&:to_sym).sort.map { |v| v == :pv ? :paravirtual : v }
-    o[:network_performance] = case i[:network_performance]
-                              when '10 Gigabit', '20 Gigabit'
-                                :very_high
-                              else
-                                i[:network_performance].downcase.tr(' ', '_').to_sym
-                              end
-    # see https://github.com/ManageIQ/manageiq/issues/741#issuecomment-57353290
```
durandom: I'd leave this comment in the code as a pointer to what 'ebs_only' means

lib/tasks_private/instance_types.rake
```diff
@@ -5,9 +5,6 @@ task :instance_types => [:environment] do
   # This task creates a instance_types.rb file in the current directory based upon the current
   # instance_types.rb file and data from
   #  curl https://raw.githubusercontent.com/powdahound/ec2instances.info/master/www/instances.json > lib/tasks_private/instance_types_data/instances.json
-  #  https://aws.amazon.com/ec2/instance-types/#instance-type-matrix > lib/tasks_private/instance_types_data/instance_types.csv
```
durandom: I see that this is now included in ec2instances.info, which is great

lib/tasks_private/instance_types.rake
```diff
@@ -5,9 +5,6 @@ task :instance_types => [:environment] do
...
-  #  https://aws.amazon.com/ec2/previous-generation/#Previous_Generation_Instance_Details_and_Pricing_ > lib/tasks_private/instance_types_data/instance_types_previous.csv
```
durandom: but this isnt included...
did you remove it, because we assume that previous generations do not change anymore?

app/models/manageiq/providers/amazon/instance_types.rb
jameswnl: The original code was 

`o[:processor_clock_speed] = row[:clock_speed_ghz].scan(/[\d\.]/).join`

And that won't work because some of the new values of `row[:clock_speed_ghz]` are like `'2.3 to 2.7'`. The `scan` will return '2.32.7' which doesn't make sense and also breaks the numeric field.

This is the part need your help. Do you know how we use this? Or who would know that I can ask?

app/models/manageiq/providers/amazon/instance_types.rb
jameswnl: Similarly, there are new values breaking the original parsing logic which was
```
 -    o[:network_performance] = case i[:network_performance]
 -                              when '10 Gigabit', '20 Gigabit'
 -                                :very_high
 -                              else
 -                                i[:network_performance].downcase.tr(' ', '_').to_sym
 -                              end
```
Yes, AWS use free text for this field. So it's hard to parse it into structured type.

lib/tasks_private/instance_types.rake
jameswnl: Yes, they just merged my https://github.com/powdahound/ec2instances.info/pull/270 hours ago 😉

lib/tasks_private/instance_types.rake
jameswnl: The original output from `ec2instance.info` already has a field `generation` which has value either 'current' or 'previous'. So that's how I determine it.

lib/tasks_private/instance_types.rake
jameswnl: sure, done!

app/models/manageiq/providers/amazon/instance_types.rb
jameswnl: Actually this won't matter, the db doesn't care about this, it only cares about 
```
new_result = {
      :type                     => ManageIQ::Providers::Amazon::CloudManager::Flavor.name,
      :ems_ref                  => uid,
      :name                     => name,
      :description              => flavor[:description],
      :enabled                  => !flavor[:disabled],
      :cpus                     => cpus,
      :cpu_cores                => 1,
      :memory                   => flavor[:memory],
      :supports_32_bit          => flavor[:architecture].include?(:i386),
      :supports_64_bit          => flavor[:architecture].include?(:x86_64),
      :supports_hvm             => flavor[:virtualization_type].include?(:hvm),
      :supports_paravirtual     => flavor[:virtualization_type].include?(:paravirtual),
      :block_storage_based_only => flavor[:ebs_only],
      :cloud_subnet_required    => flavor[:vpc_only],
      :ephemeral_disk_size      => flavor[:instance_store_size],
      :ephemeral_disk_count     => flavor[:instance_store_volumes]
    }
```
as in [parsing code here](https://github.com/jameswnl/manageiq-providers-amazon/blob/32bc2aefe1f39b86030000031f615dde650a97ae/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb#L113-L113)

jameswnl:
@miq-bot remove_label wip

app/models/manageiq/providers/amazon/instance_types.rb
bronaghs: I dont think we actually use this.

PR #309, opened 2017-10-09
Add parameter in setting for SSA support

hsong-rh:
Add some parameters in settings. Both AmazonAgentMangerWorker and the agent (EC2 instance deployed in AWS cloud) will use this parameters to share same request/reply SQS queue and S3 bucket.


hsong-rh:
@jerryk55 @roliveri @blomquisg Please review.

hsong-rh:
Move these changes to https://github.com/ManageIQ/manageiq-providers-amazon/pull/318. Closed this.

PR #313, opened 2017-10-11
Test all aws config events

Ladas:
Depends on:

- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/310
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/312

Test all aws config events


Ladas:
@bronaghs ping

PR #314, opened 2017-10-12
Cleanup refreshers using base class refresher for ManagerRefresh

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/16182
- [x] https://github.com/ManageIQ/manageiq/pull/16180

Cleanup refreshers using base class refresher for ManagerRefresh


Ladas:
@miha-plesko @mansam you will be able to do similar cleanup once the core PRs are merged :-)

PR #318, opened 2017-10-17
Add agent coordinator for smartstate support

hsong-rh:
This PR will combine two PRs' changes together: 
https://github.com/ManageIQ/manageiq/pull/16058
https://github.com/ManageIQ/manageiq-providers-amazon/pull/309.


hsong-rh:
@Fryguy @roliveri @jerryk55 Please review this.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,419 @@
...
+  end
+
+  def startup_agent
+    agent_ids.empty? ? deploy_agent : activate_agents
+  rescue => err
+    _log.error("No agent is set up to prcoess requests: #{err.message}")
+    _log.error(err.backtrace.join("\n"))
+  end
+
+  def scp_file(local_file, remote_file, instance, username = "centos")
```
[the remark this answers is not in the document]
hsong-rh: Yes, we do. This user 'centos' is predefined in the AMI image of 'CentOS Atomic Host'. After launching our agent, we can only use this user to SSH or SCP.  Maybe we should also keep this information in the setting, pair with the AMI image.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,419 @@
...
+  def activate_agents
+    agent_ids.each do |id|
+      agent = ec2.instance(id)
+      if agent.state.name == "stopped"
+        agent.start
+        agent.wait_until_running
+        _log.info("Agent #{id} is activated to serve requests.")
+        return id
+      else
+        _log.warn("Agent #{id} is in abnormal state: #{agent.state.name}.")
```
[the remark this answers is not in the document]
hsong-rh: If some agents are running, we just use it and will not activate any other agents.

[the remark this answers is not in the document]
roliveri: Why? If we configure 4 agents and only 3 are running we'd want to start one.

app/models/manageiq/providers/amazon/agent_coordinator.rb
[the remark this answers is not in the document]
roliveri: Yes, this can't be hardcoded because the value is image specific. If the image can change, so can this.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+
+  def setup_agent(instance)
+    # register if needed
+    register(instance)
+
+    # prepare work directory
+    ssh_commands(instance, ["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"], agent_ami_login_user)
+    # scp the default setting yaml file
+    create_config_yaml("config.yml")
+    scp_file("config.yml", WORK_DIR.to_s, instance, agent_ami_login_user)
```
Fryguy: You created a file, but didn't cleanup the file.  This needs to either be cleaned up or created in temp space.

Fryguy: Alternately, just create the file over SSH.

app/models/manageiq/providers/amazon/agent_coordinator.rb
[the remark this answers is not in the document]
Fryguy: Agreed this value cannot be hardcoded, and should be removed from here.

app/models/manageiq/providers/amazon/agent_coordinator.rb
Fryguy: Also, this .to_s should be moved into the scp_file method itself...the caller should not have to be concerned with that.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+
+    # run docker image
+    command_line = "sudo docker run -d --restart=always -v /dev:/host_dev -v #{WORK_DIR}/config.yml:#{WORK_DIR}/config.yml --privileged #{agent_docker_name}"
```
Fryguy: @bdunne Can you verify this?

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+    ssh_commands(instance, [command_line.to_s], agent_ami_login_user)
```
Fryguy: Why the .to_s?  It's already a string.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+
+    instance[0].id
+  end
+
+  def setup_agent(instance)
+    # register if needed
+    register(instance)
+
+    # prepare work directory
+    ssh_commands(instance, ["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"], agent_ami_login_user)
```
Fryguy: Instead of a semicolon, these can be two separate strings.

app/models/manageiq/providers/amazon/agent_coordinator.rb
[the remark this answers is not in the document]
Fryguy: This should be a private method

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,419 @@
...
+    ip = instance.public_dns_name
+    key_name = instance.key_name
+    auth_key = get_keypair(key_name).try(:auth_key)
+    _log.error("AuthKey [#{key_name}] is empty of #{instance.id}") if auth_key.nil?
+    Net::SCP.upload!(ip, username, local_file, remote_file, :ssh => {:key_data => auth_key})
+  rescue => err
+    _log.error(err.message)
+  end
+
+  def ssh_commands(instance, commands = [], username = "centos")
```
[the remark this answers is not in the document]
Fryguy: This should also be a private method.

app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
```diff
@@ -0,0 +1,8 @@
+class ManageIQ::Providers::Amazon::AgentCoordinatorWorker < MiqWorker
+  require_nested :Runner
+
+  self.required_roles = ['smartproxy']
```
Fryguy: Is this going to run for all smartproxy roles?  That is, if we have a VMware provider will it start one of these?  I think there needs to be more checks here.  cc @agrare

Fryguy: Also, isn't this supposed to be one per ems type?  I was expecting the per_ems_type_mixin.

agrare: If this was per ems type it wouldn't start I think

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+    # scp the default setting yaml file
+    create_config_yaml("config.yml")
+    scp_file("config.yml", WORK_DIR.to_s, instance, agent_ami_login_user)
+
+    # run docker image
+    command_line = "sudo docker run -d --restart=always -v /dev:/host_dev -v #{WORK_DIR}/config.yml:#{WORK_DIR}/config.yml --privileged #{agent_docker_name}"
+    ssh_commands(instance, [command_line.to_s], agent_ami_login_user)
+  end
+
+  # TODO: for downstream
```
Fryguy: Remove this.  If anything, just remove the method entirely and change the caller to `register(instance) if respond_to?(:register)`

Fryguy: Actually, a better way to handle this is to actually implement it in upstream by registering when the credentials are present.  That way, on upstream, no credentials will be present, so it will be a no-op and downstream the credentials will be required.  In fact you could make the "are credentials required" a configurable value...something like `:agent_ami_registration_required: true`

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+      :policy_arn => 'arn:aws:iam::aws:policy/AmazonS3FullAccess'
+    )
+
+    role.attach_policy(
+      :policy_arn => 'arn:aws:iam::aws:policy/AmazonEC2FullAccess'
+    )
+
+    role.attach_policy(
+      :policy_arn => 'arn:aws:iam::aws:policy/AmazonSQSFullAccess'
+    )
```
Fryguy: Simplifies to

```ruby
%w(AmazonS3FullAccess AmazonEC2FullAccess AmazonSQSFullAccess).each do |policy|
  role.attach_policy(:policy_arn => "arn:aws:iam::aws:policy/#{policy}")
end
```

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+
+    role
+  end
+
+  def role_exists?(role_name)
+    role = iam.role(role_name)
+    role.role_id
```
Fryguy: Does this line do anything?

Fryguy: Or does this lazily execute the check?  If so, perhaps this is cleaner?

```ruby
def role_exists?(role_name)
  !!iam.role(role_name).role_id
rescue ::Aws::IAM::Errors::NoSuchEntity
  false
end

File: app/models/manageiq/providers/amazon/agent_coordinator.rb (hunk 4 above)
Comment (bdunne) [reply app/models/manageiq/providers/amazon/agent_coordinator.rb]: 👍

File: app/models/manageiq/providers/amazon/agent_coordinator.rb (hunk 9 above)
Comment (bdunne) [reply app/models/manageiq/providers/amazon/agent_coordinator.rb]: To build on this... For any private docker registries, you need to `docker login`.  So I would call the method something like
```ruby
def login_to_registry
  return unless settings.docker_registry.url
  command_line = "docker login"
  command_line << " #{settings.docker_registry.url}"
  command_line << " -u #{settings.docker_registry.username} -p #{settings.docker_registry.password}" if settings.docker_registry.username
  send_command_over_ssh(command_line)
end

File: app/models/manageiq/providers/amazon/agent_coordinator.rb [hunk 12]
```diff
@@ -0,0 +1,423 @@
...
+  # Get Key Pair for SSH. Create a new one if not exists.
+  def find_or_create_keypair(keypair_name = label)
+    get_keypair(keypair_name) || begin
+      _log.info("KeyPair #{keypair_name} will be created!")
+      ManageIQ::Providers::CloudManager::AuthKeyPair.create_key_pair(@ems.id, :key_name => keypair_name)
+    end
+  end
+
+  def get_keypair(keypair_name = label)
+    kps = @ems.authentications.where(:name => keypair_name)
```
bdunne: Let the database do everything...
```ruby
@ems.authentications.find_by(:name => keypair_name)
```

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+        :name   => "name",
+        :values => [image_name]
+      }]
+    ).images
+
+    imgs[0].image_id
+  end
+
+  def create_pem_file(pair_name = keypair_name)
+    kp = find_or_create_keypair(pair_name)
```
bdunne: Why abbreviate `key_pair`?

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+  def heartbeat_prefix
+    AmazonSsaSupport::DEFAULT_HEARTBEAT_PREFIX
+  end
+
+  def heartbeat_interval
+    agent_coordinator_settings.try(:heartbeat_interval) || AmazonSsaSupport::DEFAULT_HEARTBEAT_INTERVAL
+  end
+
+  def ssa_bucket
+    @ssa_bucket ||= "#{AmazonSsaSupport::DEFAULT_BUCKET_PREFIX}-#{@ems.guid}"
```
bdunne: Should these strings be frozen to prevent modification?

app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
Fryguy: @hsong-rh Can you demo this to me?  I can't see how any of this can work from a full ManageIQ installation.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+  def reply_prefix
+    AmazonSsaSupport::DEFAULT_REPLY_PREFIX
+  end
+
+  def log_prefix
+    AmazonSsaSupport::DEFAULT_LOG_PREFIX
+  end
+
+  def agent_ami_name
+    agent_coordinator_settings.try(:agent_ami_name) || raise { "Please specify AMI image name for SSA agent" }
```
bdunne: Why are you passing a block to `raise`?

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    # Amazon providers may be added/removed. Keep monitor and update if needed.
+    latest_ems_guids = self.class.amazon_ems_guids
+    coordinator_guids = @coordinators.collect { |m| m.ems.guid }
+    self.class.refresh_coordinators(@coordinators, coordinator_guids, latest_ems_guids)
+  end
+
+  def self.agent_coordinator_by_guid(guid)
+    all_agent_coordinators_in_zone.find { |e| e.ems.guid == guid }
```
bdunne: Why is all of this being done in ruby rather than the database?

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
```diff
@@ -0,0 +1,63 @@
...
+  def self.ems_by_guid(guid)
+    all_valid_ems_in_zone.detect { |e| e.guid == guid }
+  end
+
+  def self.amazon_ems_guids
+    all_amazon_ems_in_zone.collect(&:guid)
+  end
+
+  def self.all_ems_in_zone
+    ExtManagementSystem.where(:zone_id => MiqServer.my_server.zone.id).to_a
```
bdunne: MiqServer.my_server.zone.ext_management_systems

app/models/manageiq/providers/amazon/agent_coordinator.rb
bdunne: From what I can tell, this string will never go anywhere...

```ruby
irb(main):005:0> def aaaaa
irb(main):006:1> raise { "abc" }
irb(main):007:1> end
=> :aaaaa
irb(main):008:0> begin
irb(main):009:1*   aaaaa
irb(main):010:1> rescue => error
irb(main):011:1>   puts error.message
irb(main):012:1> end

=> nil
```

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+  end
+
+  def get_keypair(keypair_name = label)
+    kps = @ems.authentications.where(:name => keypair_name)
+    kps.any? ? kps[0] : nil
+  end
+
+  def find_or_create_profile(profile_name = label, role_name = label)
+    ssa_profile = iam.instance_profile(profile_name)
+    ssa_profile = iam.create_instance_profile(:instance_profile_name => profile_name) unless ssa_profile.exists?
```
bdunne: You can tag a `.presence` onto the line above then use `||=` here and drop the `unless ssa_profile.exists?`

bdunne: I'm assuming `iam.instance_profile(profile_name)` is returning a string, so the `.presence` may not be needed?

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+        :name => label
+      },
+      :tag_specifications   => [{
+        :resource_type => "instance",
+        :tags          => [{
+          :key   => "Name",
+          :value => label
+        }]
+      }]
+    )
```
bdunne: Tack on a `.first`, then `instance` will be an instance, not an Array.  Then you don't need to `instance[0]` everywhere.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -0,0 +1,423 @@
...
+    @deploying = true
+
+    kp = find_or_create_keypair
+    zone_name = ec2.client.describe_availability_zones.availability_zones[0].zone_name
+    subnets = get_subnets(zone_name)
+    raise "No subnet_id is available for #{zone_name}!" if subnets.empty?
+    security_group_id = find_or_create_security_group(subnets[0].vpc_id)
+    find_or_create_profile
+
+    instance                = ec2.create_instances(
```
bdunne: What is this aligned with?

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
Fryguy: This is also already part of the PerEmsTypeWorkerMixin, so this is all duplication.

app/models/manageiq/providers/amazon/agent_coordinator.rb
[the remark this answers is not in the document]
hsong-rh: Changed to use setting to keep both AMI image and its user information.

app/models/manageiq/providers/amazon/agent_coordinator.rb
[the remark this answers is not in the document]
hsong-rh: Will open a PR to change these two methods into private.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Will changes these in the new PR.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: It's auto changed by rubocops. Will remove .to_s.

app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
hsong-rh: Will change this in new PR, so the coordinator worker only starts when: 
1. smart proxy role enabled;
2. Amazon provider exists in appliance;

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: @bdunne For docker login, what's the username and password for our manageiq?

Fryguy: There is no username/password for public docker images.

PR #319, opened 2017-10-18
Amazon SSA Response Handling Thread

jerryk55:
Add the Amazon SSA response thread which will be integrated with the
AgentCoordinatorWorker.  This thread checks the SQS queues set up
by the Amazon SSA Agents to see if any SSA responses have been added.
If so the responses are pulled out of the appropriate S3 buckets and
processed.  The appropriate Job is then updated with the status.

This PR is dependent upon https://github.com/ManageIQ/manageiq-providers-amazon/pull/318
and should neither be merged nor have any tests run on it until such time as the dependent PR is merged.

@Fryguy @hsong-rh @roliveri @bronaghs please review now so we are ready to merge in a timely manner.  Thank you.

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
@@ -62,10 +62,8 @@ module ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner::ResponseThre
           end
         end
       end
-      # While this is WIP we will hard code this to 10 seconds.
-      # reponse_check_sleep_seconds = response_thread_sleep_seconds
-      response_check_sleep_seconds = 10
-      _log.debug("going to sleep for #{response_check_sleep_seconds} seconds after getting all Agent Coordinators")
+      response_check_sleep_seconds = response_thread_sleep_seconds
+      _log.debug("going to sleep for #{response_check_sleep_seconds} seconds after checking all Agent Coordinators")
       sleep(response_check_sleep_seconds) unless @shutdown_instance_wait_thread
     end
   end
--- a/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
@@ -1,0 +1,179 @@
+module ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner::ResponseThread
+  require 'xml/xml_utils'
+  require 'scanning_mixin'
+  include ScanningMixin
+
+  #
+  # This thread startup method is called by the AgentCoordinatorWorker::Runner
+  # at the end of the do_before_work_loop method.
+  #
+  def start_response_thread
+    Thread.new do
+      _log.info("Starting AWS SSA Response Thread")
+      begin
+        # Wait a bit for the Agent Coordinators to be started
+        sleep(20)
+        wait_for_responses
+      rescue SystemExit
+        _log.info("SystemExit received, exiting AWS SSA Response Thread")
+        stop_response_thread
+        Thread.exit
+      rescue => err
+        _log.error("start_response_thread: Error [#{err}]")
+        _log.error("start_response_thread: Error - [#{err.backtrace.join("\n")}]")
+      end
+    end
+  end
+
+  def stop_response_thread
+    @shutdown_instance_wait_thread = true
+  end
+
+  def wait_for_responses
+    @shutdown_instance_wait_thread = nil
+    until @shutdown_instance_wait_thread
+      @coordinators.each do |coord|
+        _log.debug("Checking replies for Agent Coordinator Provider #{coord.ems.name}")
+        return if @shutdown_instance_wait_thread
+        unless coord.reply_queue_empty?
+          ssaq_args = {}
+          ssaq_args[:ssa_bucket]    = coord.ssa_bucket
+          ssaq_args[:reply_queue]   = coord.reply_queue
+          ssaq_args[:request_queue] = coord.request_queue
+          ssaq_args[:region]        = coord.ems.provider_region
+          ssaq_args[:sqs]           = coord.sqs
+          ssaq_args[:s3]            = coord.s3
+          ssaq                      = AmazonSsaSupport::SsaQueue.new(ssaq_args)
+          unless ssaq
+            _log.error "Error creating SsaQueue for #{coord.ems.name}"
+            next
+          end
+          _log.debug("Getting replies for #{coord.ems.name}")
+          begin
+            ssaq.reply_loop do |reply|
+              _log.debug("Reply for #{coord.ems.name}: #{reply}")
+              perform_metadata_sync(reply)
+            end
+          rescue => err
+            _log.error("Error #{err} processing replies for #{coord.ems.name}.  Continuing.")
+            next
+          end
+        else
+          _log.debug("No Replies visible for Provider #{coord.ems.name}")
+        end
+      end
+      # While this is WIP we will hard code this to 10 seconds.
+      # reponse_check_sleep_seconds = response_thread_sleep_seconds
+      response_check_sleep_seconds = 10
+      _log.debug("going to sleep for #{response_check_sleep_seconds} seconds after getting all Agent Coordinators")
+      sleep(response_check_sleep_seconds) unless @shutdown_instance_wait_thread
+    end
+  end
+
+  def perform_metadata_sync(extract_reply)
+    ost       = OpenStruct.new
+    ost.reply = extract_reply
+    ost.jobid = extract_reply[:job_id]
+    job       = Job.find_by(:id => ost.jobid)
+    raise _("Unable to sync data for job with id <%{number}>. Job not found.") % {:number => ost.jobid} if job.nil?
+    ost.taskid = ost.jobid
+    target_id  = job.target_id
+    vm         = VmOrTemplate.find(target_id)
+    unless vm.kind_of?(ManageIQ::Providers::Amazon::CloudManager::Vm)
+      raise "Vm #{vm.name} of class #{vm.class.name} is not an Amazon vm - unable to sync data"
+    end
+    _log.debug("Syncing Metadata for #{vm.ems_ref}")
+    ost.taskid = job.guid
+    update_job_message(ost, "Synchronization in progress")
+    status        = scan_message         = "OK"
+    status_code   = categories_processed = 0
+    ost.xml_class = XmlHash::Document
+    bb            = last_err = nil
+
+    xml_summary = ost.xml_class.createDoc(:summary)
+    xml_node    = xml_node_scan = xml_summary.root.add_element("scanmetadata")
+    xml_node_scan.add_attributes("start_time" => extract_reply[:start_time])
+    xml_summary.root.add_attributes("taskid" => ost.taskid)
+    ost.taskid = ost.jobid
+
+    data_dir = File.expand_path(Rails.root.join("data/metadata"))
+    _log.debug("creating #{data_dir}")
+    begin
+      Dir.mkdir(data_dir)
+    rescue Errno::EEXIST
+      # Ignore if the directory was created by another thread
+      _log.debug("No need to create directory #{data_dir} since it exists.")
+    end unless File.exist?(data_dir)
+    ost.skipConfig = true
+    ost.config     = OpenStruct.new(
+      :dataDir            => data_dir,
+      :forceFleeceDefault => false
+    )
+
+    begin
+      require 'blackbox/VmBlackBox'
+      _log.debug("instantiating BlackBox")
+      bb = Manageiq::BlackBox.new(vm.guid, ost)
+      _log.debug("instantiated BlackBox")
+      categories = AmazonSsaSupport::SsaQueueExtractor::CATEGORIES
+      #
+      # The amazon_ssa_support gem defines the categories as an array,
+      # But the scanning_mixin expects it to be a comma-separated string.
+      #
+      ost.category = categories.join(",")
+
+      categories.each do |c|
+        ost.taskid = job.guid
+        update_job_message(ost, "Syncing #{c}")
+        ost.taskid = ost.jobid
+        st = Time.current
+        xml = MIQRexml.load(ost.reply[:categories][c.to_sym])
+        if xml.nil?
+          _log.debug("No XML loaded for [#{c}].")
+          next
+        elsif xml.root.nil?
+          _log.debug("No XML root loaded for [#{c}]: XML is #{xml.to_s}.")
+          next
+        end
+        _log.debug("Writing scanned data to XML for [#{c}] to blackbox.")
+        bb.saveXmlData(xml, c)
+        _log.debug("Writing XML complete.")
+
+        category_node = xml_summary.class.load(xml.root.shallow_copy.to_xml.to_s).root
+        category_node.add_attributes("start_time" => st.utc.iso8601, "end_time" => Time.now.utc.iso8601)
+        xml_node << category_node
+        categories_processed += 1
+      end
+    rescue NoMethodError => scan_err
+      last_err = scan_err
+      _log.error("perform_metadata_sync Error - [#{scan_err}]")
+      _log.error("perform_metadata_sync Error - [#{scan_err.backtrace.join("\n")}]")
+    ensure
+      bb.close if bb
+      ost.taskid = job.guid
+      update_job_message(ost, "Scanning completed.")
+      ost.taskid = ost.jobid
+      if last_err
+        status = "Error"
+        status_code  = 8
+        status_code  = 16 if categories_processed.zero?
+        scan_message = last_err.to_s
+        _log.error("ScanMetadata error status:[#{status_code}]:  message:[#{last_err}]")
+        _log.debug { last_err.backtrace.join("\n") }
+      end
+
+      xml_node_scan.add_attributes(
+        "end_time"    => Time.now.utc.iso8601,
+        "status"      => status,
+        "status_code" => status_code.to_s,
+        "message"     => scan_message
+      )
+      vm.save_metadata_op(MIQEncode.encode(xml_summary.to_xml.to_s), "b64,zlib,xml", job.guid)
+      _log.info("Completed: Sending scan summary to server.  TaskId:[#{ost.taskid}]  target:[#{vm.name}]")
+    end
+  end
+
+  def response_thread_sleep_seconds
+    @response_thread_sleep_seconds ||= Settings.ems.ems_amazon.agent_coordinator.response_thread_sleep_seconds
+  end
+end
--- a/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
@@ -77,14 +77,22 @@ module ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner::ResponseThre
     ost.jobid = extract_reply[:job_id]
     job       = Job.find_by(:id => ost.jobid)
     raise _("Unable to sync data for job with id <%{number}>. Job not found.") % {:number => ost.jobid} if job.nil?
-    ost.taskid = ost.jobid
     target_id  = job.target_id
     vm         = VmOrTemplate.find(target_id)
-    unless vm.kind_of?(ManageIQ::Providers::Amazon::CloudManager::Vm)
-      raise "Vm #{vm.name} of class #{vm.class.name} is not an Amazon vm - unable to sync data"
+    ost.taskid = job.guid
+    unless vm.kind_of?(ManageIQ::Providers::Amazon::CloudManager::Vm) ||
+           vm.kind_of?(ManageIQ::Providers::Amazon::CloudManager::Template)
+      error = "Vm #{vm.name} of class #{vm.class.name} is not an Amazon instance or image - unable to sync data"
+      update_job_message(ost, error)
+      job.signal(:abort, error, "error")
+      return
     end
     _log.debug("Syncing Metadata for #{vm.ems_ref}")
-    ost.taskid = job.guid
+    if extract_reply[:error]
+      update_job_message(ost, extract_reply[:error])
+      job.signal(:abort, extract_reply[:error], "error")
+      return
+    end
     update_job_message(ost, "Synchronization in progress")
     status        = scan_message         = "OK"
     status_code   = categories_processed = 0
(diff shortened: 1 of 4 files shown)


jerryk55:
@miq-bot add_label WIP

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,190 @@
...
+      sleep(response_check_sleep_seconds) unless @shutdown_instance_wait_thread
+    end
+  end
+
+  def perform_metadata_sync(extract_reply)
+    ost       = OpenStruct.new
+    ost.reply = extract_reply
+    ost.jobid = extract_reply[:job_id]
+    job       = Job.find_by(:id => ost.jobid)
+    raise _("Unable to sync data for job with id <%{number}>. Job not found.") % {:number => ost.jobid} if job.nil?
```
roliveri: Why not just use `#{ost.jobid}`?

app/models/manageiq/providers/amazon/cloud_manager/vm_or_template_shared/scanning.rb
```diff
@@ -39,117 +41,18 @@ def perform_metadata_scan(ost)
...
-        raise "Unable to get reply to scan request from instance: #{err}"
-      end
-    end
-    raise "No reply received for scan request from instance #{ems_ref}" unless extract_reply
-    ost.reply = extract_reply
-    perform_metadata_sync(ost)
+    msg = "Scanning in AWS region #{ssaq_args[:region]} in progress"
+    update_job_message(ost, msg)
+    job = Job.find_by(:id => ost.jobid)
+    raise _("Unable to process data for job with id <%{number}>. Job not found.") % {:number => ost.jobid} if job.nil?
```
roliveri: Same comment as above.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,190 @@
+module ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner::ResponseThread
+  require 'xml/xml_utils'
+  require 'scanning_mixin'
+  include ScanningMixin
+
+  #
+  # This thread startup method is called by the AgentCoordinatorWorker::Runner
+  # at the end of the do_before_work_loop method.
```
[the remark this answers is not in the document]
jerryk55: Actually no.  That's one reason it might be better to implement this as a worker.  If the AgentCoordinatorWorker dies then it will be restarted and the thread will too.  But if the worker lives and the thread dies then no.  I can look into adding something if you think that is something we should do, or compare and contrast that effort with that of making it a separate worker.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,190 @@
...
+  #
+  def start_response_thread
+    Thread.new do
+      _log.info("Starting AWS SSA Response Thread")
+      begin
+        # Wait a bit for the Agent Coordinators to be started
+        sleep(20)
+        wait_for_responses
```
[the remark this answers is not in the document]
jerryk55: I can change that.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,190 @@
...
+              _log.debug("Reply for #{coord.ems.name}: #{reply}")
+              perform_metadata_sync(reply)
+            end
+          rescue => err
+            _log.error("Error #{err} processing replies for #{coord.ems.name}.  Continuing.")
+            next
+          end
+        end
+      end
+      break if @shutdown_instance_wait_thread
```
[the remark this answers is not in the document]
jerryk55: We don't *need* it - but it does allow us to handle a thread shutdown without going to sleep for response_check_sleep_seconds first, then waking up and shutting down.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Some of this code was "acquired" from the VmScan and ScanningMixins code.  This line did look overly complicated and I will change it.

jerryk55:
@roliveri I've made all changes except removing the second check for @shutdown_instance_wait_thread and a mechanism for keeping the thread alive.  Let me know if you think we should do that in place or consider reimplementing as a separate worker.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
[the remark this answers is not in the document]
roliveri: But we won't sleep, there's an `unless`:
```ruby
sleep(response_check_sleep_seconds) unless @shutdown_instance_wait_thread
```

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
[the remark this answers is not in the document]
roliveri: Keep it as is for now. We'll address it at a later time.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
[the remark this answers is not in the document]
jerryk55: Oh!  Right!  Cool.  I'll remove the break.

jerryk55:
@roliveri the redundant break has been removed.  Unless there are other comments we can remove the WIP-specific lines and merge when the dependent PR https://github.com/ManageIQ/manageiq-providers-amazon/pull/318 is merged.

jerryk55:
@miq-bot remove_label WIP

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
+module ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner::ResponseThread
+  require 'xml/xml_utils'
+  require 'scanning_mixin'
```
Fryguy: :-1: modules should not be loaded this way.  Autoload should find this, otherwise require_nested / include_concern should be used instead.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
+module ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner::ResponseThread
+  require 'xml/xml_utils'
```
Fryguy: Is this the one from manageiq-gems-pending?  If so, this should say `'util/xml/xml_utils'` for consistency  (and so we can find it later to delete it)

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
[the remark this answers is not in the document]
Fryguy: We must track this thread and handle it dying within the worker context.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+  #
+  # This thread startup method is called by the AgentCoordinatorWorker::Runner
+  # at the end of the do_before_work_loop method.
+  #
+  def start_response_thread
+    Thread.new do
+      _log.info("Starting AWS SSA Response Thread")
+      begin
+        # Wait a bit for the Agent Coordinators to be started
+        sleep(20)
```
Fryguy: Is this really necessary?  Random sleep feels like a dangerous way to handle the timing.  Might be better to wait until we get a valid queue handle, then after that we start the response_handler_loop.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+  end
+
+  def stop_response_thread
+    @shutdown_instance_wait_thread = true
+  end
+
+  def response_handler_loop
+    @shutdown_instance_wait_thread = nil
+    until @shutdown_instance_wait_thread
+      @coordinators.each do |coord|
```
Fryguy: Is there a possibility of cross-thread races with the coordinator modifying the shared `@coordinators` and you using them?  I wonder if there should be a mutex around access to `@coordinators`...probably better to use a attr_accessor method at a minimum than hitting the instance variable directly.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+        if coord.reply_queue_empty?
+          _log.debug("No Replies visible for Provider #{coord.ems.name}")
+        else
+          ssaq_args = {}
+          ssaq_args[:ssa_bucket]    = coord.ssa_bucket
+          ssaq_args[:reply_queue]   = coord.reply_queue
+          ssaq_args[:request_queue] = coord.request_queue
+          ssaq_args[:region]        = coord.ems.provider_region
+          ssaq_args[:sqs]           = coord.sqs
+          ssaq_args[:s3]            = coord.s3
```
Fryguy: Prefer a hash literal instead of a hash, then one-by-one adding keys...

```ruby
ssaq_args = {
  :ssa_bucket    => coord.ssa_bucket,
  :reply_queue   => coord.reply_queue,
  :request_queue => coord.request_queue
  ...
}
```

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+          ssaq                      = AmazonSsaSupport::SsaQueue.new(ssaq_args)
+          unless ssaq
+            _log.error("Error creating SsaQueue for #{coord.ems.name}")
```
Fryguy: `.new` can't return a nil, so there's no way this line could get hit.  If anything it could raise an exception, which isn't being handled anyway.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+            next
+          end
+          _log.debug("Getting replies for #{coord.ems.name}")
+          begin
+            ssaq.reply_loop do |reply|
+              _log.debug("Reply for #{coord.ems.name}: #{reply}")
+              perform_metadata_sync(reply)
```
Fryguy: Should we use the event catcher approach where we don't actually process the data in the loop, but instead put it on an in-memory queue, and a separate thread walks that in-memory queue.  Only downside to that approach is that you are probably deleting the remote queue entries, so there's the possibility of losing a scan.  cc @agrare

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+        "status_code" => status_code.to_s,
+        "message"     => scan_message
+      )
+      vm.save_metadata_op(MIQEncode.encode(xml_summary.to_xml.to_s), "b64,zlib,xml", job.guid)
+      _log.info("Completed: Sending scan summary to server.  TaskId:[#{ost.taskid}]  target:[#{vm.name}]")
+    end
+  end
+
+  def response_thread_sleep_seconds
+    @response_thread_sleep_seconds ||= Settings.ems.ems_amazon.agent_coordinator.response_thread_sleep_seconds
```
Fryguy: There's no need to memoize this value.  In fact, that's bad because if the user changes the Settings, you probably want to live pick up the value.  Getting a value out of Settings has almost no overhead, also, so the caching is unnecessary.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+              _log.debug("Reply for #{coord.ems.name}: #{reply}")
+              perform_metadata_sync(reply)
+            end
+          rescue => err
+            _log.error("Error #{err} processing replies for #{coord.ems.name}.  Continuing.")
+            next
+          end
+        end
+      end
+      response_check_sleep_seconds = response_thread_sleep_seconds
```
Fryguy: If this is just a Settings check, then no need for the temporary variable.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
Fryguy: If this blows up because, say, it can't find a Job, it will blow out the entire reply_loop since the rescue is outside of the loop.  Then it will move on to the next coordinator.  It might be better to put the rescue *inside* the reply_loop so it only fails that one sync but then can move on to the next one.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+    ost.xml_class = XmlHash::Document
+    bb            = last_err = nil
+
+    xml_summary = ost.xml_class.createDoc(:summary)
+    xml_node    = xml_node_scan = xml_summary.root.add_element("scanmetadata")
+    xml_node_scan.add_attributes("start_time" => extract_reply[:start_time])
+    xml_summary.root.add_attributes("taskid" => ost.taskid)
+    ost.taskid = ost.jobid
+
+    data_dir = File.expand_path(Rails.root.join("data", "metadata"))
```
Fryguy: Prefer `Rails.root.join("data", "metadata").expand_path`

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+    _log.debug("creating #{data_dir}")
+    unless File.exist?(data_dir)
+      begin
+        Dir.mkdir(data_dir)
```
Fryguy: You probably want `FileUtils.mkdir_p` instead

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+      rescue Errno::EEXIST
+        # Ignore if the directory was created by another thread
+        _log.debug("No need to create directory #{data_dir} since it exists.")
```
Fryguy: This is an expensive log message in debug mode, since 99.99% of the time the directory will exist.

Fryguy: Also, it's not needed with mkdir_p, since that won't raise an expcetion.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+    bb            = last_err = nil
+
+    xml_summary = ost.xml_class.createDoc(:summary)
+    xml_node    = xml_node_scan = xml_summary.root.add_element("scanmetadata")
+    xml_node_scan.add_attributes("start_time" => extract_reply[:start_time])
+    xml_summary.root.add_attributes("taskid" => ost.taskid)
+    ost.taskid = ost.jobid
+
+    data_dir = File.expand_path(Rails.root.join("data", "metadata"))
+    _log.debug("creating #{data_dir}")
```
Fryguy: Also an expensive debug message.

In fact, this entire "build a data directory" could just be done at initialize time once, not in the hot loop.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+      :dataDir            => data_dir,
+      :forceFleeceDefault => false
+    )
+
+    begin
+      require 'blackbox/VmBlackBox'
+      _log.debug("instantiating BlackBox")
+      bb = Manageiq::BlackBox.new(vm.guid, ost)
+      _log.debug("instantiated BlackBox")
+      categories = AmazonSsaSupport::SsaQueueExtractor::CATEGORIES
```
Fryguy: @roliveri Is the eventual goal to use ScanProfiles as opposed to hardcoding things?

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+      # But the scanning_mixin expects it to be a comma-separated string.
+      #
+      ost.category = categories.join(",")
+
+      categories.each do |c|
+        ost.taskid = job.guid
+        update_job_message(ost, "Syncing #{c}")
+        ost.taskid = ost.jobid
+        st = Time.current
+        xml = MIQRexml.load(ost.reply[:categories][c.to_sym])
```
Fryguy: Do we *have* to use MIQRexml?  We're trying to get away from that (i.e. this is adding more technical debt)

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+        if xml.nil?
+          _log.debug("No XML loaded for [#{c}].")
```
Fryguy: Should this be info or even warn?  Same with the next error message.

app/models/manageiq/providers/amazon/cloud_manager/vm_or_template_shared/scanning.rb
```diff
@@ -26,10 +26,12 @@ def perform_metadata_scan(ost)
     @s3                    = ext_management_system.connect(connect_args)
     raise "Unable to obtain a new S3 resource" unless @s3
     ssaq_args              = {}
-    ssaq_args[:ssa_bucket] = ext_management_system.guid
-    ssaq_args[:region]     = ext_management_system.provider_region
-    ssaq_args[:sqs]        = @sqs
-    ssaq_args[:s3]         = @s3
+    ssaq_args[:ssa_bucket]    = "#{AmazonSsaSupport::DEFAULT_BUCKET_PREFIX}-#{ext_management_system.guid}"
+    ssaq_args[:request_queue] = "#{AmazonSsaSupport::DEFAULT_REQUEST_QUEUE}-#{ext_management_system.guid}"
+    ssaq_args[:reply_queue]   = "#{AmazonSsaSupport::DEFAULT_REPLY_QUEUE}-#{ext_management_system.guid}"
```
Fryguy: This key generation is duplicated in the amazon_ssa_agent.  It would be preferable if there were a method on AmazonSsaSupport that returned the value(s), thus keeping the key generation in a single place, in case we have to change it later.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
```diff
@@ -0,0 +1,187 @@
...
+          end
+        end
+      end
+      response_check_sleep_seconds = response_thread_sleep_seconds
+      _log.debug("going to sleep for #{response_check_sleep_seconds} seconds after checking all Agent Coordinators")
+      sleep(response_check_sleep_seconds) unless @shutdown_instance_wait_thread
+    end
+  end
+
+  def perform_metadata_sync(extract_reply)
```
Fryguy: I see most of this was a move, so something we can clean up later

Fryguy:
@jerryk55 Please see my post-merge comments above.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360 via a mutex.

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

app/models/manageiq/providers/amazon/agent_coordinator_worker/runner/response_thread.rb
jerryk55: Fixed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/360

PR #325, opened 2017-10-25
Include ResponseThread in Runner

jerryk55:
The ResponseThread "include" statement was inadvertantly left out of the Runner
for the AgentCoordinatorWorker when it was ported out of the ManageIQ repo
into the manageiq-provider-amazon repo in https://github.com/ManageIQ/manageiq-providers-amazon/pull/319.  This PR adds it back in.

@Fryguy you found this issue - this fixes it.  Please review and merge.
@roliveri @hsong-rh @bronaghs fyi - one line fix for an omitted line of code.

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
@@ -1,4 +1,6 @@
 class ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner < MiqWorker::Runner
+
+  include ResponseThread
   def do_before_work_loop
     @coordinators = self.class.all_agent_coordinators_in_zone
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
```diff
@@ -1,4 +1,5 @@
 class ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner < MiqWorker::Runner
+  include ResponseThread
```
Fryguy: If this is a nested concern module,  it's preferable to use `include_concern`.

Fryguy:
I'll merge for now, as this will un-break the current workers, but please fix the above in a follow up PR (probably with the other changes requested in #319 )

PR #326, opened 2017-10-25
Raise the intended message

bdunne:
raise ignores the block


bdunne:
@Fryguy please review

PR #327, opened 2017-10-25
chmod expects the first arg to be an integer

bdunne:


bdunne:
@Fryguy Please review

Fryguy:
This is fine...he is giving the integer in octal form (i.e.  `0xabc` is hex, `0b010101` is  binary, `0o400` is octal and `0400` with the leading zero is *also* octal).  I actually like this form for permissions.

bdunne:
I guess it looked like an accident to me.  I was expecting parameters similar to the `chmod` command.

PR #328, opened 2017-10-25
Cleanup

bdunne:
@Fryguy Please review

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -373,15 +373,15 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
   end
 
   def ssa_bucket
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -344,13 +333,10 @@ def create_pem_file(pair_name = keypair_name)
   end
 
   def create_config_yaml(yml = "config.yml")
-    defaults = agent_coordinator_settings.to_hash
-    defaults.delete("agent_ami_name")
-    defaults.delete("agent_docker_name")
-    defaults.delete("agent_label")
+    defaults = agent_coordinator_settings.to_hash.except("agent_ami_name", "agent_docker_name", "agent_label")
```
Fryguy: Does this create a Hash with String or Symbol?

PR #329, opened 2017-10-26
Various changes for SSA support

hsong-rh:
This PR contains the following changes:
1. Start AgentCoordinatorWorker only when role of 'smartproxy' is enabled and Amazon provider exists;
2. Add register flag for both docker register and Redhat downstream register;
3. Change KeyPair name into uniq, so different users from the same AWS account use their own key pair.
4. Other changes based on reviewers' comments.

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -192,7 +192,7 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
   end
 
   def docker_auth
-    @ems.authentications.find_by(:type => "Smartstate_Docker")
+    @ems.authentications.find_by(:type => "ssa_docker")
   end
 
   # Get Key Pair for SSH. Create a new one if not exists.
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -171,10 +171,15 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     # prepare work directory
     ssh.perform_commands(["sudo mkdir -p #{WORK_DIR}"])
     ssh.perform_commands(["sudo chmod go+w #{WORK_DIR}"])
+
     # scp the default setting yaml file
-    Tempfile.open('config.yml') do |config|
+    config = Tempfile.new('config.yml')
+    begin
       config.write(create_config_yaml)
-      scp_file(ip, agent_ami_login_user, auth_key, config, WORK_DIR)
+      config.close
+      out = scp_file(ip, agent_ami_login_user, auth_key, config.path, "#{WORK_DIR}/config.yml")
+    ensure
+      config.unlink
     end
 
     # docker register
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -57,30 +57,22 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
   def startup_agent
     agent_ids.empty? ? deploy_agent : activate_agents
   rescue => err
-    _log.error("No agent is set up to prcoess requests: #{err.message}")
+    _log.error("No agent is set up to process requests: #{err.message}")
     _log.error(err.backtrace.join("\n"))
   end
 
-  def scp_file(local_file, remote_file, instance, username = "centos")
-    ip = instance.public_dns_name
-    key_name = instance.key_name
-    auth_key = get_keypair(key_name).try(:auth_key)
-    _log.error("AuthKey [#{key_name}] is empty of #{instance.id}") if auth_key.nil?
+  private
+
+  def scp_file(ip, username, auth_key, local_file, remote_file)
     Net::SCP.upload!(ip, username, local_file, remote_file, :ssh => {:key_data => auth_key})
   rescue => err
     _log.error(err.message)
   end
 
-  def ssh_commands(instance, commands = [], username = "centos")
-    ip = instance.public_dns_name
-    key_name = instance.key_name
-    auth_key = get_keypair(key_name).try(:auth_key)
-    _log.error("AuthKey [#{key_name}] is empty of #{instance.id}") if auth_key.nil?
+  def ssh_commands(ip, username = "centos", auth_key = '', commands = [])
     ssh = LinuxAdmin::SSH.new(ip, username, auth_key)
     ssh.perform_commands(commands)
   end
-
-  private
 
   def agent_ids
     # reset to empty
@@ -152,43 +144,68 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
       :max_count            => 1,
       :min_count            => 1,
       :placement            => {:availability_zone => zone_name},
-      :security_group_ids   => [security_group_id],
-      :subnet_id            => subnets[0].subnet_id,
       :tag_specifications   => [{:resource_type => "instance", :tags => [{:key => "Name", :value => label}]}]
+      :network_interfaces   => [{
+        :associate_public_ip_address => true,
+        :delete_on_termination       => true,
+        :device_index                => 0,
+        :subnet_id                   => subnets[0].subnet_id,
+        :groups                      => [security_group_id]
+      }],
     ).first
     ec2.client.wait_until(:instance_status_ok, :instance_ids => [instance.id])
 
     _log.info("Start to load SSA application, this may take a while ...")
 
     setup_agent(instance)
-    _log.info("SSA agent is ready to receive requests.")
+    _log.info("Docker #{agent_docker_name} is loaded. Start to heartbeat.")
 
     instance.id
   end
 
   def setup_agent(instance)
-    # register if needed
-    register(instance)
+    # Somehow instance.public_dns_name is empty, need to reinitialize to get it back
+    ip = ec2.instance(instance.id).public_dns_name || raise("Failed to get agent's public ip!")
+    key_name = instance.key_name
+    auth_key = get_keypair(key_name).try(:auth_key)
+    _log.error("Key [#{key_name}] is missing. Cannot SSH to the agent:#{instance.id}") if auth_key.nil?
+
+    ssh = LinuxAdmin::SSH.new(ip, agent_ami_login_user, auth_key)
 
     # prepare work directory
-    ssh_commands(instance, ["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"], agent_ami_login_user)
+    ssh.perform_commands(["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"])
     # scp the default setting yaml file
     create_config_yaml("config.yml")
-    scp_file("config.yml", WORK_DIR.to_s, instance, agent_ami_login_user)
+    scp_file(ip, agent_ami_login_user, auth_key, "config.yml", WORK_DIR)
+    File.delete("config.yml")
+
+    # docker register
+    if agent_ami_registration_required?
+      raise "Need credentials to login" unless docker_auth
+
+      docker_username = docker_auth.userid
+      docker_password = docker_auth.password
+      command_line = "docker login"
+      command_line << " #{agent_docker_name}"
+      command_line << " -u #{docker_username} -p #{docker_password}"
+      ssh.perform_commands([command_line])
+    end
 
     # run docker image
     command_line = "sudo docker run -d --restart=always -v /dev:/host_dev -v #{WORK_DIR}/config.yml:#{WORK_DIR}/config.yml --privileged #{agent_docker_name}"
-    ssh_commands(instance, [command_line.to_s], agent_ami_login_user)
+    ssh.perform_commands([command_line])
   end
 
-  # TODO: for downstream
-  def register(instance)
+  def docker_auth
+    @ems.authentications.find_by(:type => "Smartstate_Docker")
   end
 
   # Get Key Pair for SSH. Create a new one if not exists.
-  def find_or_create_keypair(keypair_name = label)
+  def find_or_create_keypair(keypair_name = default_keypair_name)
     get_keypair(keypair_name) || begin
       _log.info("KeyPair #{keypair_name} will be created!")
+      # Delete from Aws if existing
+      ec2.key_pair(keypair_name).try(:delete)
       ManageIQ::Providers::CloudManager::AuthKeyPair.create_key_pair(@ems.id, :key_name => keypair_name)
     end
   end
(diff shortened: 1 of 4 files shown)


hsong-rh:
@roliveri @jerryk55 @Fryguy @bdunne Please review.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -144,68 +131,92 @@ def deploy_agent
...
+
+    # docker register
+    if agent_ami_registration_required?
+      raise "Need credentials to login" unless docker_auth
+
+      docker_username = docker_auth.userid
+      docker_password = docker_auth.password
+      command_line = "docker login"
+      command_line << " #{agent_docker_name}"
+      command_line << " -u #{docker_username} -p #{docker_password}"
```
bdunne: As Jason mentioned yesterday, we may want to shell escape the parameters to prevent command line injection attacks.  See `CommandLineBuilder#build` in AwesomeSpawn https://github.com/ManageIQ/awesome_spawn/blob/master/lib/awesome_spawn/command_line_builder.rb#L46

config/settings.yml
```diff
@@ -52,11 +52,11 @@
     :agent_coordinator:
       :agent_ami_name: CentOS Atomic Host 7 x86_64 HVM EBS 1706_01
       :agent_ami_login_user: centos
-      # TODO: will replace with official docker image name
-      :agent_docker_name: docker.io/hsong/ssa_support
+      :agent_ami_registration_required: false
+      :agent_docker_name: manageiq/amazon-smartstate:latest
```
bdunne: This should probably be named `:docker image`

config/settings.yml
```diff
@@ -52,11 +52,11 @@
     :agent_coordinator:
       :agent_ami_name: CentOS Atomic Host 7 x86_64 HVM EBS 1706_01
       :agent_ami_login_user: centos
-      # TODO: will replace with official docker image name
-      :agent_docker_name: docker.io/hsong/ssa_support
+      :agent_ami_registration_required: false
```
bdunne: We're not registering the ami, we're logging into a docker registry so maybe `:docker_login_required`?

bdunne: We'll also need a `docker_registry` or `docker_registry_url`

config/settings.yml
```diff
@@ -52,11 +52,11 @@
...
+      :agent_docker_name: manageiq/amazon-smartstate:latest
       :agent_idle_period: 900
       :agent_label: smartstate
-      :agent_log_level: INFO
+      :log_level: INFO
```
bdunne: Is this the log level of the agent coordinator or the fleecing code running on the remote machine?

hsong-rh: It's the fleecing code running on the remote agent.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -144,68 +131,92 @@ def deploy_agent
...
 
-  # TODO: for downstream
-  def register(instance)
+  def docker_auth
+    @ems.authentications.find_by(:type => "ssa_docker")
   end
 
   # Get Key Pair for SSH. Create a new one if not exists.
-  def find_or_create_keypair(keypair_name = label)
+  def find_or_create_keypair(keypair_name = default_keypair_name)
```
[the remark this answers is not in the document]
hsong-rh: It's better to have this parameter, so we can use it to diagnose any other key pairs.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -144,68 +131,92 @@ def deploy_agent
...
+    ssh = LinuxAdmin::SSH.new(ip, agent_ami_login_user, auth_key)
 
     # prepare work directory
-    ssh_commands(instance, ["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"], agent_ami_login_user)
+    ssh.perform_commands(["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"])
     # scp the default setting yaml file
     create_config_yaml("config.yml")
-    scp_file("config.yml", WORK_DIR.to_s, instance, agent_ami_login_user)
+    scp_file(ip, agent_ami_login_user, auth_key, "config.yml", WORK_DIR)
+    File.delete("config.yml")
```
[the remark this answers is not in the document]
Fryguy: :+1: for tempfile.  The file should also be deleted, so use the block form.

app/models/manageiq/providers/amazon/agent_coordinator.rb
Fryguy: :+1: this is a must

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -144,68 +131,92 @@ def deploy_agent
...
+    ip = ec2.instance(instance.id).public_dns_name || raise("Failed to get agent's public ip!")
+    key_name = instance.key_name
+    auth_key = get_keypair(key_name).try(:auth_key)
+    _log.error("Key [#{key_name}] is missing. Cannot SSH to the agent:#{instance.id}") if auth_key.nil?
+
+    ssh = LinuxAdmin::SSH.new(ip, agent_ami_login_user, auth_key)
 
     # prepare work directory
-    ssh_commands(instance, ["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"], agent_ami_login_user)
+    ssh.perform_commands(["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"])
```
bdunne: This array should be two strings, they're separate commands

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -144,68 +131,92 @@ def deploy_agent
...
     # run docker image
     command_line = "sudo docker run -d --restart=always -v /dev:/host_dev -v #{WORK_DIR}/config.yml:#{WORK_DIR}/config.yml --privileged #{agent_docker_name}"
-    ssh_commands(instance, [command_line.to_s], agent_ami_login_user)
+    ssh.perform_commands([command_line])
   end
 
-  # TODO: for downstream
-  def register(instance)
+  def docker_auth
+    @ems.authentications.find_by(:type => "ssa_docker")
```
Fryguy: I thought we were going with `smartstate_docker`  @jerryk55 ?

app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
```diff
@@ -1,8 +1,15 @@
 class ManageIQ::Providers::Amazon::AgentCoordinatorWorker < MiqWorker
   require_nested :Runner
 
+  include PerEmsWorkerMixin
```
Fryguy: This is going to do one-per-ems instead of one-per-ems-type, so I think this is the wrong choice.

Fryguy: We also have to ensure that there is only 1 *total* in the entire zone, even if smart_proxy is enabled on multiple appliances.  We usually do this via the role by making it zonal, but since you are overloading the smart_proxy role for this and it requires multiple workers for other purposes, we may have to do something clever where the query zone.miq_workers or something.  @agrare Thoughts here on how to implement this?

Fryguy: https://github.com/ManageIQ/manageiq/blob/481eec1cbaad1259a080f296b6a46c810f989c60/app/models/miq_vim_broker_worker.rb#L16-L20 is an example where the broker handles it, but we still have 1-per-appliance with the role, so we'll have to be *more* clever.

bdunne:
Should this be split into 3 PR's?
- Fix the auth_key issue
- Fix for the missing config file keys
- General cleanup

Fryguy:
> Should this be split into 3 PR's?
> 
> - Fix the auth_key issue
> - Fix for the missing config file keys
> - General cleanup

I agree, splitting into PRs will allow us to move forward much more quickly.

app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
hsong-rh: @Fryguy the overwrite method desired_queue_names here always return an array with one element. This will guarantee the worker number returns 1, if there is any amazon cloud managers.

app/models/manageiq/providers/amazon/agent_coordinator.rb
jerryk55: Oh.  That goes to show that I can't remember a name from one moment to the next.  I thought just the label on the Tab was going to be "SmartState Docker" and the authtype could be shortened....  I can change my code and then @hsong-rh can update this.  Apologies all.

app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
Fryguy: Ah..yes, that works too...very clever :+1:

app/models/manageiq/providers/amazon/agent_coordinator.rb
Fryguy: Yes, we will fix the name in a follow up PR

app/models/manageiq/providers/amazon/agent_coordinator.rb
Fryguy: We can do it in a follow up PR

roliveri:
@Fryguy Did you say you pushed commits to this PR? I don't see them.
Plus, it looks like it needs to be rebased.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -171,10 +171,15 @@ def setup_agent(instance)
...
     ssh.perform_commands(["sudo chmod go+w #{WORK_DIR}"])
+
     # scp the default setting yaml file
-    Tempfile.open('config.yml') do |config|
+    config = Tempfile.new('config.yml')
+    begin
       config.write(create_config_yaml)
-      scp_file(ip, agent_ami_login_user, auth_key, config, WORK_DIR)
+      config.close
+      out = scp_file(ip, agent_ami_login_user, auth_key, config.path, "#{WORK_DIR}/config.yml")
```
roliveri: It seems there is no need to assign the return value to `out`.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -171,10 +171,15 @@ def setup_agent(instance)
     # prepare work directory
     ssh.perform_commands(["sudo mkdir -p #{WORK_DIR}"])
     ssh.perform_commands(["sudo chmod go+w #{WORK_DIR}"])
+
     # scp the default setting yaml file
-    Tempfile.open('config.yml') do |config|
```
roliveri: why was this change made?

chessbyte:
@roliveri @hsong-rh are you splitting this into 3 PRs based on https://github.com/ManageIQ/manageiq-providers-amazon/pull/329#issuecomment-339718390 ?

chessbyte:
@roliveri can you connect ALL the Amazon Startstate PRs by providing links and dependencies in the description of each PR.

hsong-rh:
@chessbyte Discussed this with Jason yesterday, I don't need split this into three, because most changes were tested already.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Yes, forgot to remove it. I'll change it in next PR.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Two reasons: 1. need delete the temp file; 2. the generate temp file just prefix with name I given, need right name in the remote side.

roliveri: 1. isn't the temp file deleted when you leave the `Tempfile.open('config.yml') do |config|` block?
2. The name of the temp file on the appliance doesn't need to have any relation to the name on the remote side. You just need to scp it to the name you want.

hsong-rh: 1. No, I thought it too, but the block only sure to close it. 
2. Yes, now I give the exact name in remote side.

roliveri:
@chessbyte That's the only other open PR I'm aware of at the moment. I think Jason mentioned he had a PR in another repo, but I haven't seen it. @Fryguy can you `@` me on that PR?

jerryk55:
@chessbyte @roliveri well except for the UI changes PR.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -152,43 +140,75 @@ def deploy_agent
...
   end
 
   def setup_agent(instance)
-    # register if needed
-    register(instance)
+    # Somehow instance.public_dns_name is empty, need to reinitialize to get it back
+    ip = ec2.instance(instance.id).public_dns_name || raise("Failed to get agent's public ip!")
+    key_name = instance.key_name
+    auth_key = get_keypair(key_name).try(:auth_key)
+    _log.error("Key [#{key_name}] is missing. Cannot SSH to the agent:#{instance.id}") if auth_key.nil?
```
bdunne: Should we raise?  The next call won't work if the `auth_key` is `nil`

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -152,43 +140,75 @@ def deploy_agent
...
+
+    ssh = LinuxAdmin::SSH.new(ip, agent_ami_login_user, auth_key)
 
     # prepare work directory
-    ssh_commands(instance, ["sudo mkdir -p #{WORK_DIR}; sudo chmod go+w #{WORK_DIR}"], agent_ami_login_user)
+    ssh.perform_commands(["sudo mkdir -p #{WORK_DIR}"])
```
bdunne: `perform_commands` accepts an array of commands to perform. I was suggesting that the array have 2 strings instead of a string with a `;` between the commands.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -152,43 +140,75 @@ def deploy_agent
...
+    ssh.perform_commands(["sudo chmod go+w #{WORK_DIR}"])
+
     # scp the default setting yaml file
-    create_config_yaml("config.yml")
-    scp_file("config.yml", WORK_DIR.to_s, instance, agent_ami_login_user)
+    config = Tempfile.new('config.yml')
```
bdunne: I thought you were going to use the block syntax so that you don't need to `unlink` at the end.

PR #336, opened 2017-10-27
:type is for STI, :authtype is a string

bdunne:


Fryguy:
Spec failures are on master, so merging.

PR #338, opened 2017-10-27
Settings

bdunne:


Fryguy:
Should it also be removed from here? https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/agent_coordinator.rb#L349

cc @hsong-rh ?

bdunne:
It feels like those should be two completely different hashes to me ¯\\\_(ツ)_/¯

Fryguy:
Spec failures are on master, so merging.

PR #342, opened 2017-10-30
Replace SSA with smartstate in log messages

hsong-rh:
In log messages, replace all 'SSA' with 'smartstate'.


hsong-rh:
@Fryguy @bdunne @roliveri @jerryk55 Please review.

PR #353, opened 2017-11-01
Combine registry and image together when registry needs.

hsong-rh:
Both registry and image are needed for docker to pull image from private docker repos.

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -206,7 +206,7 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     end
 
     # run docker image
-    image = docker_login_required? ? "#{docker_registry}/#{docker_image}" : docker_image
+    image = docker_registry.present? ? "#{docker_registry}/#{docker_image}" : docker_image
     command_line = "sudo docker run -d --restart=always -v /dev:/host_dev -v #{WORK_DIR}/config.yml:#{WORK_DIR}/config.yml --privileged #{image}"
     ssh.perform_commands([command_line])
   end
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -206,7 +206,8 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     end
 
     # run docker image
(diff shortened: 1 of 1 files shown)


hsong-rh:
@Fryguy @bdunne @roliveri @jerryk55 Please review. This should be backported to new G branch.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -206,7 +206,8 @@ def setup_agent(instance)
     end
 
     # run docker image
-    command_line = "sudo docker run -d --restart=always -v /dev:/host_dev -v #{WORK_DIR}/config.yml:#{WORK_DIR}/config.yml --privileged #{docker_image}"
+    image = docker_login_required? ? "#{docker_registry}/#{docker_image}" : docker_image
```
roliveri: @hsong-rh Is `docker_login_required?` the correct test here? I would think we' want to join them whenever `docker_registry` is not empty.

Fryguy: Correct, the proper test is when docker_registry is `.present?`

hsong-rh: docker_registry is only needed when docker_login_required? is true. I feel checking docker_login_required? is easier to understand.

bdunne: > docker_registry is only needed when docker_login_required? is true

That's not true, registries other than docker hub don't always require authentication.

hsong-rh: Changed to check docker_registry.present? for combination.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -206,7 +206,8 @@ def setup_agent(instance)
     end
 
     # run docker image
-    command_line = "sudo docker run -d --restart=always -v /dev:/host_dev -v #{WORK_DIR}/config.yml:#{WORK_DIR}/config.yml --privileged #{docker_image}"
+    image = docker_registry.present? ? "#{docker_registry}/#{docker_image}" : docker_image
```
bdunne: Any reason not to use `File.join(docker_registry, docker_image)`?

simaishi:

```
$ git log -1
commit ce1453eb9f7ec77ac8e0fa1c4a164d530b955479
Date:   Wed Nov 1 15:15:07 2017 -0400

    Merge pull request #353 from hsong-rh/combine_docker_registry_and_image
    
    Combine registry and image together when registry needs.
    (cherry picked from commit c83987284c6b234437565cdbcce86ef95c5cca60)
```

PR #354, opened 2017-11-01
Choose VPC with enabled DNS support to deploy SSA Agent

hsong-rh:
Amazon smartstate agent needs to have public DNS when it is deployed. This requires its VPC host must have DNS support enabled. 

This PR adds the logic to check existing VPCs, choose the one with DNS support enabled, then deploy an agent on it. 

https://github.com/ManageIQ/manageiq-providers-amazon/issues/355

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -230,7 +230,7 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
         return subnet if subnet
       end
     end
-    raise("No subnet is qualiified to deploy smartstate agent!")
+    raise("No subnet is qualified to deploy smartstate agent!")
   end
 
   # To run SSA, VPC needs to turn on enableDnsSupport and enableDnsHostnames
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -138,7 +138,7 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
 
     kp = find_or_create_keypair
     vpc_ids = get_dns_enabled_vpc_ids
-    raise "Smartstate analysis needs a VPC whose enablDnsHostnames setting is true!" if vpc_ids.empty?
+    raise "Smartstate analysis needs a VPC whose enableDnsSupport/enableDnsHostnames settings are true!" if vpc_ids.empty?
 
     zone_names = ec2.client.describe_availability_zones.availability_zones.map(&:zone_name)
     subnets = []
@@ -229,18 +229,18 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     @ems.authentications.find_by(:authtype => "smartstate_docker")
   end
 
-  # To run SSA, VPC needs to turn on enableDnsSupport and enableDnsHostname
+  # To run SSA, VPC needs to turn on enableDnsSupport and enableDnsHostnames
   def get_dns_enabled_vpc_ids
     vpc_ids = ec2.client.describe_vpcs.vpcs.map(&:vpc_id)
     vpc_ids.select { |id| enabled_dns_support?(id) && enabled_dns_hostnames?(id) }
   end
 
   def enabled_dns_hostnames?(vpc_id)
-    ec2.vpc(vpc_id).describe_attribute({:attribute => 'enableDnsHostnames', :vpc_id => vpc_id}).enable_dns_hostnames.value
+    ec2.vpc(vpc_id).describe_attribute(:attribute => 'enableDnsHostnames', :vpc_id => vpc_id).enable_dns_hostnames.value
   end
 
   def enabled_dns_support?(vpc_id)
-    ec2.vpc(vpc_id).describe_attribute({:attribute => 'enableDnsSupport', :vpc_id => vpc_id}).enable_dns_support.value
+    ec2.vpc(vpc_id).describe_attribute(:attribute => 'enableDnsSupport', :vpc_id => vpc_id).enable_dns_support.value
   end
 
   # Get Key Pair for SSH. Create a new one if not exists.
@@ -356,14 +356,16 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
 
   def get_subnets(az, vpc_id)
     ec2.client.describe_subnets(
-      :filters => [{
-        :name   => "availability-zone",
-        :values => [az]
-      },
-      {
-        :name   => "vpc-id",
-        :values => [vpc_id]
-      }]
+      :filters => [
+        {
+          :name   => "availability-zone",
+          :values => [az]
+        },
+        {
+          :name   => "vpc-id",
+          :values => [vpc_id]
+        }
+      ]
     ).subnets
   end
 
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -229,10 +229,18 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     @ems.authentications.find_by(:authtype => "smartstate_docker")
   end
 
-  # To run SSA, VPC needs to turn on enableDnsHostname
+  # To run SSA, VPC needs to turn on enableDnsSupport and enableDnsHostname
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -137,10 +137,26 @@ def deploy_agent
     @deploying = true
 
     kp = find_or_create_keypair
-    zone_name = ec2.client.describe_availability_zones.availability_zones[0].zone_name
-    subnets = get_subnets(zone_name)
-    raise "No subnet_id is available for #{zone_name}!" if subnets.empty?
-    security_group_id = find_or_create_security_group(subnets[0].vpc_id)
+
+    vpc_ids = get_dns_enabled_vpc_ids
+    raise "Smartstate analysis needs a VPC whose enablDnsHostnames setting is true!" if vpc_ids.empty?
```
jerryk55: Typo in "enablDnsHostnames" - should be "enableDnsHostnames"

hsong-rh: Thanks, I'll change it.

hsong-rh:
@jerryk55 @roliveri Please review.

roliveri:
@Fryguy @bdunne I'm thinking there may be better ways to query the DB for VPCs, etc.
Can you give it a look and let us know if that's true?

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -137,10 +137,26 @@ def deploy_agent
...
-    zone_name = ec2.client.describe_availability_zones.availability_zones[0].zone_name
-    subnets = get_subnets(zone_name)
-    raise "No subnet_id is available for #{zone_name}!" if subnets.empty?
-    security_group_id = find_or_create_security_group(subnets[0].vpc_id)
+
+    vpc_ids = get_dns_enabled_vpc_ids
+    raise "Smartstate analysis needs a VPC whose enableDnsSupport/enableDnsHostnames settings are true!" if vpc_ids.empty?
+
+    zone_names = ec2.client.describe_availability_zones.availability_zones.map(&:zone_name)
+    subnets = []
```
bdunne: Does this need to be an Array?  It looks like you're only using the first one.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -214,6 +230,20 @@ def docker_auth
     @ems.authentications.find_by(:authtype => "smartstate_docker")
   end
 
+  # To run SSA, VPC needs to turn on enableDnsSupport and enableDnsHostnames
+  def get_dns_enabled_vpc_ids
+    vpc_ids = ec2.client.describe_vpcs.vpcs.map(&:vpc_id)
```
bdunne: Why `map` and `select`?  You should be able to:
```ruby
ec2.client.describe_vpcs.vpcs.select do |vpc|
  id = vpc.vpc_id
  enabled_dns_support?(id) && enabled_dns_hostnames?(id)
end
```

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Changed.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Changed.

roliveri:
@bdunne @jerryk55 Have all your comment been addressed?

jerryk55:
@roliveri mine was.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -137,11 +137,28 @@ def deploy_agent
...
+    raise "Smartstate analysis needs a VPC whose enableDnsSupport/enableDnsHostnames settings are true!" if vpcs.empty?
+
+    zone_names = ec2.client.describe_availability_zones.availability_zones.map(&:zone_name)
+    subnet = nil
+    zone_names.each do |zone_name|
+      vpcs.each do |vpc|
+        subnet = get_subnets(zone_name, vpc.vpc_id).try(:first)
+        break if subnet
+      end
+      break if subnet
```
bdunne: Same here...
Actually, this logic looks like it should probably be its own method.
```ruby
private def something_about_finding_the_subnet
  ec2.client.describe_availability_zones.availability_zones.detect do |availability_zone|
    vpcs.detect { get_subnets(availability_zone.zone_name, vpc.vpc_id).try(:first) }
  end || raise("No subnet is qualiified to deploy smartstate agent!")
end
```

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -215,6 +232,22 @@ def docker_auth
...
 
+  # To run SSA, VPC needs to turn on enableDnsSupport and enableDnsHostnames
+  def get_dns_enabled_vpcs
+    ec2.client.describe_vpcs.vpcs.select do |vpc|
+      id = vpc.vpc_id
+      enabled_dns_support?(id) && enabled_dns_hostnames?(id)
+    end
+  end
+
+  def enabled_dns_hostnames?(vpc_id)
```
bdunne: Wouldn't it use fewer API calls if you pass a VPC object in to this method?  Then you wouldn't need to do the initial lookup here

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -215,6 +232,22 @@ def docker_auth
...
+    ec2.vpc(vpc_id).describe_attribute(:attribute => 'enableDnsHostnames', :vpc_id => vpc_id).enable_dns_hostnames.value
+  end
+
+  def enabled_dns_support?(vpc_id)
```
bdunne: Same

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: No, such a method will return us either vpc or availability_zone. What we need is subnet.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Yes, changed accordingly.

app/models/manageiq/providers/amazon/agent_coordinator.rb
bdunne: `detect` fails me every time 😢 but `return` would work.

```ruby
private def something_about_finding_the_subnet
  ec2.client.describe_availability_zones.availability_zones.each do |availability_zone|
    vpcs.each do
      subnet = get_subnets(availability_zone.zone_name, vpc.vpc_id).try(:first)
      return subnet if subnet
    end
  end
  raise("No subnet is qualified to deploy smartstate agent!")
end
```

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -137,11 +137,28 @@ def deploy_agent
...
+    zone_names = ec2.client.describe_availability_zones.availability_zones.map(&:zone_name)
+    subnet = nil
+    zone_names.each do |zone_name|
+      vpcs.each do |vpc|
+        subnet = get_subnets(zone_name, vpc.vpc_id).try(:first)
+        break if subnet
+      end
+      break if subnet
+    end
+    raise "No subnet is qualiified to deploy smartstate agent!" if subnet.nil?
```
bdunne: qualified

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Agreed! Will change as this, thanks!

simaishi:

```
$ git log -1
commit d1348859152c57bb3bf17fcf2733496ea6b38f44
Date:   Mon Nov 20 13:18:04 2017 -0500

    Merge pull request #354 from hsong-rh/vpc_attributes
    
    Choose VPC with enabled DNS support to deploy SSA Agent
    (cherry picked from commit f0e3bc4634ccfe4a870a983769fbdf87bce523a0)
```

bdunne:
Looks like the noise commits weren't squashed before merge 😟

PR #360, opened 2017-11-08
Suggested Changes from Initial Implementation

jerryk55:
At the time of the initial implementation of the
AgentCoordinatorWorker::Runner::ResponseThread several suggestions were made
post-merge by @fryguy that are being looked at now.
Those comments may be found here: https://github.com/ManageIQ/manageiq-providers-amazon/pull/319.

I am omitting the code to restart the thread if it dies at this point and will include it under a separate PR.  Some other comments were included in the original PR.

Those included here are:

1) Change the require of xml_utils for consistency.
2) Protect access to the @coordinators list with a Mutex.  Since the Runner
   modifies this list every so often while the ResponseThread reads from it
   between processing responses and sleeping the ResponseThread will lock all
   access to the list while not sleeping.  The Runner will only lock it when
   making changes to the list.
3) Change the rescue in the response handling loop so that if an exception occurs
   the next response for the same provider will be handled instead of skipping
   all other responses for the provider and checking on the next provider.
4) No need to memoize response_thread_sleep_seconds from the Settings, nor any need for
   a temporary variable to hold it.
5) a) Make the "build a data directory" code a method called once during initialization of
      the thread,
   b) No need for the rescue,
   c) Change Dir.mkdir to FileUtils.mkdir_p,
   d) Change File.expand_path(Rails.root.join("data", "metadata") to
      Rails.root.join("data", "metadata").expand_path
6) Change two debug log messages during loading of the XML from the response to info.

@Fryguy @roliveri Please review. Thanks.

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator_worker/runner.rb
@@ -1,7 +1,8 @@
 class ManageIQ::Providers::Amazon::AgentCoordinatorWorker::Runner < MiqWorker::Runner
   include_concern 'ResponseThread'
   def do_before_work_loop
-    @coordinators = self.class.all_agent_coordinators_in_zone
+    @coordinators       = self.class.all_agent_coordinators_in_zone
+    @coordinators_mutex = Mutex.new
     start_response_thread
   end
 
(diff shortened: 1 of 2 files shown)


Fryguy:
@jerryk55 Can you remove the @mention of me from the commit message?  The problem is I will get emails forever into the future anytime anyone pushes that commit to a new fork.

jerryk55:
@Fryguy done.

simaishi:

```
$ git log -1
commit 9b383838b07d6f61c01f2113e8abe0669052e0ae
Date:   Fri Nov 17 10:22:45 2017 -0500

    Merge pull request #360 from jerryk55/ssa_response_thread_followup
    
    Suggested Changes from Initial Implementation
    (cherry picked from commit d720def8d6cb32d6e734060cfdb84f532cc86429)
```

Issue #334, opened 2017-10-27
Failure during smartstate agent setup leaves a zombie instance and job


Fryguy:
If you start a scan job, and it tries to docker login when credentials are missing or one of the ssh commands fail, you get the error as expected in the log, but the job and the instance seems to be running forever.  This is because it never is part of the heartbeat check.  Even so, we should ensure that any failure during setup is caught directly and handled as opposed to relying on heartbeat.

djberg96:
@roliveri Any update on this?

roliveri:
@djberg96 I think this was resolved by Hui's PR:
https://github.com/ManageIQ/manageiq-providers-amazon/pull/367

@hsong-rh Is that correct? Should we just close this issue?

hsong-rh:
Rich, yes, this issue have been solved already in #367.

djberg96:
@hsong-rh Thanks for the update, closing.

PR #367, opened 2017-11-20
Add cleanup logic to remove zombie instance in unexpected cases

hsong-rh:
#334


hsong-rh:
@roliveri @jerryk55 @bdunne @Fryguy Please review.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -200,7 +200,7 @@ def setup_agent(instance)
 
     # docker register
     if docker_login_required?
-      raise "Need credentials to login" unless docker_auth
+      raise("Need credentials to login") unless docker_auth
```
bdunne: I don't think you need this change.

hsong-rh: Want to use the same format for raising exception in this method.

PR #382, opened 2017-12-21
Implement tag mapping in graph refresh

cben:
- [x] Depends on https://github.com/ManageIQ/manageiq/pull/16734 — merged

Subsumes #384, done with lots of help from Ladislav & Julian :earth_africa: 

https://bugzilla.redhat.com/show_bug.cgi?id=1506404

cc @bronaghs @djberg96 @Ladas @agrare (in case anybody still here, and happy holidays)

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -9,7 +9,6 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
     flavors
 
     # @mapper has to be initialized for parsing images & instances
-    # TODO: maybe do this in Persister?
     @tag_mapper = ContainerLabelTagMapping.mapper
 
     # The order of the below methods doesn't matter since they refer to each other using only lazy links
@@ -65,8 +64,8 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
 
       image_hardware(persister_image, image)
       image_operating_system(persister_image, image)
-      # TODO map labels to tags
       vm_and_template_labels(persister_image, image["tags"] || [])
+      vm_and_template_taggings(persister_image, map_labels("Image", image["tags"] || []))
     end
   end
 
@@ -99,6 +98,20 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
         :value   => tag["value"],
         :source  => 'amazon'
       )
+    end
+  end
+
+  # Returns array of InventoryObject<Tag>.
+  def map_labels(model_name, labels)
+    label_hashes = labels.collect do |tag|
+      {:name => tag["key"], :value => tag["value"]}
+    end
+    @tag_mapper.map_labels(model_name, label_hashes)
+  end
+
+  def vm_and_template_taggings(resource, tags_inventory_objects)
+    tags_inventory_objects.each do |tag|
+      persister.vm_and_template_taggings.build(:taggable => resource, :tag => tag)
     end
   end
 
@@ -208,41 +221,8 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
 
       instance_hardware(persister_instance, instance, flavor)
       instance_operating_system(persister_instance, instance)
-
-      labels = parse_labels(instance["tags"])
-      tags = map_labels_to_tags("Vm", labels)
-      vm_and_template_taggings(persister_instance, tags)
       vm_and_template_labels(persister_instance, instance["tags"] || [])
-    end
-  end
-
-  def parse_labels(entity)
-    parse_identifying_attributes(entity, 'labels')
-  end
-
-  def parse_identifying_attributes(attributes, section, source = "amazon")
-    result = []
-    return result if attributes.nil?
-      attributes.each do |tag|
-        custom_attr = {
-          :section => section,
-          :name    => tag["key"],
-          :value   => tag["value"],
-          :source  => source
-        }
-      result << custom_attr
-    end
-    result
-  end
-
-  def map_labels_to_tags(model_name, labels)
-    @tag_mapper.map_labels(model_name, labels)
-  end
-
-  # Conveniently, the tags map_labels emits are already in InventoryObject<Tag> form
-  def vm_and_template_taggings(parent, tags_inventory_objects)
-    tags_inventory_objects.each do |tag|
-      persister.vm_and_template_taggings.build(:taggable => parent, :tag => tag)
+      vm_and_template_taggings(persister_instance, map_labels("Vm", instance["tags"] || []))
     end
   end
 
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -8,11 +8,7 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
     # The order of the below methods does matter, because they are searched using find instead of lazy_find
     flavors
 
-    # @mapper has to be initialized for parsing images & instances
-    @tag_mapper = ContainerLabelTagMapping.mapper
-
     # The order of the below methods doesn't matter since they refer to each other using only lazy links
-    persister.collections[:tags_to_resolve] = @tag_mapper.tags_to_resolve_collection
     availability_zones
     key_pairs
     stacks
@@ -106,7 +102,7 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
     label_hashes = labels.collect do |tag|
       {:name => tag["key"], :value => tag["value"]}
     end
-    @tag_mapper.map_labels(model_name, label_hashes)
+    persister.tag_mapper.map_labels(model_name, label_hashes)
   end
 
   def vm_and_template_taggings(resource, tags_inventory_objects)
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -211,6 +211,7 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
 
       labels = parse_labels(instance["tags"])
       tags = map_labels_to_tags("Vm", labels)
+      vm_and_template_taggings(persister_instance, tags)
       vm_and_template_labels(persister_instance, instance["tags"] || [])
     end
   end
@@ -237,6 +238,14 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
   def map_labels_to_tags(model_name, labels)
     @tag_mapper.map_labels(model_name, labels)
   end
+
+  # Conveniently, the tags map_labels emits are already in InventoryObject<Tag> form
+  def vm_and_template_taggings(parent, tags_inventory_objects)
+    tags_inventory_objects.each do |tag|
+      persister.vm_and_template_taggings.build(:taggable => parent, :tag => tag)
+    end
+  end
+
   def instance_hardware(persister_instance, instance, flavor)
     persister_hardware = persister.hardwares.find_or_build(persister_instance).assign_attributes(
       :bitness              => architecture_to_bitness(instance['architecture']),
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -4,10 +4,16 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
   def parse
     log_header = "MIQ(#{self.class.name}.#{__method__}) Collecting data for EMS name: [#{collector.manager.name}] id: [#{collector.manager.id}]"
     $aws_log.info("#{log_header}...")
+
     # The order of the below methods does matter, because they are searched using find instead of lazy_find
     flavors
 
+    # @mapper has to be initialized for parsing images & instances
+    # TODO: maybe do this in Persister?
+    @tag_mapper = ContainerLabelTagMapping.mapper
+
     # The order of the below methods doesn't matter since they refer to each other using only lazy links
+    persister.collections[:tags_to_resolve] = @tag_mapper.tags_to_resolve_collection
     availability_zones
     key_pairs
     stacks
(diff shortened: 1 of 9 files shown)


cben:
@juliancheal @Ladas This is still WIP but I think something like this...

Depends on couple core changes, PR in a minute.

cben:
rebased

app/models/manageiq/providers/amazon/inventory_collection_default/cloud_manager.rb
```diff
@@ -176,6 +176,30 @@ def vm_and_template_labels(extra_attributes = {})
...
+        :manager_ref                  => [:taggable, :tag],
+        :parent_inventory_collections => [:vms, :miq_templates],
+        :inventory_object_attributes  => [
+          :taggable,
+          :tag,
+        ]
+      }
+
+      attributes[:targeted_arel] = lambda do |inventory_collection|
+        ids = inventory_collection.parent_inventory_collections.collect_concat(&:id)
```
cben: I was trying to "simplify" this to something like I did containers and avoid going through Vm.vm_and_template_taggings relation, just because it scared me.

But `add_inventory_collection` requires `:association => :foos` to be set, this determines the name of the `persister.foos` accessor to the collection :-(
Anyway, now that I added that relation in https://github.com/ManageIQ/manageiq/pull/16734 I should probably do what vm_and_template_labels does...

Ladas: ids are not available here yet, so we need to use
```manager_uuids = inventory_collection.parent_inventory_collections.collect(&:manager_uuids).map(&:to_a).flatten```

manager uuids are :ems_ref columns of Vm in this case. So we need a nested query that transforms that to ids

File: app/models/manageiq/providers/amazon/inventory_collection_default/cloud_manager.rb [hunk 2]
```diff
@@ -176,6 +176,30 @@ def vm_and_template_labels(extra_attributes = {})
...
+        Tagging.where(
+          'taggable_type' => 'VmOrTemplate',
+          'taggable_id'   => ids
```
Ladas: so then here it would be 
```
'taggable_id'  => inventory_collection.parent.vm_and_templates.where(:ems_ref => manager_uuids)
```

app/models/manageiq/providers/amazon/inventory_collection_default/cloud_manager.rb
```diff
@@ -190,11 +190,13 @@ def vm_and_template_taggings(extra_attributes = {})
...
 
       attributes[:targeted_arel] = lambda do |inventory_collection|
-        ids = inventory_collection.parent_inventory_collections.collect_concat(&:id)
+        manager_uuids = inventory_collection.parent_inventory_collections.collect(&:manager_uuids).map(&:to_a).flatten
+        ems = inventory_collection.parent
         Tagging.where(
           'taggable_type' => 'VmOrTemplate',
-          'taggable_id'   => ids
-        ).joins(:tag).merge(Tag.controlled_by_mapping)
+          'taggable_id'   => ems.vm_and_template_taggings.where(:ems_ref => manager_uuids)
```
Ladas: just 
```
ems.vm_and_templates.where(:ems_ref => manager_uuids)
```

since the manager uuis are fro ma table VmOrTemplate

spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/refresher_inventory_object_spec.rb
```diff
@@ -32,6 +33,7 @@
           end
 
           assert_common
+          assert_mapped_tags_on_vm
```
Ladas: could we move this inside of the specific Vms asserts?

Especially  assert_specific_vm_powered_on, assert_specific_vm_on_cloud_network_public_ip and assert_specific_vm_on_cloud_network

That way it will be automatically tested in old, new and targeted refresh

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -1416,9 +1416,27 @@ def assert_relationship_tree
     expect(@ems.descendants_arranged).to match_relationship_tree({})
   end
 
-  # TODO: Add some real specs here
-  # def assert_specific_tags_on_vm
-  #   vm = ManageIQ::Providers::Amazon::CloudManager::Vm.where(:name => "EmsRefreshSpec-PoweredOn-Basic3").first
-  #   expect(vm.tags).to be_empty
-  # end
+  def assert_specific_labels_on_vm
+    vm = ManageIQ::Providers::Amazon::CloudManager::Vm.find_by(:name => "ladas_test_5")
```
Ladas: We are usually testing only Vms having EmsRefreshSpec prefix, those are the ones nobody is touching (so we can re-record VCR easily)

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -4,10 +4,16 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
...
+
     # The order of the below methods does matter, because they are searched using find instead of lazy_find
     flavors
 
+    # @mapper has to be initialized for parsing images & instances
+    # TODO: maybe do this in Persister?
+    @tag_mapper = ContainerLabelTagMapping.mapper
+
     # The order of the below methods doesn't matter since they refer to each other using only lazy links
+    persister.collections[:tags_to_resolve] = @tag_mapper.tags_to_resolve_collection
```
Ladas: cool, you already use custom IC here, I was about to ask about it :-)

Ladas: @cben could we move the IC definition into the defaults, and the init into the Persitors?

app/models/manageiq/providers/amazon/inventory_collection_default/cloud_manager.rb
cben: thanks, fixed

cben:
Hooray!  I was stuck on mysterious hardware,disk,network count mismatches, turned out I dropped a line in merging conflicts.
This works now for VMs :tada: 
Adding images.

spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/refresher_inventory_object_spec.rb
cben: done for vms

For templates the only label I found at all is `Name` :neutral_face: 
and both `assert_specific_template*` have 0 labels (not even `Name`) so only calling this function explicitly in some tests.
This means targeted image taggings refresh is not well tested.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
cben: I could.  Note that constructing the Mapper actually reads whole mappings table.
- Is it OK to do it each Persister initialization?
- Will it still be re-read each refresh? (eg. if it's only read once when refresh worker starts that would be bad, changing mapping would have no effect).  
- Conversely, can re-reading it each targetted refresh be a performance problem?
- Will Persister initialization remain a good place with collector/persister split?

cben:
PTAL.  Added images, addressed most comments, except https://github.com/ManageIQ/manageiq-providers-amazon/pull/382#discussion_r159463356

Don't understand how refresher_other_region_spec.rb passes tags check in assert_specific_vm when I didn't `create_tag_mapping`?

Ladas:
@cben yes, another copy of assert_specific_vm_powered_on. I think the specs are great now, we can ignore the other region specs

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
Ladas: hum, ok, if it reads the whole mapping table, I will need to look on it more closely tomorrow. We will need to limit it only to Vms and Templates that are part of the targeted refresh

Ladas: - Is it OK to do it each Persister initialization?

Right now that would be the same as having it in parser, that is being called each refresh

- Will it still be re-read each refresh? (eg. if it's only read once when refresh worker starts that would be bad, changing mapping would have no effect).

Yes, right now, the persister defines ICs dynamically (that might change in the future, as some providers use static definition, but there can be always a dynamic part)

- Conversely, can re-reading it each targetted refresh be a performance problem?

Yes looking at this, it will definitely hit targeted refresh performance, with huge amount of tags present

- Will Persister initialization remain a good place with collector/persister split?

There will be some refactoring of the format, but the Persistor should remain mostly the same.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -4,10 +4,15 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
...
     $aws_log.info("#{log_header}...")
+
     # The order of the below methods does matter, because they are searched using find instead of lazy_find
     flavors
 
+    # @mapper has to be initialized for parsing images & instances
+    @tag_mapper = ContainerLabelTagMapping.mapper
+
     # The order of the below methods doesn't matter since they refer to each other using only lazy links
+    persister.collections[:tags_to_resolve] = @tag_mapper.tags_to_resolve_collection
```
Ladas: @cben ok, so continuing from https://github.com/ManageIQ/manageiq-providers-amazon/pull/382#discussion_r159463356

lets move this to persister init, we can put shared method to ManageIQ::Providers::Amazon::Inventory::Persister and just call it in Cloud and TargetCollection Persisters for now

And we should there add a TODO for perf, that this will always load all ContainerLabelTagMapping (or other?). But this table is fairly small? It's the Tagging that is huge, right?

cben: Mapper init is reading ContainerLabelTagMapping.all.  It's fairly small, 1 per mapping defined in UI, independent of inventory size and number of tags created/assigned.
Moving...

cben: was about to ask about TargetCollection then saw you already anticipated that :-)
Done, PTAL.

cben:
I believe the "targeting" of the targeted_arel is not really tested.  It does create the expected taggings, but are we sure it doesn't delete other taggings?
However, I'll leave this to you folks to test better if you wish.

Targeted query, as diff from normal `ems.vm_and_template_taggings` query:
```diff
 SELECT "taggings".* FROM "taggings"
 INNER JOIN "tags" ON "tags"."id" = "taggings"."tag_id"
 LEFT OUTER JOIN "classifications" ON "classifications"."tag_id" = "tags"."id" 
 INNER JOIN "vms" ON "taggings"."taggable_id" = "vms"."id" 
 WHERE "vms"."ems_id" = 12000000009214 
   AND "taggings"."taggable_type" = 'VmOrTemplate' 
   AND (("tags"."name" LIKE '/managed/amazon:%') OR ("tags"."name" LIKE '/managed/kubernetes:%'))
   AND "classifications"."read_only" = 't' 
   AND ("classifications"."parent_id" != 0) 
+  AND "taggings"."taggable_type" = 'VmOrTemplate' 
+  AND "taggings"."taggable_id" IN (SELECT "vms"."id" FROM "vms" WHERE "vms"."ems_id" = 12000000009214)
```

cben:
Hmm, that last SELECT seems wrong!  I don't see a `ems_ref` constraint on vms.  Debugging...

The `taggable_type = 'VmOrTemplate'` constraint is duplicate, I'm going to I drop it from targeted_arel.

cben:
OK, this was silly.  Turns out `record.association(args)` (note missing `.where`, see last commit) ignores the args, sort-of.
<details>

The reason associations accept args at all is side effect of a deprecated interface, good thing it's going away:
```
[4] pry(main)> ems.hosts
  Host Load (3.0ms)  SELECT "hosts".* FROM "hosts" WHERE "hosts"."ems_id" = $1  [["ems_id", 7]]

[5] pry(main)> ems.hosts(:ems_ref => 'ignored')
DEPRECATION WARNING: Passing an argument to force an association to reload is now deprecated and will be removed in Rails 5.1. Please call `reload` on the result collection proxy instead. (called from <main> at (pry):5)
  Host Load (0.4ms)  SELECT "hosts".* FROM "hosts" WHERE "hosts"."ems_id" = $1  [["ems_id", 7]]
```

</details>

----

Fixed targeted query, again as diff from `ems.vm_and_template_taggings` query:
```diff
 SELECT "taggings".* FROM "taggings" 
 INNER JOIN "tags" ON "tags"."id" = "taggings"."tag_id" 
 LEFT OUTER JOIN "classifications" ON "classifications"."tag_id" = "tags"."id" 
 INNER JOIN "vms" ON "taggings"."taggable_id" = "vms"."id" 
 WHERE "vms"."ems_id" = 12000000010061 
   AND "taggings"."taggable_type" = 'VmOrTemplate' 
   AND (("tags"."name" LIKE '/managed/amazon:%') OR ("tags"."name" LIKE '/managed/kubernetes:%'))
   AND "classifications"."read_only" = 't' 
   AND ("classifications"."parent_id" != 0)
+  AND "taggings"."taggable_id" IN (SELECT "vms"."id" FROM "vms" 
+                                   WHERE "vms"."ems_id" = 12000000010061 
+                                     AND "vms"."ems_ref" IN ('i-c72af2f6', 'ami-2051294a'))
```

I think this supports the need for a targeted-after-full-doesn't-delete spec :-), but will leave that to you.
I believe this is ready now.

app/models/manageiq/providers/amazon/inventory_collection_default/cloud_manager.rb
```diff
@@ -193,7 +193,7 @@ def vm_and_template_taggings(extra_attributes = {})
         manager_uuids = inventory_collection.parent_inventory_collections.collect(&:manager_uuids).map(&:to_a).flatten
         ems = inventory_collection.parent
         ems.vm_and_template_taggings.where(
-          'taggable_id'   => ems.vms_and_templates(:ems_ref => manager_uuids)
+          'taggable_id'   => ems.vms_and_templates.where(:ems_ref => manager_uuids)
```
Ladas: ah :-o

agrare: Ha that'll do it

juliancheal:
👍 thanks @cben

simaishi:

```
$ git log -1
commit 7bdb89b65ad4f39dfe65ab60ef62d99ea2f820cc
Date:   Thu Jan 4 14:31:04 2018 +0100

    Merge pull request #382 from cben/tag-mapping
    
    Implement tag mapping in graph refresh
    (cherry picked from commit 53cc7b01248aa5addc0e8d36d1f1dc152de75acf)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1531291
```

PR #383, opened 2018-01-02
Use proper nested references in parser

Ladas:
Depends on:

- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/375
- [x] https://github.com/ManageIQ/manageiq/pull/16659

Use proper nested references in parser. Adding nested
lazy_finds to correctly model a complex link.

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -199,6 +199,8 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
       name = get_from_tags(instance, :name)
       name = name.blank? ? uid : name
 
+      lazy_vm = persister.vms.lazy_find(uid)
+
       persister_instance = persister.vms.find_or_build(uid).assign_attributes(
         :uid_ems             => uid,
         :name                => name,
@@ -209,7 +211,14 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
         :flavor              => flavor,
         :genealogy_parent    => persister.miq_templates.lazy_find(instance['image_id']),
         :key_pairs           => [persister.key_pairs.lazy_find(instance['key_name'])].compact,
-        :location            => persister.networks.lazy_find("#{uid}__public", :key => :hostname, :default => 'unknown'),
+        :location            => persister.networks.lazy_find({
+                                                               :hardware    => persister.hardwares.lazy_find(:vm_or_template => lazy_vm),
+                                                               :description => "public"
+                                                             },
+                                                             {
+                                                               :key     => :hostname,
+                                                               :default => 'unknown'
+                                                             }),
         :orchestration_stack => persister.orchestration_stacks.lazy_find(
           get_from_tags(instance, "aws:cloudformation:stack-id")
         ),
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -206,7 +215,7 @@ def instance_hardware(persister_instance, instance, flavor)
       :cpu_total_cores      => flavor[:cpus],
       :memory_mb            => flavor[:memory] / 1.megabyte,
       :disk_capacity        => flavor[:ephemeral_disk_size],
-      :guest_os             => persister.hardwares.lazy_find(instance['image_id'], :key => :guest_os),
+      :guest_os             => persister.hardwares.lazy_find(persister.vms.lazy_find(instance['image_id']), :key => :guest_os),
```
Ladas: so, I am still keeping the old way that we do not need to define a key, if the manager_ref is size 1

so
```persister.hardwares.lazy_find(persister.vms.lazy_find(instance['image_id']))```
is equivalent to:
```persister.hardwares.lazy_find(:vm_or_temaplte => persister.vms.lazy_find(:ems_ref => instance['image_id']))```

Or would you prefer just the hash syntax?

File: app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb [hunk 2]
```diff
@@ -206,7 +215,7 @@ def instance_hardware(persister_instance, instance, flavor)
       :cpu_total_cores      => flavor[:cpus],
       :memory_mb            => flavor[:memory] / 1.megabyte,
       :disk_capacity        => flavor[:ephemeral_disk_size],
-      :guest_os             => persister.hardwares.lazy_find(instance['image_id'], :key => :guest_os),
+      :guest_os             => persister.hardwares.lazy_find(persister.miq_templates.lazy_find(instance['image_id']), :key => :guest_os),
```
Ladas: @agrare  so, I am still keeping the old way that we do not need to define a key, if the manager_ref is size 1

so
```ruby
persister.hardwares.lazy_find(persister.miq_templates.lazy_find(instance['image_id']))
```
is equivalent to:
```ruby
persister.hardwares.lazy_find(:vm_or_template => persister.miq_templates.lazy_find(:ems_ref => instance['image_id']))
```

and instead of lazy_find, we can use the built object
```
persister.hardwares.lazy_find(persister.miq_templates.find_or_build(instance['image_id']))
```

Or would you prefer just the hash syntax?

Ladas: cc @cben what do you think ^ , I am now working on building Db queries automatically, using these (also for secondary indexes)

cben: OK, VmOrTemplate has_one hardware, hardwares manager_ref is [:vm_or_template].  Now I can read this ;-)

- Supporting both single value and hash makes sense, consistent with existing interface, though for hashes I prefer using `lazy_find_by` and eventually phasing out `lazy_find(hash)`.
- Allowing `find_or_build` too makes sense.  A major convenience of inventory objects is making actual records, records-to-be, and lazy refs all interchangeable.

`:key` with hash syntax gets a little tricky.  I don't think we'd want it mixed on same level:
```
persister.hardwares.lazy_find_by(
  :vm_or_template => persister.miq_templates.lazy_find(:ems_ref => instance['image_id']), 
  :key            => :guest_os  # confusing, this is not a column
)
```
but I think requiring a sub-hash would be fine:
```
persister.hardwares.lazy_find_by(
  {
    :vm_or_template => persister.miq_templates.lazy_find(:ems_ref => instance['image_id']),
    # would be a better example with a composite manager_ref of several columns
  }
  :key => :guest_os
)
```

Ladas: @cben right, so this kind of sucks with kwargs

persister.hardwares.lazy_find_by(
  :vm_or_template => persister.miq_templates.lazy_find(:ems_ref => instance['image_id']), 
  :key            => :guest_os  # confusing, this is not a column
)

Would be taken as the 1 argument is Hash, not sure how to enforce and assert to pass a subhash.

Also, then plan was to get rid of _by methods, right now those are just calling the normal methods.  (lazy_find_by just calls lazy_find). So maybe we can keep the _by variant after all? With smart asserts, we could avoid it being confusing like it is now. @agrare what do you think?

PR #384, opened 2018-01-02
[WIP] Using the new tag mapping for graph refresh

juliancheal:
Fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1506404

Tag mapping is missing for Amazon Graph refresh.

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -221,10 +221,8 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
     return result if attributes.nil?
       attributes.each do |tag|
         custom_attr = {
-          :section => section,
           :name    => tag["key"],
           :value   => tag["value"],
-          :source  => source
         }
       result << custom_attr
     end
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -199,9 +199,50 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
         ),
       )
 
+      @tag_mapper = ContainerLabelTagMapping.mapper
+
+      labels = parse_labels(instance["tags"])
+      taggings = map_labels_to_tags("Vm", labels)
+      map_taggings(persister_instance, taggings)
+
       instance_hardware(persister_instance, instance, flavor)
       instance_operating_system(persister_instance, instance)
+
       vm_and_template_labels(persister_instance, instance["tags"] || [])
+    end
+  end
+
+  def parse_labels(entity)
+    parse_identifying_attributes(entity, 'labels')
+  end
+
+  def parse_identifying_attributes(attributes, section, source = "amazon")
+    result = []
+    return result if attributes.nil?
+      attributes.each do |tag|
+        custom_attr = {
+          :section => section,
+          :name    => tag["key"],
+          :value   => tag["value"],
+          :source  => source
+        }
+      result << custom_attr
+    end
+    result
+  end
+
+  def map_labels_to_tags(model_name, labels)
+    @tag_mapper.map_labels(model_name, labels)
+  end
+
+  # Conveniently, the tags map_labels emits are already in InventoryObject<Tag> form
+  def map_taggings(parent, tags_inventory_objects)
+    model_name = parent.inventory_collection.model_class.base_class.name
+    key = [:taggings_for, model_name]
+    collection = @inv_collections[key]
+    raise("can't save: missing @inv_collections[#{key}]") if collection.nil?
+    tags_inventory_objects.each do |tag|
+      collection.build(:taggable => parent, :tag => tag)
     end
   end
 


juliancheal:
I'm currently getting this error.

`EmsRefresh::Refreshers::EmsRefresherMixin::PartialRefreshError: undefined method `[]' for nil:NilClass` from `manageiq/app/models/ems_refresh/refreshers/ems_refresher_mixin.rb:50:in `refresh'`

juliancheal:
I'm not sure if I'm parsing the tags into the correct format expected? @Ladas @cben

cben:
> manageiq/app/models/ems_refresh/refreshers/ems_refresher_mixin.rb:50:in `refresh'

Line 50 is notorious as a front, hiding all the shady business going on in refresh :-) :door: :slot_machine:
Try setting `EmsRefresh.debug_failures = true` to see the actual exception.

I'll also run this now.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -191,11 +191,49 @@ def instances
...
+
+  def parse_identifying_attributes(attributes, section, source = "amazon")
+    result = []
+    return result if attributes.nil?
+      attributes.each do |tag|
+        custom_attr = {
+          :section => section,
+          :name    => tag["key"],
+          :value   => tag["value"],
+          :source  => source
```
cben: If you *only* use this function to feed map_labels, you only need :name and :value.

juliancheal:
> Line 50 is notorious as a front, hiding all the shady business going on in refresh :-) 🚪 🎰
> Try setting EmsRefresh.debug_failures = true to see the actual exception.
> 
> I'll also run this now.

Ok so it's showing the error as

```
NoMethodError: undefined method `[]' for nil:NilClass
```

```
/manageiq-providers-amazon/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb:231:in `map_taggings'
```

juliancheal:
In case it helps, this is the data I'm passing to `map_taggings`


parent:
```
InventoryObject:('i-07cf95973c3875be3', InventoryCollection:<ManageIQ::Providers::Amazon::CloudManager::Vm>, blacklist: [genealogy_parent])
```
tags_inventory_objects:
```
[InventoryObject:('141__yipikaye__yipikaye', InventoryCollection:<Tag>)]
```

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
juliancheal: Ok thanks!

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -199,12 +199,51 @@ def instances
...
+
+  def map_labels_to_tags(model_name, labels)
+    @tag_mapper.map_labels(model_name, labels)
+  end
+
+  # Conveniently, the tags map_labels emits are already in InventoryObject<Tag> form
+  def map_taggings(parent, tags_inventory_objects)
+    model_name = parent.inventory_collection.model_class.base_class.name
+    key = [:taggings_for, model_name]
+    collection = @inv_collections[key]
```
Ladas: there are no @inv_collections[key]

first, you need to add the inv collection as @cben  is doing it https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/162

then the collection will be here as in ```@collections```

then call it e.g. as ```persister.taggings_for_vm```

Ladas:
@juliancheal please look closely at https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/162 and also add commits with specs @cben created https://github.com/ManageIQ/manageiq-providers-amazon/pull/382/files

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
Ladas: the inventory collections should be added here https://github.com/Ladas/manageiq-providers-amazon/blob/f9123d84236a66348c4bd05d499c74f94f6d7594/app/models/manageiq/providers/amazon/inventory/persister/cloud_manager.rb#L26

Ladas: btw. ```NoMethodError: undefined method `[]' for nil:NilClass``` is because @inv_collections are nil

We have @collections here, but we should not access it directly, that is why we call it as persister.collection_name

But unless you place it on @collections[collection.name] = InventoryCollection.new..., it will not be there

cben: I think the taggings collections can be added in app/models/manageiq/providers/amazon/inventory_collection_default/cloud_manager.rb similar to `vm_and_template_labels` (and called from https://github.com/Ladas/manageiq-providers-amazon/blob/f9123d84236a66348c4bd05d499c74f94f6d7594/app/models/manageiq/providers/amazon/inventory/persister/cloud_manager.rb#L5)
playing with that now...

Also need to add `persistor.add_inventory_collection(@tag_mapper.tags_to_resolve_collection)` somewhere.

Ladas: @cben yes, https://github.com/Ladas/manageiq-providers-amazon/blob/f9123d84236a66348c4bd05d499c74f94f6d7594/app/models/manageiq/providers/amazon/inventory/persister/cloud_manager.rb#L5 is a good place to have shared

Then we need to add it to cloud persitor:
https://github.com/Ladas/manageiq-providers-amazon/blob/f1a67bdf880ad381231e234962284c49717f5ea7/app/models/manageiq/providers/amazon/inventory/persister/cloud_manager.rb#L3

And also to target refresh persistor:
https://github.com/Ladas/manageiq-providers-amazon/blob/f1a67bdf880ad381231e234962284c49717f5ea7/app/models/manageiq/providers/amazon/inventory/persister/target_collection.rb#L2

For targeted refresh, we will need to add targeted_arel and parent_inventory_collections. Just enhanced arel condition we already have, it will look like  https://github.com/Ladas/manageiq-providers-amazon/blob/f1a67bdf880ad381231e234962284c49717f5ea7/app/models/manageiq/providers/amazon/inventory_collection_default/cloud_manager.rb#L143

(targeted refresh specs share asserts with full refresh, so we should see it fail with missing tag)

juliancheal:
calling, `persister.taggings_for_vm` just results in
```
NoMethodError: undefined method `taggings_for_vm for
<ManageIQ::Providers::Amazon::Inventory::Persister::CloudManager
```

juliancheal:
I've added @cben's specs to my PR and implemented more of the code you both suggested.

cben:
I think you forgot to git push?
(we're somewhat duplicating efforts, but I don't mind, in the end we'll converge to something that (1) works (2) we both understand ;-)

juliancheal:
@cben I think I'm following along. Your PR https://github.com/ManageIQ/manageiq-providers-amazon/pull/382 seems much better now than mine :(

juliancheal:
Superseded by https://github.com/ManageIQ/manageiq-providers-amazon/pull/382

PR #380, opened 2017-12-20
[WIP] Adds spec for cloud network with availability zone

d-m-u:
This is a test for a bug fix where availability zone wasn't being updated after selection of cloud network. 

expected failure is expected cause the code that makes this test pass is in linked pr
## depends on
Test for https://github.com/ManageIQ/manageiq/pull/16688

PR #391, opened 2018-01-11
Changes allowed cloud network to be updated by zone

d-m-u:
Per discussion of #16804, I think this should be reverted to the way it was before 16688 broke the provider subclass code

Previously, cloud network list was only updated after selection of security group, or change of tab, or a select set of other actions that completely voided the list of parameters passed to the filtering method. This adds a change so that availability zone updates the selection of cloud network, preventing incorrect cloud networks from showing up in the after the zone dropdown changes. This code should live in the specific provider that needs it though and that seems to only be Amazon afaik.

Should fix broken test in main and [UI issue that Roman brought up](https://github.com/ManageIQ/manageiq/pull/16804) (that 16804 also fixes but probably not correctly)

## related to:
https://github.com/ManageIQ/manageiq/pull/16806

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/provision_workflow.rb
@@ -41,6 +41,12 @@ class ManageIQ::Providers::Amazon::CloudManager::ProvisionWorkflow < ManageIQ::P
     end
   end
 
+  def allowed_cloud_networks(_options = {})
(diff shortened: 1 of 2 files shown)


d-m-u:
@syncrou

d-m-u:
@miq-bot add_label provisioning

d-m-u:
tests are here: https://github.com/ManageIQ/manageiq-providers-amazon/pull/380

d-m-u:
@miq-bot assign @gmcculloug

Issue #374, opened 2017-11-30
Memory metrics on AWS are not captured


rbergami:
AWS provider doesn't capture data related to memory usage of instances. Data is not showed in the UI.

[1] manageiq-providers-amazon/app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb

djberg96:
Was added by https://github.com/ManageIQ/manageiq-providers-amazon/pull/393.

PR #393, opened 2018-01-15
AWS memory usage collection

AlexanderZagaynov:
Metrics collection for memory usage.
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1519872
GH issue: https://github.com/ManageIQ/manageiq-providers-amazon/issues/374

Notes:
- instance's internal counters setup: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/mon-scripts.html
- crontab line: `*/5 * * * * ~/aws-scripts-mon/mon-put-instance-data.pl --mem-util --swap-util --from-cron`


app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
```diff
@@ -1,66 +1,85 @@
...
+      :amazon_counters       => %w[NetworkIn NetworkOut].freeze,
       :calculation           => ->(*stats, interval) { stats.compact.sum / 1024.0 / interval },
-      :vim_style_counter_key => "net_usage_rate_average"
-    },
-  ]
+      :vim_style_counter_key => "net_usage_rate_average",
+    }.freeze,
 
-  COUNTER_NAMES = COUNTER_INFO.collect { |i| i[:amazon_counters] }.flatten.uniq
+    # Memory counters setup: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/mon-scripts.html
```
Ladas: @AlexanderZagaynov we could probably fetch also the swap ? Like https://github.com/Ladas/manageiq-providers-openstack/blob/cbb3191e2ef0a28143b1d3339bd900cb6c3c7d72/app/models/manageiq/providers/openstack/infra_manager/metrics_capture.rb#L58

Seems to be available

bronaghs:
Hi @AlexanderZagaynov 
We generally like to avoid comments in the code other than maybe `# TODO: `

AlexanderZagaynov:
@bronaghs @Ladas please review

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
```diff
@@ -54,4 +54,23 @@
       end
     end
   end
+
+  context 'counters presence' do
```
Ladas: I would like to see also ```perf_capture('realtime')``` with a test if there is some actual value stored into the DB. There are several columns in Metric table, I would like to see which are being filled (we should have at least total_ram filled somewhere)

app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
```diff
@@ -1,66 +1,96 @@
+# frozen_string_literal: true
+
 class ManageIQ::Providers::Amazon::CloudManager::MetricsCapture < ManageIQ::Providers::BaseManager::MetricsCapture
-  INTERVALS = [5.minutes, 1.minute]
+  INTERVALS = [5.minutes.freeze, 1.minute.freeze].freeze
```
Ladas: @AlexanderZagaynov btw, it's usually a good idea to have the rubocop fixes in another commit, while starting with a consistent style, just adding the feature. 

This PR will probably get backported, so then it can be less pain. In some cases I wouldn't even backport a commit with style changes, since there is possibility of more conflicts, since the style changes are across many lines.

Ladas:
@AlexanderZagaynov I would like to see specs for the saved values, just for sure. Also the metrics should show in the Timelines in the UI, under a VM, could you check it works as expect? 

Check e.g. https://github.com/Ladas/manageiq/blob/69de0eaac67bf270752e9665e93e7a59647a0500/product/charts/layouts/realtime_perf_charts/VmOrTemplate.yaml#L33

These are the configs for what the timelines should show, what columns, so we should test those are stored. 

We do not show the swap by default, in UI. Maybe we could do that as an enhancement later. @bronaghs ?

But otherwise it looks great, you are learning fast @AlexanderZagaynov . :+1:

bronaghs:
@Ladas - agreed, displaying swap can be done in another PR, I am not sure what is involved beyond making the change to: https://github.com/Ladas/manageiq/blob/69de0eaac67bf270752e9665e93e7a59647a0500/product/charts/layouts/realtime_perf_charts/VmOrTemplate.yaml#L33

AlexanderZagaynov:
changed:

@Ladas 
- key code changes are now in the first commit
- specs for the db values (`perf_capture`) added

@bronaghs 
- about Rubocop comment (on the `Time.now` line) - I removed it now. However Rubocop condemns it, not able to see that next line doing what it want.
- about constant deep freezing - I believe it is a good practice to make constants real constants. It should not provide any errors, otherwise it means we have some smell in some other code. It also potentially affects thread safety. If you don't mind, I would like to keep those changes.

simaishi:

```
$ git log -1
commit ccf96331b72dbee4d0ea1512732f6739f824bf0a
Date:   Wed Jan 31 09:40:53 2018 -0500

    Merge pull request #393 from AlexanderZagaynov/BZ-1519872
    
    AWS memory usage collection
    (cherry picked from commit 5909c162326b98cb1989527c4d2e4edb104a73c4)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1552288
```

AlexanderZagaynov:
@simaishi Sorry, I don't get your question. Do you mean technical aspect, or feature itself?
In the first case - need to check, do you want me to do that?
In the second one - I'm not that person who can answer it.

simaishi:
@AlexanderZagaynov The former, as this is being asked to be bakported to Fine branch in the BZ (so that answers the latter!). Please check if this PR can go to Fine branch as is, or if you need to create a separate PR.

simaishi:
Fine backport details:

```
$ git log -1
commit 263d69a1c7102d4147bdf90a4a4e77c0abd43784
Date:   Wed Jan 31 09:40:53 2018 -0500

    Merge pull request #393 from AlexanderZagaynov/BZ-1519872
    
    AWS memory usage collection
    (cherry picked from commit 5909c162326b98cb1989527c4d2e4edb104a73c4)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1570060
```

PR #398, opened 2018-01-18
Migrate model display names from locale/en.yml to plugin

mzazrivec:
This is a continuation of the work started in https://github.com/ManageIQ/manageiq/pull/16596 where we want to migrate display model names from locale/en.yml into particular models, where we'll use standard gettext. Main goal here is to make this one aspect of our application pluggable.

Core PR: https://github.com/ManageIQ/manageiq/pull/16836


bronaghs:
hi @mzazrivec - Once you rebase, the specs should pass.

mzazrivec:
@bronaghs Rebased, thanks.

Issue #404, opened 2018-01-30
Raise exceptions if SSH commands failed in docker running image


hsong-rh:
Amazon SSA uses SSH to connect to agent and docker run the commands to start container. If one of the command fails, an exception should be raised, so the following cleanup logic will remove the agent properly. Make sure no zombie instance is left in ec2 environment.

PR #405, opened 2018-01-30
Raise exceptions if SSH command fail

hsong-rh:
https://github.com/ManageIQ/manageiq-providers-amazon/issues/404

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -59,7 +59,6 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     agent_ids.empty? ? deploy_agent : activate_agents
   rescue => err
     _log.error("No agent is set up to process requests: #{err.message}")
-    _log.error(err.backtrace.join("\n"))
   end
 
   def cleanup_agents
@@ -105,6 +104,7 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     Net::SCP.upload!(ip, username, local_file, remote_file, :ssh => {:key_data => auth_key})
   rescue => err
     _log.error(err.message)
+    raise("Failed to copy #{local_file} to #{ip}:#{remote_file}")
   end
 
   def agent_ids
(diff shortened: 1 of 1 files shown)


hsong-rh:
@agrare @bronaghs @jerryk55 @roliveri Please review.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -59,7 +59,6 @@ def startup_agent
     agent_ids.empty? ? deploy_agent : activate_agents
   rescue => err
     _log.error("No agent is set up to process requests: #{err.message}")
-    _log.error(err.backtrace.join("\n"))
```
roliveri: Should we remove this backtrace, or just change it to `_log.error`?

hsong-rh: Most time this backtrace is duplicated with https://github.com/hsong-rh/manageiq-providers-amazon/blob/530f58b7bf0df538fe99859b0dd71cf6bf973efa/app/models/manageiq/providers/amazon/agent_coordinator.rb#L208

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -204,6 +204,11 @@ def deploy_agent
     _log.info("Docker #{docker_image} is loaded. Start to heartbeat.")
 
     instance.id
+  rescue => err
+    _log.error(err.backtrace.join("\n"))
```
roliveri: It seems we're logging the backtrace for the error, but we're not logging the actual error itself.
True, we do re-raise a new error with the original message, but that will be logged after the backtrace, possibly leading to confusion.

Will the re-raised error be rescued and logged as well? If so, maybe we should just log the backtrace there.
If you want to log the context at this point, I would log the `err.message` before the backtrace.

app/models/manageiq/providers/amazon/agent_coordinator.rb
roliveri: OK.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Changed as you suggested.

PR #409, opened 2018-02-14
Return created topic

Ladas:
Return created topic instead of log.info, which always returns true.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1545147


simaishi:
`gaprindashvili/yes` ?

PR #410, opened 2018-02-14
Fix refresh for stack without parameters

Ladas:
Fix refresh for stack without parameters and add specs for
more combinations of missing stack relations.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1544863

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -165,7 +165,7 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
   end
 
   def stack_parameters(persister_orchestration_stack, stack)
-    return unless stack['outputs']
+    return unless stack['parameters']
 
     stack['parameters'].each do |parameter|
       uid = compose_ems_ref(stack['stack_id'].to_s, parameter['parameter_key'])
--- a/spec/models/manageiq/providers/amazon/cloud_manager/specific_stubbed_refresher_spec.rb
+++ b/spec/models/manageiq/providers/amazon/cloud_manager/specific_stubbed_refresher_spec.rb
@@ -1,0 +1,114 @@
+require_relative '../aws_helper'
+require_relative '../aws_stubs'
+require_relative '../aws_refresher_spec_common'
+
+describe ManageIQ::Providers::Amazon::NetworkManager::Refresher do
+  include AwsRefresherSpecCommon
+  include AwsStubs
+
+  describe "refresh" do
+    before do
+      _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
+      @ems                 = FactoryGirl.create(:ems_amazon, :zone => zone)
+      @ems.update_authentication(:default => {:userid => "0123456789", :password => "ABCDEFGHIJKL345678efghijklmno"})
+      EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+    end
+
+    (AwsRefresherSpecCommon::ALL_GRAPH_REFRESH_SETTINGS + AwsRefresherSpecCommon::ALL_OLD_REFRESH_SETTINGS
+    ).each do |settings|
+      context "with settings #{settings}" do
+        before :each do
+          stub_refresh_settings(
+            settings.merge(
+              :get_private_images => true,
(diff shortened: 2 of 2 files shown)


gmcculloug:
@bzwei Please review

bzwei:
@Ladas The new spec you added is for the case of zero count. I assume you already have spec for a refresh with reasonable content.

Ladas:
@bzwei yeah, the stubs are all prepared, so adding just specific combination of data

@gmcculloug @d-m-u any idea what causes the workflow provision spec failure?

d-m-u:
@Ladas I've been seeing that error for a bunch of the provider repos today and I have no clue why.

Ladas:
@NickLaMuro do you think this can be causing the spec failures https://github.com/ManageIQ/manageiq/pull/16990 ?

NickLaMuro:
@Ladas Geez... I hope not... looking

NickLaMuro:
For reference, here is the error:

```
Failures:
  1) ManageIQ::Providers::Amazon::CloudManager::ProvisionWorkflow pass platform attributes to automate
     Failure/Error: loggers.each { |l| l.send(:add, *args, &block) }
       #<Double (anonymous)> received unexpected message :to_expanded_xml with (no args)
     # ./spec/manageiq/lib/vmdb/loggers/multicast_logger.rb:23:in `block in add'
     # ./spec/manageiq/lib/vmdb/loggers/multicast_logger.rb:23:in `add'
     # ./spec/manageiq/app/models/miq_request_workflow.rb:697:in `dialog_name_from_automate'
     # ./spec/manageiq/app/models/miq_provision_virt_workflow.rb:49:in `dialog_name_from_automate'
     # ./spec/manageiq/app/models/manageiq/providers/cloud_manager/provision_workflow.rb:135:in `dialog_name_from_automate'
     # ./app/models/manageiq/providers/amazon/cloud_manager/provision_workflow.rb:53:in `dialog_name_from_automate'
     # ./spec/manageiq/app/models/miq_request_workflow.rb:664:in `get_pre_dialogs'
     # ./spec/manageiq/app/models/miq_provision_virt_workflow.rb:25:in `initialize'
     # ./spec/models/manageiq/providers/amazon/cloud_manager/provision_workflow_spec.rb:19:in `new'
     # ./spec/models/manageiq/providers/amazon/cloud_manager/provision_workflow_spec.rb:19:in `block (2 levels) in <top (required)>'
```

@Ladas okay, so my guess is no, but I don't want to rule it out.  Does this fail locally?

NickLaMuro:
> so my guess is no

To clarify, the only real change in that PR was updating the default log level of the main loggers that `MulticastLogger` was monitoring, and having it match the `DEBUG` of `MulticastLogger` out of the gate.

It is possible that this was never working with `MulticastLogger`, if the logger here was always set to `DEBUG`, and the underlying logger was just swallowing the messages.  I will have to try and re-create.

gmcculloug:
@NickLaMuro I was able to recreate locally and if I roll back the changes from https://github.com/ManageIQ/manageiq/pull/16990 it passes.

NickLaMuro:
@gmcculloug Yeah, I am too.

Looks like it was as I was suggesting above.  The code in question where it is breaking is here:

https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_engine.rb#L302

I think this wasn't a problem previously because `debug` was no-oping in the logger, but now it isn't (like it should), but the `ws` object (which is your `described_class` in this case for this `stub_dialog` method:  https://github.com/ManageIQ/manageiq/blob/master/spec/support/workflow_spec_helper.rb#L4-L6) isn't stubbing enough.

I can work on a fix.

NickLaMuro:
Fix can be found here:  https://github.com/ManageIQ/manageiq/pull/17001

Will be needed probably on `gaprindashvili` as well.

simaishi:

```
$ git log -1
commit fa016dcfe314f159417311a8a90a43d61aea4e95
Author: Greg McCullough <gmccullo@redhat.com>
Date:   Wed Feb 14 16:18:50 2018 -0500

    Merge pull request #410 from Ladas/fix_refresh_for_stack_without_parameters
    
    Fix refresh for stack without parameters
    (cherry picked from commit 2df2defc30846440fc5220ebb913c0bacb0edfea)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1545425
```

PR #413, opened 2018-02-19
Fix deprecated use of `have_attributes`

cben:
ManageIQ has its own `have_attributes` which has diverged from upstream rspec's. It'll go away at some point, for now shows warning (see https://github.com/ManageIQ/manageiq/pull/16188):
```
Randomized with seed 13795
............................................................................
WARNING: Use of `have_attributes` with array access (:[]) is deprecated and will be removed shortly.
If you're matching attributes in hashes, use appropriate hash matchers instead (`include`, `eq`).
Called from /home/bpaskinc/miq/quota-scopes/plugins/manageiq-providers-amazon/spec/models/manageiq/providers/amazon/cloud_manager/orchestration_stack_spec.rb:119:in `block (5 levels) in <top (required)>'
```
Using `include` will work when we drop the custom matcher.


cben:
@bzwei @chrisarcand please review.

@miq-bot add-label test, technical debt

PR #422, opened 2018-03-15
Fixes to add a retry logic in EC2 instance creation when IAM role is not ready

hsong-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1554049

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -235,7 +235,7 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
     #
     # (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#launch-instance-with-role-console)
 
-    max_retries = 3
+    max_retries = 5
     begin
       instance = ec2.create_instances(
         :iam_instance_profile => {:name => label},
@@ -255,8 +255,8 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
         }],
       ).first
     rescue Aws::EC2::Errors::InvalidParameterValue => e
-      if max_retries > 0
-        sleep 5
+      if max_retries.positive?
+        sleep 10
         max_retries -= 1
         _log.warn("Will retry #{max_retries} times due to error: #{e.message}")
         retry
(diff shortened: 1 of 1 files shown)


hsong-rh:

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -225,23 +225,46 @@ def deploy_agent
...
+    #
+    # Important
+    #
+    # After you create an IAM role, it may take several seconds for the permissions to propagate.
+    # If your first attempt to launch an instance with a role fails, wait a few seconds before trying again.
+    # For more information, see Troubleshooting Working with Roles in the IAM User Guide.
+    #
+    # (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#launch-instance-with-role-console)
+
+    max_retries = 3
```
roliveri: You may want to add some more retries, 20 seconds may not be enough.

jerryk55:
@hsong-rh if your intent was to retry 3 times the code is wrong - you retry one more time than the max_retries variable (it does 4 the way the code is written).  You should change the logic ...  and maybe do 5 retries total while you're making the change?  Thanks.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Changed retry time to 5, totally wait for 60 seconds.

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -255,8 +255,8 @@ def deploy_agent
         }],
       ).first
     rescue Aws::EC2::Errors::InvalidParameterValue => e
-      if max_retries > 0
-        sleep 5
+      if max_retries.positive?
+        sleep 10
```
roliveri: I wouldn't change the sleep interval, just the retries. If the wait time is in the order of seconds, you want to wait less and retry more.

hsong-rh: Ok, changed sleep interval back to 5.

simaishi:

```
$ git log -1
commit 0b3d8444fac273728eb3f00ea2d057c1a1cf76ce
Date:   Fri Mar 16 15:39:54 2018 -0400

    Merge pull request #422 from hsong-rh/retry_for_iam_role
    
    Fixes to add a retry logic in EC2 instance creation when IAM role is not ready
    (cherry picked from commit 4fc85e465f7c2983d565f6cb4ae69c57f440dfcd)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1558040
```

PR #424, opened 2018-03-20
Fixes VPC selection to make sure it has internet gateway attached.

hsong-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1557377

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator.rb
@@ -335,7 +335,7 @@ class ManageIQ::Providers::Amazon::AgentCoordinator
   end
 
   def get_subnet_from_vpc_zone
-    vpcs = get_dns_enabled_vpcs
+    vpcs = validated_vpcs
     raise "Smartstate analysis needs a VPC whose enableDnsSupport/enableDnsHostnames settings are true!" if vpcs.empty?
 
     ec2.client.describe_availability_zones.availability_zones.each do |availability_zone|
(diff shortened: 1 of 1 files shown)


hsong-rh:

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -362,6 +362,29 @@ def enabled_dns_support?(vpc)
...
 
+  def enabled_internet_gateways?(vpc)
+    igw_ids = vpc.internet_gateways.map(&:internet_gateway_id)
+
+    unless igw_ids.empty? # Gateway attached
+      _log.debug("Found a gateway [#{igw_ids.first}] on VPC [#{vpc.id}]")
+
+      # Make sure gateway has vaild route
+      route_tables = vpc.route_tables.select do |route_table|
+        route_table.routes.any? { |route| route.gateway_id == igw_ids.first }
```
jerryk55: Can't there be more than one gateway?  It looks like you're only checking if the route's gateway matches the first gateway id - what if the route's gateway matches another one of the gateways on the VPC?

app/models/manageiq/providers/amazon/agent_coordinator.rb
```diff
@@ -362,6 +362,29 @@ def enabled_dns_support?(vpc)
...
+      end
+
+      unless route_tables.empty?
+        _log.debug("Found route tables #{route_tables} on the gateway [#{igw_ids.first}]")
+        subnets = route_tables.map { |rt| rt.associations.first.subnet_id }
```
jerryk55: I guess here we only care if there is one subnet so this use of "first" is ok, right?

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: It's a one to one relationship between VPC and gateway.

app/models/manageiq/providers/amazon/agent_coordinator.rb
hsong-rh: Yes, the API only has one entry per association returned.

jerryk55:
👍 LGTM

simaishi:

```
$ git log -1
commit 431b61aa9a1df03234e45f924c5ba91e7497682b
Date:   Wed Mar 21 14:33:00 2018 -0400

    Merge pull request #424 from hsong-rh/check_vpc_gateway
    
    Fixes VPC selection to make sure it has internet gateway attached.
    (cherry picked from commit 0f4f60b92327d9c5fa819518b5fd934035c29ac8)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1559609
```

PR #427, opened 2018-03-28
Fix firewall rules parse

tumido:
Update firewall rules parsing with:

- [x] Display :host_protocol correctly when "All" is selected
- [x] Enable IPv6 rules
- [x] Fix port ranges (_Update: fixed in https://github.com/ManageIQ/manageiq-ui-classic/pull/3710_ )
- [x] Update specs

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1540283
AWS Documentation: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IpPermission.html
Related: https://github.com/ManageIQ/manageiq-ui-classic/pull/3710

@miq-bot add_label bug, inventory

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/parser/network_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/network_manager.rb
@@ -101,7 +101,7 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::NetworkManager < ManageIQ:
   def firewall_rule(persister_security_group, perm, direction)
     common = {
       :direction             => direction,
-      :host_protocol         => perm['ip_protocol'].to_s.upcase,
+      :host_protocol         => perm['ip_protocol'].to_s == "-1" ? _("All") : perm['ip_protocol'].to_s.upcase,
       :port                  => perm['from_port'],
       :end_port              => perm['to_port'],
       :resource              => persister_security_group,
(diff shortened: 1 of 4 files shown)


tumido:
I need to go through the `from_port` and `to_port` definitions because it is really confusing. The API seems to use `-1` value in meaning of both "All" and "N/A" and I'm not sure how to deal with it yet.

Also, we display the `from_port` and `to_port` in UI, which doesn't make sense in cases like "All" or "N/A", a unified field in UI telling the port range would be more appropriate, what do you think?

Any suggestions anyone?

tumido:
@miq-bot add_label wip

tumido:
@miq-bot remove_label wip

tumido:
@miq-bot add_label gaprindashvili/yes

Please review :wink:

bronaghs:
Seems like you are enhancing or adding support for a number of scenarios, I think it would be worth writing some spec tests for them. Otherwise, looks great!
Also, see my [comment](https://github.com/ManageIQ/manageiq-ui-classic/pull/3710/files#r180542040) on the related classic UI PR.

tumido:
The ipv6 rules are already covered in our tests, they just had to be counted in, which I did already.

I'll add a testcase for the `:host_protocol => "All"`, thanks for pointing that out! :+1:

tumido:
Sorry I forgot about this PR, I'll add the test ASAP and we can merge it soon. :)

tumido:
Ready to be reviewed and merged. :tada:

PR #430, opened 2018-04-11
Taggings redoubling fix

AlexanderZagaynov:
I've found that `Tagging` records being doubled on each EMS refresh 
_(true for Amazon and Azure providers)_
Related core repo PR: https://github.com/ManageIQ/manageiq/pull/17287


AlexanderZagaynov:
I was erred, closing

spec/models/manager_refresh/inventory_collection_default_spec.rb
```diff
@@ -24,6 +26,36 @@
     end
   end
 
+  context 'label tag mapping' do
+    before(:context) do
```
cben: note that `before(:context)` aka `before(:all)`, runs outside the transaction that automatically cleans up DB changes made in `before(:each)` and inside example.
https://makandracards.com/makandra/11507-using-before-all-in-rspec-will-cause-you-lots-of-trouble-unless-you-know-what-you-are-doing

This means the Tag, Classification and ContainerLabelTagMapping will leak out of this test, may causing sporadic failures elsewhere depending on random test ordering seed...

if you must use `before(:context)`, add `after(:context)` to clean up.

AlexanderZagaynov: yeah, I noticed that too, thank you and @Ladas

PR #436, opened 2018-04-23
load regions list from aws-sdk gem's source

### LINKED ISSUE #395: Gather AWS regions list from AWS's GitHub repo
Regions list currently hard-coded in the https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/regions.rb#L8 file.

We can fetch it (either dynamically or at build time) from: https://github.com/aws/aws-sdk-ruby/blob/version-2/aws-sdk-core/endpoints.json#L12

AlexanderZagaynov:
Closes #395

### DIFF
--- a/app/models/manageiq/providers/amazon/regions.rb
+++ b/app/models/manageiq/providers/amazon/regions.rb
@@ -1,92 +1,15 @@
-# The aws-sdk gem can get us this information, however it talks to EC2 to get it.
-# For cases where we don't yet want to contact EC2, this information is hardcoded.
+# frozen_string_literal: true
+
+# NOTE: use 'aws:extract:regions' rake task to renew the list
+
+require 'yaml'
 
 module ManageIQ
   module Providers::Amazon
     module Regions
-      # From http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region
-      REGIONS = {
-        "us-east-1"      => {
-          :name        => "us-east-1",
-          :hostname    => "ec2.us-east-1.amazonaws.com",
-          :description => "US East (Northern Virginia)",
-        },
-        "us-east-2"      => {
-          :name        => "us-east-2",
-          :hostname    => "ec2.us-east-2.amazonaws.com",
-          :description => "US East (Ohio)",
-        },
-        "us-west-1"      => {
-          :name        => "us-west-1",
-          :hostname    => "ec2.us-west-1.amazonaws.com",
-          :description => "US West (Northern California)",
-        },
-        "us-west-2"      => {
-          :name        => "us-west-2",
-          :hostname    => "ec2.us-west-2.amazonaws.com",
-          :description => "US West (Oregon)",
-        },
-        "eu-west-1"      => {
-          :name        => "eu-west-1",
-          :hostname    => "ec2.eu-west-1.amazonaws.com",
-          :description => "EU (Ireland)",
-        },
-        "eu-west-2"      => {
-          :name        => "eu-west-2",
-          :hostname    => "ec2.eu-west-2.amazonaws.com",
-          :description => "EU (London)",
-        },
-        "eu-west-3"      => {
-          :name        => "eu-west-3",
-          :hostname    => "ec2.eu-west-3.amazonaws.com",
-          :description => "EU (Paris)",
-        },
-        "eu-central-1"   => {
-          :name        => "eu-central-1",
-          :hostname    => "ec2.eu-central-1.amazonaws.com",
-          :description => "EU (Frankfurt)",
-        },
-        "ap-south-1"     => {
-          :name        => "ap-south-1",
-          :hostname    => "ec2.ap-south-1.amazonaws.com",
-          :description => "Asia Pacific (Mumbai)",
-        },
-        "ap-southeast-1" => {
-          :name        => "ap-southeast-1",
-          :hostname    => "ec2.ap-southeast-1.amazonaws.com",
-          :description => "Asia Pacific (Singapore)",
-        },
-        "ap-southeast-2" => {
-          :name        => "ap-southeast-2",
-          :hostname    => "ec2.ap-southeast-2.amazonaws.com",
-          :description => "Asia Pacific (Sydney)",
-        },
-        "ap-northeast-1" => {
-          :name        => "ap-northeast-1",
-          :hostname    => "ec2.ap-northeast-1.amazonaws.com",
-          :description => "Asia Pacific (Tokyo)",
-        },
-        "ca-central-1"   => {
-          :name        => "ca-central-1",
-          :hostname    => "ec2.ca-central-1.amazonaws.com",
-          :description => "Canada (Central)",
-        },
-        "ap-northeast-2" => {
-          :name        => "ap-northeast-2",
-          :hostname    => "ec2.ap-northeast-2.amazonaws.com",
-          :description => "Asia Pacific (Seoul)",
-        },
-        "sa-east-1"      => {
-          :name        => "sa-east-1",
-          :hostname    => "ec2.sa-east-1.amazonaws.com",
-          :description => "South America (Sao Paulo)",
-        },
-        "us-gov-west-1"  => {
-          :name        => "us-gov-west-1",
-          :hostname    => "ec2.us-gov-west-1.amazonaws.com",
-          :description => "GovCloud (US)",
-        }
-      }
+      REGIONS = YAML.load_file(
+        ManageIQ::Providers::Amazon::Engine.root.join('db/fixtures/aws_regions.yml')
+      ).each_value(&:freeze).freeze
 
       def self.regions
         additional_regions = Hash(Settings.ems.ems_amazon.try!(:additional_regions)).stringify_keys
--- a/lib/tasks_private/regions.rake
+++ b/lib/tasks_private/regions.rake
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+namespace 'aws:extract' do
+  desc 'Get / renew regions list from online source of AWS gem'
+  task :regions do
+    require 'uri'
+    require 'json'
+    require 'yaml'
+    require 'net/http'
+
+    # path valid for aws-sdk v3 gem's source
+    uri = URI('https://raw.githubusercontent.com/aws/aws-sdk-ruby/master/gems/aws-partitions/partitions.json')
+    out_file = ManageIQ::Providers::Amazon::Engine.root.join('db/fixtures/aws_regions.yml')
+    service_name = 'ec2'
+
+    response = Net::HTTP.get_response(uri)
+    data = JSON.parse(response.body)
+
+    regions = data['partitions'].each_with_object({}) do |partition, memo|
+      dns_suffix = partition['dnsSuffix']
+      regions_info = partition['regions']
+
+      partition['services'][service_name]['endpoints'].each_key do |region_name|
+        raise "Repetitive region name: #{region_name}" if memo.key?(region_name)
+        memo[region_name] = {
+          :name        => region_name,
+          :hostname    => "#{service_name}.#{region_name}.#{dns_suffix}",
+          :description => regions_info.fetch(region_name).fetch('description'),
+        }.freeze
+      end
+    end
+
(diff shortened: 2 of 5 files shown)


AlexanderZagaynov:
@bronaghs @Ladas please review

spec/models/manageiq/providers/amazon/regions_spec.rb
```diff
@@ -1,18 +1,22 @@
...
+
   it "has all the regions" do
     ems = FactoryGirl.create(:ems_amazon_with_vcr_authentication)
 
     VCR.use_cassette(described_class.name.underscore) do
       current_regions = described_class.regions.map do |_name, config|
         {:region_name => config[:name], :endpoint => config[:hostname]}
       end
-      current_regions.reject! { |r| r[:region_name] == 'us-gov-west-1' }
+      current_regions.select! { |r| r[:region_name] =~ default_regions_regexp }
```
juliancheal: How does this and the regex above handle gov cloud? And potentially other regions like this PR https://github.com/aws/aws-sdk-ruby/blob/5fe5795e8910bb667996dfc75e4f16b7e69e3980/gems/aws-partitions/partitions.json#L11 aka China.

I'm not sure I'm understanding the regex.

juliancheal:
Over all looks really cool. 

1) When would we run the `regions.rake`? On builds of ManageIQ? I guess if a customer had an issue we could look at their regions file, to check it is in sync?

2) I assume customers adding their own regions/removing regions still works the same?

AlexanderZagaynov:
Answers to @juliancheal:
1. Regions list was hard-coded into the class before, and now it is stored in a separate `yaml` file. Both are checked in to the repo. Same way as before - if we need to do some changes - we just make a new PR. The only difference that previously we had to add those changes manually, and now we can simply run that task. Maybe it will be nice to allow customer to edit/update this file, but we need to think how it will be looks like, and I believe it's out of the scope of current PR (this one just an improvement).
2. Right, everything else works the same as before.

juliancheal:
@AlexanderZagaynov awesome, thanks!

juliancheal:
LGTM 👍

spec/models/manageiq/providers/amazon/regions_spec.rb
AlexanderZagaynov: I took that regexp from that file as it is. It's something from AWS internals.

You can find the China's regexp here: https://github.com/aws/aws-sdk-ruby/blob/5fe5795e8910bb667996dfc75e4f16b7e69e3980/gems/aws-partitions/partitions.json#L2058
And GovCloud's here: https://github.com/aws/aws-sdk-ruby/blob/5fe5795e8910bb667996dfc75e4f16b7e69e3980/gems/aws-partitions/partitions.json#L2359

This spec works the same way as before - it is hard to test those partitions (in the AWS gem's terminology) simply because our credentials designated for default partition only, and AWS's `DescribeRegions` call (no matter, will it be gem, some other library, or cli) returns only those, whose available for our account).

juliancheal:
@agrare what do you think?

app/models/manageiq/providers/amazon/regions.rb
```diff
@@ -1,92 +1,15 @@
...
-          :description => "South America (Sao Paulo)",
-        },
-        "us-gov-west-1"  => {
-          :name        => "us-gov-west-1",
-          :hostname    => "ec2.us-gov-west-1.amazonaws.com",
-          :description => "GovCloud (US)",
-        }
-      }
+      REGIONS = YAML.load_file(
+        ManageIQ::Providers::Amazon::Engine.root / 'lib' / 'data' / 'aws_regions.yaml'
```
Fryguy: Prefer the normal method approach as opposed to the `/` method for consistency...

```ruby
ManageIQ::Providers::Amazon::Engine.root.join("lib/data/aws_regions.yaml")
```

Fryguy: Why `lib/data`?  The lib directory is normally used for source.  Perhaps config directory is better as that's where we typically store yml files?  Also, prefer yml over yaml as the extension for consistency.

lib/tasks_private/regions.rake
```diff
@@ -0,0 +1,35 @@
...
+  desc 'Get / renew regions list from online source of AWS gem'
+  task :regions do
+    require 'uri'
+    require 'json'
+    require 'yaml'
+    require 'net/http'
+
+    # path valid for aws-sdk v3 gem's source
+    uri = URI('https://raw.githubusercontent.com/aws/aws-sdk-ruby/master/gems/aws-partitions/partitions.json')
+    out_file = ManageIQ::Providers::Amazon::Engine.root / 'lib' / 'data' / 'aws_regions.yaml'
```
Fryguy: Same on the `root.join`.  However, you may want to make the location a constant in the ManageIQ::Providers::Amazon::Engine class, so that you don't have the location duplicated in multiple places.

lib/tasks_private/regions.rake
```diff
@@ -0,0 +1,35 @@
...
+        raise "Repetitive region name: #{region_name}" if memo.key?(region_name)
+        memo[region_name] = {
+          :name        => region_name,
+          :hostname    => "#{service_name}.#{region_name}.#{dns_suffix}",
+          :description => regions_info.fetch(region_name).fetch('description'),
+        }.freeze
+      end
+    end
+
+    File.open(out_file, 'w') { |file| file << YAML.dump(regions) }
```
Fryguy: Prefer `File.write('out_file', YAML.dump(regions))`

lib/data/aws_regions.yaml
```diff
@@ -0,0 +1,73 @@
...
+  :hostname: ec2.cn-north-1.amazonaws.com.cn
+  :description: China (Beijing)
+cn-northwest-1:
+  :name: cn-northwest-1
+  :hostname: ec2.cn-northwest-1.amazonaws.com.cn
+  :description: China (Ningxia)
+us-gov-west-1:
+  :name: us-gov-west-1
+  :hostname: ec2.us-gov-west-1.amazonaws.com
+  :description: AWS GovCloud (US)
```
Fryguy: I recommend sorting this list by key before saving.  This way, future diffs are easier to read.

spec/models/manageiq/providers/amazon/regions_spec.rb
Fryguy: > and AWS's DescribeRegions call (no matter, will it be gem, some other library, or cli) returns only those, whose available for our account).

Doesn't this potentially cause us to miss regions if our account doesn't see them?

Fryguy: >  I took that regexp from that file as it is. It's something from AWS internals.

This regex ignores a bunch of regions, but the purpose of this test it to verify we have *all* regions, so I'm not sure the purposes match.

lib/tasks_private/regions.rake
```diff
@@ -0,0 +1,35 @@
...
+  task :regions do
+    require 'uri'
+    require 'json'
+    require 'yaml'
+    require 'net/http'
+
+    # path valid for aws-sdk v3 gem's source
+    uri = URI('https://raw.githubusercontent.com/aws/aws-sdk-ruby/master/gems/aws-partitions/partitions.json')
+    out_file = ManageIQ::Providers::Amazon::Engine.root / 'lib' / 'data' / 'aws_regions.yaml'
+    service_name = 'ec2'
```
Fryguy: ec2-only feels like it may be insufficient, because we also connect to s2, sqs, sns, etc, but it might be ok if ec2 is always a superset of the others.

spec/models/manageiq/providers/amazon/regions_spec.rb
AlexanderZagaynov: Hi @Fryguy, we are not able to get other regions from AWS, with any tool. It is limited by our AWS account. As you can see, previous test version was doing the same.

AlexanderZagaynov: P.S. I even don't see any advantage in this test. It just tests one recording against another.

lib/tasks_private/regions.rake
AlexanderZagaynov: I prefer to avoid full environment loading for a such simple task.

app/models/manageiq/providers/amazon/regions.rb
AlexanderZagaynov: Can you please point me to some similar file? I believe that **data** have nothing common with **configuration**. If you don't like `lib` directory, let's discuss what place will be good for a data files. It can be somewhat like `db/fixtures`, but current gem doesn't have it, should I create?

lib/tasks_private/regions.rake
AlexanderZagaynov: It was the same before, please take a look at the previous module code version. All those URLs are about `ec2`.

lib/tasks_private/regions.rake
Fryguy: I don't understand what that means considering you still need to access the same constant either way.

app/models/manageiq/providers/amazon/regions.rb
Fryguy: It's an interesting distinction. I guess `db/fixtures` works as a directory.  That would be consistent with the main repo at least.

spec/models/manageiq/providers/amazon/regions_spec.rb
Fryguy: @AlexanderZagaynov Yeah, I think I came to the same conclusion that the test is really a nonsense test.  I do wonder if we can create a test that shows we have *all* possible regions.

AlexanderZagaynov: @Fryguy since my script takes all possible regions from AWS gem's repo source, I see no point at all to have a such test (it will match anyway). If it will be some VCR recording, we will be simply comparing two recordings (one from VCR, another from that `yml` file). On each regions update both recordings should be updated.
Even if we will have proper AWS credentials to check China and Gov clouds - we will stay with the same VCR recordings.
So the only point to have that test - to check if `yml` file have proper format?
If your goal to check do we have _all_ possible regions, considering that we can have AWS's regions list change unnoticed - I would suggest to check that manually during QA stage. One should open Amazon's official regions list (from their docs site), and compare with what we have. Even if it will be automated - VCR records will make no sense, need to have a fresh info on each test run.

lib/tasks_private/regions.rake
AlexanderZagaynov: @Fryguy will it be better if I'll use something like `ManageIQ::Providers::Amazon::Engine::Paths::Fixtures::Regions`?
It will give you the same need to remember full path, this time - to constant.
So, what's the win? Code complication (need to keep that constant in some third file too, and remember about it's existence) and 15 chars economy?
**What that means** - my task loads in milliseconds. Task with the `task_name: :environment` will load a full app environment, which takes several seconds at least.

lib/data/aws_regions.yaml
AlexanderZagaynov: It possible, but it will make list inconsistent - China's and Gov Cloud's regions will be in the middle of the regular list. Now they're at the bottom, as defined in the AWS gem.
In case of any need to have specific sorting - I suggest to do that manually as needed.

AlexanderZagaynov:
Fixed on remarks:
- https://github.com/ManageIQ/manageiq-providers-amazon/pull/436#discussion_r184706154
- https://github.com/ManageIQ/manageiq-providers-amazon/pull/436#discussion_r184718526
- https://github.com/ManageIQ/manageiq-providers-amazon/pull/436#discussion_r184707411

spec/models/manageiq/providers/amazon/regions_spec.rb
Fryguy: @AlexanderZagaynov I agree.  I'll merge as is for now, and we can choose to drop this spec in a separate PR.

Fryguy:
Everything looks pretty good...I agree with all of your responses.

Fryguy:
https://bugzilla.redhat.com/show_bug.cgi?id=1500613

PR #438, opened 2018-04-26
Persister: InventoryCollection building through add_collection()

slemrmartin:
**Issue**: https://github.com/ManageIQ/manageiq/issues/17396
- [x] (**depends on**) PR: https://github.com/ManageIQ/manageiq/pull/17348/files

New Interface for definition/creating InventoryCollection from persister.
Using add_collection() instead of add_inventory_collection() to build collections.

Builder classes are defined on core project, provider specific InventoryCollections specified in concerns

Cc @Ladas, @agrare, @kbrock


app/models/manageiq/providers/amazon/inventory/persister/cloud_manager.rb
```diff
@@ -1,31 +1,40 @@
...
-
-    add_inventory_collection(
-      cloud.orchestration_stack_ancestry(
-        :dependency_attributes => {
-          :orchestration_stacks           => [collections[:orchestration_stacks]],
-          :orchestration_stacks_resources => [collections[:orchestration_stacks_resources]]
-        }
-      )
-    )
+    %i(vms
```
Ladas: @agrare @kbrock lets probably start the review from here. It's a half way between what the old compact way is, vs. the verbose way Marcel did. I think I like this. What do you think?

app/models/manageiq/providers/amazon/inventory/persister/shared/cloud_collections.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  included do
+    # Builder class for Cloud
+    def cloud
+      ::ManagerRefresh::InventoryCollection::Builder::CloudManager
+    end
+
+    def add_miq_templates(extra_properties = {})
+      add_collection(cloud, :miq_templates, extra_properties) do |builder|
+        builder.add_properties(:model_class => ::ManageIQ::Providers::Amazon::CloudManager::Template)
```
Ladas: @kbrock @agrare so instead of options that are nested hash, we have separate method for each (.add_properties, .add_builder_params, etc.). I think l like this aproach. It makes it cleaner and it allows us to redefine individual keys in nested hash, in specific providers. What do you think?

app/models/manageiq/providers/amazon/inventory/persister/cloud_manager.rb
Ladas: One difference with Marcel's way, this is not class definition. I wonder if we need it on class. While it looks a bit better, it requires a lot of lambdas to pass the dynamic params. What do you think?

app/models/manageiq/providers/amazon/inventory/persister/shared/cloud_collections.rb
```diff
@@ -0,0 +1,84 @@
+module ManageIQ::Providers::Amazon::Inventory::Persister::Shared::CloudCollections
+  extend ActiveSupport::Concern
+
+  included do
```
Ladas: Do we need this? It's all just method definitions, we shouldn't need this? Unless I'm missing something :-)

app/models/manageiq/providers/amazon/inventory/persister/shared/storage_collections.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def add_cloud_volumes(extra_properties = {})
+      add_collection(storage, :cloud_volumes, extra_properties) do |builder|
+        builder.add_properties(
+          :model_class => ::ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume
+        )
+
+        builder.add_builder_params(
+          :ems_id => ->(persister) { persister.manager.try(:ebs_storage_manager).try(:id) || persister.manager.id }
```
Ladas: this is all in persister context now? So we should be able to just do:
```
builder.add_builder_params(
  :ems_id => manager.try(:ebs_storage_manager).try(:id) || persister.manager.id
)
```

probably we could extract the managers bit into method
```ruby
def block_storage_manager
  persister.manager.try(:ebs_storage_manager).try(:id) || persister.manager.id
end

def object_storage_manager
  persister.manager.try(:s3_storage_manager).try(:id) || persister.manager.id
end
```

app/models/manageiq/providers/amazon/inventory/persister/shared/cloud_collections.rb
slemrmartin: more exactly ```add_properties()``` is 'top level' general method.
```add_builder_params()``` and ```add_dependency_attributes()``` are methods for nested hashes (with purpose @ladas wrote above). 
And ```add_targeted_arel()``` is method_missing's substitution for add_properties(:targeted_arel => ...)

File: app/models/manageiq/providers/amazon/inventory/persister/target_collection.rb [hunk 5]
```diff
@@ -1,145 +1,148 @@
...
+    add_orchestration_stack_ancestry
   end
 
-  def add_inventory_collection_with_references(inventory_collections_data, name, manager_refs, options = {})
-    options = shared_options.merge(inventory_collections_data.send(
-      name,
-      :manager_uuids => manager_refs,
-    ).merge(options))
+  # Top level models with direct references for Cloud
+  def init_cloud_ics_top_level_models
```
Ladas: Hm, instead of top/child level separation, I would rather like to see these shared with ManageIQ::Providers::Amazon::Inventory::Persister::CloudManager and ManageIQ::Providers::Amazon::Inventory::Persister::NetworkManager (and storage)

The only difference is that the TargetCollection is passing references(XY) to few collections. And I think we can even pass the references everytime. Because they are not used if we pass ```:targeted => false```, which is extracetd to method above and should be passed to every IC

app/models/manageiq/providers/amazon/inventory/persister/shared/cloud_collections.rb
slemrmartin: We have to replace Amazon's InventoryCollectionDefault classes with something else.
1) initialize_inventory_collection() method is more readable - this mixin is the reason why you wrote you like this approach / it simply defines "add_orchestration_stacks") method etc.
2) Amazon's ```Persister::CloudManager``` and ```Persister::TargetCollection``` (partially with NetworkManager) - and not only amazon's persisters - shares provider specific collections with slightly different settings. 
---
So without this mixin, you have to 1) repeat yourself in code and 2) have code uglier

app/models/manageiq/providers/amazon/inventory/persister.rb
```diff
@@ -51,11 +39,9 @@ def strategy
   end
 
   def shared_options
```
Ladas: hm, do we still use these somewhere?

app/models/manageiq/providers/amazon/inventory/persister/shared/storage_collections.rb
slemrmartin: Yes, it's possible, but I wanted to keep the same approach in core classes and the rest. 
New developer would be confused, what (s)he have to use on which place, this way there is no source of confusing and result is the same.

app/models/manageiq/providers/amazon/inventory/persister/shared/cloud_collections.rb
slemrmartin: Or if you're talking about ```ActiveSupport::Concern``` - why not? :)

Ladas: I mean, do we need the ```included do``` ? :-)

app/models/manageiq/providers/amazon/inventory/persister/shared/storage_collections.rb
Ladas: Right, so probably lets just drop the lambda? That should work consistent across ```core classes and the rest``` right?

app/models/manageiq/providers/amazon/inventory/persister/target_collection.rb
Ladas: right, so you can still pass the references only optionally. The point is sharing the definitions, which will help the readibility and will make it more composable.

app/models/manageiq/providers/amazon/inventory/persister/shared/cloud_collections.rb
Ladas: Yeah, I think we can remove it from most (all) of the places. It will shift it to left and make it more readable. :-)

app/models/manageiq/providers/amazon/inventory/persister/shared/storage_collections.rb
slemrmartin: I thought about keeping lambda...but you're right, it's not as difficult and nobody should be confused

app/models/manageiq/providers/amazon/inventory/persister.rb
slemrmartin: Do you mean advanced settings? I don't see all contexts yet, but because it has lowest importance from all definitions, you practically cannot add anything which can be useful because it doesn't have ability to overwrite anything - but this decision is up to you guys, I'm new there :)

Ladas: right, so this one was used to turn targeted refresh on (passing 1 setting to every IC) and to pass settings from advanced settings

slemrmartin: it is passing 2 settings (some providers 3) to every IC - ```strategy``` and ```targeted``` (at low level priority - when ic defines its own, it overwrites this one)

and settings from advanced settings

slemrmartin:
@miq-bot add_label refactoring

agrare:
Core PR was merged, kicking tests.

slemrmartin:
I think these test errors are not related

agrare:
@slemrmartin yeah agreed I pinged the amazon gitter room to get these fixed up.

agrare:
Okay https://github.com/ManageIQ/manageiq-providers-amazon/pull/452 should fix the spec failures

app/models/manageiq/providers/amazon/inventory/persister/definitions/cloud_collections.rb
```diff
@@ -0,0 +1,122 @@
...
+  end
+
+  def add_key_pairs(extra_properties = {})
+    add_collection(cloud, :key_pairs, extra_properties) do |builder|
+      builder.add_properties(:model_class => ::ManageIQ::Providers::Amazon::CloudManager::AuthKeyPair)
+      builder.add_properties(:manager_uuids => name_references(:key_pairs)) if targeted?
+    end
+  end
+
+  # TODO: mslemr - parent model_class used anywhere? If not, should be deleted
```
Ladas: not sure what is this todo about?

slemrmartin: There (in core):
https://github.com/ManageIQ/manageiq/pull/17348/files#diff-c4efbe5c5fc56829014b92fd356b974bR35
is explicitly defined model_class which cannot be derived automatically. 
But I'm not sure if it is necessary. If not, it can be removed from both definitions

Ladas: right so just having ::OrchestrationStack would fix it? I guess we can do that.

slemrmartin: Deleting it from core would fix it so auto-deriving is not replaced by manual definition.
But core's PR is merged now. It will be moved in another PR like your comments above

PR #439, opened 2018-05-03
handle volume snapshot status changes

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1561937
Related: https://github.com/ManageIQ/manageiq-content/pull/285

Note:
AWS CloudWatch should send specific events. 
To setup, one should create CloudWatch rule and target it to the `AWSConfig_topic` SNS topic.
This rule's event pattern should be:
```json
{
  "source": [
    "aws.ec2"
  ],
  "detail-type": [
    "EBS Snapshot Notification"
  ]
}
```
More info: https://aws.amazon.com/ru/blogs/aws/new-cloudwatch-events-for-ebs-snapshots

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
@@ -23,6 +23,13 @@ module ManageIQ::Providers::Amazon::CloudManager::EventParser
     event_hash[:availability_zone_ems_ref] = nil # Can't get it, needs to go through VM
   end
 
+  def self.parse_cloud_watch_ec2_ebs_snapshot_event!(event, event_hash)
+    event_hash[:message]                   = event['detail']
+    event_hash[:timestamp]                 = event['time']
+    event_hash[:vm_ems_ref]                = nil
+    event_hash[:availability_zone_ems_ref] = nil
+  end
+
   def self.parse_cloud_watch_alarm_event!(event, event_hash)
     event_hash[:message]                   = event["AlarmName"]
     event_hash[:timestamp]                 = event["StateChangeTime"]
@@ -38,11 +45,12 @@ module ManageIQ::Providers::Amazon::CloudManager::EventParser
       :ems_id     => ems_id
     }
 
-    unless %i(config cloud_watch_api cloud_watch_ec2 cloud_watch_alarm).include?(event["event_source"])
+    parse_method_name = "parse_#{event["event_source"]}_event!"
+    if singleton_class.method_defined?(parse_method_name)
+      send(parse_method_name, event, event_hash)
+    else
       raise "Unsupported event source #{event["event_source"]}"
     end
-
-    send("parse_#{event["event_source"]}_event!", event, event_hash)
 
     log_header = "ems_id: [#{ems_id}] " unless ems_id.nil?
     _log.debug("#{log_header}event: [#{event[:message]}]")
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
@@ -61,6 +61,14 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolumeSnapshot < ::
 
   def raw_delete_snapshot
     with_provider_object(&:delete)
+    update!(:status => 'deleting')
+    EmsRefresh.queue_refresh(
+      ManagerRefresh::Target.new(
+        :association => :cloud_volume_snapshots,
+        :manager_ref => { :ems_ref => ems_ref },
+        :manager_id  => ems_id,
+      )
+    )
   rescue => e
     _log.error "volume=[#{name}], error: #{e}"
     raise MiqException::MiqVolumeSnapshotDeleteError, e.to_s, e.backtrace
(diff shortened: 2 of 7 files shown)


app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -38,12 +45,13 @@ def self.event_to_hash(event, ems_id)
       :ems_id     => ems_id
     }
 
-    unless %i(config cloud_watch_api cloud_watch_ec2 cloud_watch_alarm).include?(event["event_source"])
+    parse_method_name = "parse_#{event["event_source"]}_event!"
+    if singleton_class.method_defined?(parse_method_name)
+      send(parse_method_name, event, event_hash)
```
Ladas: ```public_send``` would be better

Ladas:
@AlexanderZagaynov right, you have an event parser, that will store the event into our DB and Automate handler that will invoke target parsing and refresh.

So now we need the target parsing code:
https://github.com/Ladas/manageiq-providers-amazon/blob/4c66893fd1a91dc5fd3aa259a7c0ddd044c65576/app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb#L26

With specs, verifying we can parse targets out of the event like:
https://github.com/Ladas/manageiq-providers-amazon/blob/a9a281f710075e348354e853614b30d8faed0460/spec/models/manageiq/providers/amazon/cloud_manager/event_target_parser_spec.rb#L176

Looks like we already parse snapshot as part of volumes?:
https://github.com/Ladas/manageiq-providers-amazon/blob/4c66893fd1a91dc5fd3aa259a7c0ddd044c65576/app/models/manageiq/providers/amazon/cloud_manager/event_target_parser.rb#L105
----

Then it looks like we already have targeted collector and perister for snapshot:
https://github.com/Ladas/manageiq-providers-amazon/blob/52636410835d6319627afd5e489cefe199e81a01/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb#L139

https://github.com/Ladas/manageiq-providers-amazon/blob/2b85fe1aa4bd65c4ff8933157cadce7740d4de50/app/models/manageiq/providers/amazon/inventory/persister/target_collection.rb#L62

And there is already a spec, we could just add a test for isolated snapshot:
https://github.com/Ladas/manageiq-providers-amazon/blob/240c429c52f3cda61067aa560d3efe157451229b/spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/targeted_refresh_spec.rb#L332

AlexanderZagaynov:
@Ladas please review

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -17,7 +17,7 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
   require_nested :Template
   require_nested :Vm
 
-  OrchestrationTemplate.register_eligible_manager(self)
+  self::OrchestrationTemplate.register_eligible_manager(self)
```
Ladas: why is this change needed?

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
```diff
@@ -61,6 +61,14 @@ def self.create_snapshot(cloud_volume, options = {})
 
   def raw_delete_snapshot
     with_provider_object(&:delete)
+    update!(:status => 'deleting')
+    EmsRefresh.queue_refresh_task(
```
Ladas: so, there is no event for deletion?

AlexanderZagaynov: yep

app/models/manageiq/providers/amazon/cloud_manager.rb
AlexanderZagaynov: There was an issue, probably related to the rails dev env lazy code loading.
Rails was looking at the top level constant first, and there is no such method.
However, I forgot how to reproduce it now, so, reverting.

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume_snapshot.rb
Ladas: @agrare is this ok? If the event is missing, I think we do not have much choice. (except full refresh)

agrare: Can we do a targeted refresh of the vm the cloud_volume is connected to?  Not sure what amazon supports here for targets.
Also are you using the task that is returned?  Can you just use `EmsRefresh.queue_refresh`?

Ladas: it's a volume snapshot, so Vm is only optional (volume doesn't have to be attached to anything)

agrare: Okay sounds like a full is the only way to go then.

Ladas: And why not to just refresh the snapshot like it is now?

agrare: Oh we can target the snapshot?  If we can do a targeted definitely lets do that.

AlexanderZagaynov: @agrare fixing `queue_refresh`

AlexanderZagaynov: done

PR #419, opened 2018-03-13
Renamed method for consistency reasons

romanblanco:
**The PR should be merged together with the PR to the core repo: https://github.com/ManageIQ/manageiq/pull/17143**

---

Related to the changes in:

https://github.com/ManageIQ/manageiq/pull/17143
https://github.com/ManageIQ/manageiq-ui-classic/pull/3607

cc @miha-plesko

PR #441, opened 2018-05-09
Renamed method for consistency reasons

romanblanco:
The other half of changes for https://github.com/ManageIQ/manageiq-providers-amazon/pull/419


h-kataria:
@bronaghs looks good, please merge

PR #444, opened 2018-05-16
Add endpoint support to raw_connect

djberg96:
This adds a uri endpoint argument to the `raw_connect` method. It is nil by default, but if present it will be passed to the AWS connection code.

Note that this PR depends on a change on the classic UI side. The UI change will add an endpoint text field, which is handled as a URI object, and its parts saved to the `endpoints` table.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1339398

See also: https://github.com/ManageIQ/manageiq-ui-classic/issues/3907

### DIFF
--- a/app/models/manageiq/providers/amazon/manager_mixin.rb
+++ b/app/models/manageiq/providers/amazon/manager_mixin.rb
@@ -48,11 +48,11 @@ module ManageIQ::Providers::Amazon::ManagerMixin
     # Connections
     #
 
-    def raw_connect(access_key_id, secret_access_key, service, region, proxy_uri = nil, validate = false)
+    def raw_connect(access_key_id, secret_access_key, service, region, proxy_uri = nil, validate = false, uri = nil)
       require 'aws-sdk'
       require 'patches/aws-sdk-core/seahorse_client_net_http_pool_patch'
 
-      connection = Aws.const_get(service)::Resource.new(
+      options = {
         :access_key_id     => access_key_id,
         :secret_access_key => MiqPassword.try_decrypt(secret_access_key),
         :region            => region,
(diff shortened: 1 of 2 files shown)


djberg96:
@Ladas @AlexanderZagaynov Please review

AlexanderZagaynov:
LGTM :+1:

AlexanderZagaynov:
However (I mentioned this here: https://github.com/ManageIQ/manageiq-ui-classic/pull/3958#discussion_r190510607)
user input should be escaped for safety.
I'm not sure, which place can be an entry point for such escaping, typically, if it some model, that should be made in some `before` callback.

djberg96:
@AlexanderZagaynov Thanks. I will leave that to the UI team. :)

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -48,19 +48,23 @@ module ClassMethods
...
         :region            => region,
         :http_proxy        => proxy_uri,
         :logger            => $aws_log,
         :log_level         => :debug,
         :log_formatter     => Aws::Log::Formatter.new(Aws::Log::Formatter.default.pattern.chomp)
-      )
+      }
+
+      options[:endpoint] = uri.to_s if uri && uri.to_s.present?
+
```
bronaghs: @djberg96 -`present` tests for nil too so there is no need to explicitly check it.

spec/models/manageiq/providers/amazon/cloud_manager_spec.rb
```diff
@@ -14,6 +14,12 @@
       described_class.raw_connect('access_key', 'secret_access_key', :EC2, 'region', 'uri', true)
     end
 
+    it "validates credentials with an optional uri endpoint" do
+      expect(described_class).to receive(:validate_connection)
+      uri = URI.parse('https://apigateway.us-east-1.amazonaws.com')
+      described_class.raw_connect('access_key', 'secret_access_key', :EC2, 'region', 'uri', true, uri)
```
bronaghs: the `uri` parameter is listed twice, can you make the first one `proxy_uri` otherwise it is confusing. Or make the second one `endpoint_uri` if thats easier.

djberg96:
@bronaghs ok, updated.

bronaghs:
Thanks, I will merge when green.

PR #447, opened 2018-05-18
load regions list from aws-sdk gem's source & disable special ones

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1500613
This is a cherry-picks from:
https://github.com/ManageIQ/manageiq-providers-amazon/pull/265, https://github.com/ManageIQ/manageiq-providers-amazon/pull/436, https://github.com/ManageIQ/manageiq-providers-amazon/pull/443, https://github.com/ManageIQ/manageiq-providers-amazon/pull/446


simaishi:
@AlexanderZagaynov @bronaghs A few questions...

1) The BZ is marked to be fixed for 'master' branch only. If we need/want to fix in Fine branch, we must also fix in gaprindashvili branch. Please make sure `gaprindashvili/yes` labels exist on all needed PRs.

2) Are those cherry-picks not clean?  We always prefer a cherry-pick over a separate PR for G/F branches, if possible.

AlexanderZagaynov:
@simaishi `Version` field's value in that BZ is `5.8.0`.
I understood that as I should prepare a PR for **Fine**. 
Even more you mentioned PR necessity here: https://github.com/ManageIQ/manageiq-providers-amazon/pull/443#discussion_r188750825
Those commits can't be simply cherry-picked, I had to solve a couple of merge conflicts.
For **Gaprindashvili** I guess those commits can be cherry-picked directly.

simaishi:
Thanks for clarifying about cherry-pick.

"version" indicates where the issue was found, and doesn't always match which version we want to fix in...  The version we want to fix is "Target Release", and cfme-future is 'master' branch. That's why I asked... @bronaghs please confirm we want to fix this in G/F branch as well.

PR #448, opened 2018-05-22
Explicitly disable ems_network_new creation

djberg96:
Currently the Amazon provider does not explicitly disable EMS network creation and editing. Based on the current UI code, as well the Nuage provider code, support for `:ems_network_new` should be disabled.

Note that this will also require at least one code change on the classic UI side. Please see https://github.com/ManageIQ/manageiq-ui-classic/pull/3976, which properly uses the `SupportsFeatureMixin` hook.

While I was here I also added the beginnings of a test suite for the Amazon NetworkManager.

Partial fix for https://bugzilla.redhat.com/show_bug.cgi?id=1577888


djberg96:
@Ladas @tumido Please review

PR #449, opened 2018-05-22
AWS instance types data refactoring and update

### LINKED ISSUE #396: Gather AWS instance types list from AWS's GitHub repo
Instance types list currently hard-coded in the https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/instance_types.rb#L7 file.

We can fetch it (either dynamically or at build time) from: https://github.com/aws/aws-sdk-ruby/blob/master/apis/ec2/2016-11-15/api-2.json#L11007 

_(link to the latest version's date available, all the version's dates can be seen here: https://github.com/aws/aws-sdk-ruby/blob/master/apis/ec2)_

### LINKED ISSUE #397: Gather AWS instances details from AWS's Price List API
Instances details are currently hard-coded in the https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/instance_types.rb#L12 file.

We can fetch it (either dynamically or at build time) following the Amazon's Price List API instruction: 
https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/using-ppslong.html

P.S. It can be useful also to use price list changes notifications: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/price-notification.html

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1562062
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1590288

Changes:
- Instance types getting task was removed together with intermediate data files (see below about further plans).
- Data moved to and loaded from yaml file.
- Specs updated to include new types amount.

Further plans:
- Second PR will update `aws-sdk` gem to `v3`.
- Third one should remove current yaml, and update data as a part of standard EMS refresh process.

Old vs New data: https://goo.gl/p2qFKq


lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,61 @@
...
+    out_file_old = data_dir.join('aws_instance_types_old.yml')
+
+    # 'Dedicated Host', # TODO: do we need bare metal types here? e.g. "m5", "p3", etc.
+    products = AwsProductsDataCollector.new(
+      'AmazonEC2', 'Compute Instance', AwsInstanceTypesParser::REQUIRED_ATTRIBUTES,
+      'instanceType', 'currentGeneration').products_list
+
+    instances = AwsInstanceTypesParser.new(products).instances_list
+
+    # prevent yaml anchors/aliases
```
Fryguy: This should not be necessary. To avoid aliases, there are multiple approaches.  If you are creating the data structures in question, then you can just avoid putting references into other container instances (e.g. don't put the same array into multiple hashes).  Instead, use `.dup` or `.clone`.  If you aren't creating the hashes, and instead they come from Amazon, you can use `.deep_clone` or `Marshal.load(Marshal.dump(obj)` which will ultimately serialize and then deserialize, which should give you what you need

lib/tasks_private/lib/simple_logging.rb
```diff
@@ -0,0 +1,12 @@
...
+require 'logger'
+require 'active_support/core_ext/module/delegation'
+
+module Kernel
+  private
+  def logger
+    @logger ||= Logger.new(STDOUT)
+  end
+  delegate *%i(info debug warn error fatal), :to => :logger
+end
```
Fryguy: Why do you need this?  if all you are trying to do is call a logger, just set a global `$log = Logger.new(STDOUT)`, and be done with it.  I'd prefer that over patching Kernel.

lib/tasks_private/lib/github_file.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+require 'uri'
+require 'json'
+require 'base64'
+require 'pathname'
+require 'net/http'
+require_relative 'simple_logging'
+
+class GithubFile
```
Fryguy: I think it would be easier to just use Octokit than to build your own.  Though I don't understand the purpose of the cache_dir, nor the commits_uri

AlexanderZagaynov:
@Fryguy @Ladas please review

AlexanderZagaynov:
@Fryguy @Ladas @bronaghs  please review

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,277 @@
...
+      GH_CACHE.fetch("#{file_path}.#{sha}".tr('/', '.')) do
+        path = "/#{SDK_REPO_PATH}/#{sha}/#{file_path}"
+        http_get(URI::HTTPS.build(:host => GH_RAW_HOST, :path => path))
+      end
+    end
+
+    discontinued_types, types_list = isolated do
+      previous_list = []
+      versions = get_gh_data("contents/#{SDK_DATA_DIR}")
+      versions = versions.lazy.select { |v| v['type'] == 'dir' }.map { |v| v['name'] }.sort
```
Fryguy: If you're ultimately iterating every object, then the lazy is not necessary.

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,277 @@
...
+      all_types, latest = versions.reduce([[], []]) do |(all, _latest), version|
+        data = get_gh_file("#{SDK_DATA_DIR}/#{version}/#{SDK_FILE_NAME}")
+        data = data['shapes']['InstanceType']['enum']
+        old_minus_new = previous_list - data
+        new_minus_old = data - previous_list
```
Fryguy: removed/added might be better variable names for readability.

lib/tasks_private/instance_types.rake
AlexanderZagaynov: There is `select` first, then `map`, so I'm saving one loop.

lib/tasks_private/instance_types.rake
AlexanderZagaynov: is that critical?

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,277 @@
...
+      instance_data.except!(:default, :deprecated, :discontinued, :disabled)
+      instance_data[:default] = true if instance_type == default_type
+      if discontinued_types.include?(instance_type) || !instance_data[:current_version]
+        instance_data[:discontinued] = true
+        instance_data[:disabled] = true
+      end
+      if instance_data[:current_version] && !instance_data[:current_generation]
+        instance_data[:deprecated] = true
+      end
+      types_list.index(instance_type) || 1_000_000
```
Fryguy: How do unindexed types get sorted amongst themselves?  I would expect by instance_type name or something.

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,277 @@
...
+      report << 'th,td{border:1px dotted gray}'
+      report << '.added{background-color:rgb(223,240,216)}'
+      report << '.removed{background-color:rgb(242,222,222)}'
+      report << '.changed{background-color:rgb(252,248,227);}'
+      report << '</style>'
+      report << '</head>'
+      report << '<body>'
+      report << '<h1>Generated: <script>'
+      report << "document.write(new Date('#{report_date.iso8601}').toString())"
+      report << '</script></h1>'
```
Fryguy: I don't understand the point of this report

lib/tasks_private/instance_types.rake
AlexanderZagaynov: They should keep previous order. And it is impossible to sort by numbers together with the strings.

lib/tasks_private/instance_types.rake
AlexanderZagaynov: It shows what attributes was changed - for control. I've used it to control data consistency before my PR and after.
However, this report doesn't go anywhere besides local `tmp` dir.

Fryguy:
I think you can just require a GITHUB_API_TOKEN.  I don't see why that needs to be made optional.  This should simplify the code a bit.

Fryguy:
Unless I'm mistaken, the report in unnecessary.  The code can generate the yml file, and `git diff` is good enough at that point for a report.

lib/tasks_private/instance_types.rake
Fryguy: Ok, yeah, then IMO it's completely unnecessary...`git diff` should be more than sufficient for those purposes.

lib/tasks_private/instance_types.rake
Fryguy: True...it's still an over-optimization (i.e. this code probably goes from less than a second to also less than a second)

lib/tasks_private/instance_types.rake
Fryguy: Others would say no, but to me readability and maintenance are crticial. :smile:  old_minus_new says nothing about the intent, which causes more cognitive overhead, as compared to a name like `removed`

lib/tasks_private/instance_types.rake
Fryguy: It's possible, you just need a multi-level sort...

```ruby
[types_list.index(instance_type) || 1_000_000, instance_data[:instance_type]
```

If you return that from the sort_by! block it will do multi-level sort on previous index, then name.

Fryguy: That being said, it's much more intuitive to simplify this to just sort by name instead of previous index.  New values will get inserted where they belong and the diff will still only include the new stuff.

AlexanderZagaynov:
@Fryguy 
- `GITHUB_API_TOKEN` only needed in case of "Rate limit" error. It should not arise in the foreseen usage case. Getting such token requires some additional actions from user, which can complicate things a bit.
- `git diff` can be messed at least due to data movement from 'active' to 'discontinued'. Moreover, it doesn't affect usage experience in the worse way, and doesn't affect overall app performance (because it is rake task, which supposed to be run rarely). You can see example report screenshot in attachment.
screenshot from 2018-07-03 16-36-35

lib/tasks_private/instance_types.rake
AlexanderZagaynov: Are you insist to remove `.lazy` then? :)

lib/tasks_private/instance_types.rake
AlexanderZagaynov: 1) It complicates code and readability, and (maybe less important, but...) it will decrease speed and increase memory usage.
2) Types list order can't be sorted alphabetically, that's why I'm taking types list from AWS GitHub repo, to make it consistent with AWS lists.

lib/tasks_private/instance_types.rake
AlexanderZagaynov: `git diff` can't guarantee you consistent experience.
You can see it here for example: https://github.com/AlexanderZagaynov/manageiq-providers-amazon/compare/3a9a45211d9374470c5051b87e0b1187cdcc30ac...a92033eb4852eda82bb5a8263305e63067380a0e#diff-d8dde9392fe3a31bd3f5580c14571a1cL1681
(I'm attaching a screenshot of this)
screenshot from 2018-07-03 16-53-27

Fryguy:
> GITHUB_API_TOKEN only needed in case of Rate limit error. It typically should not arise in the foreseen usage case. Getting such token requires some additional actions from user, which can complicate things a bit.

The gem maintainers are the only users, and it's rather trivial to get a one-time token.  We enforce GITHUB_API_TOKEN in things like manageiq-release and miq-bot, so I don't really see your argument.

Fryguy:
> git diff can be messed at least due to data movement from 'active' to 'discontinued'. Moreover, it doesn't affect usage experience in the worse way, and doesn't affect overall app performance (because it is rake task, which supposed to be run rarely). You can see example report screenshot in attachment.

If you sort by name, and keep the sorting consistent, then there is no data movement.  You would literally see only 1 key, "discontinued" changing from false to true.

While I agree that having a pretty output is pretty, and doesn't impact overall performance, "no code is better than no code".  *Not* having to maintain something, both in time and human readability, is generally better, especially when something already exists to solve your problem (i.e. visualizing changes).  The fact that we are even discussing it (where we wouldn't be discussing it otherwise) is proof enough, not to mention the time it will take to review all of that code and go through all of those comments on it.

lib/tasks_private/instance_types.rake
Fryguy: The link doesn't work...do you have another link with an example?  Wanted to see if `git diff --patience` made a difference at all

AlexanderZagaynov:
@Fryguy I've done with your remarks.

One note: as I see you still don't understand (https://github.com/ManageIQ/manageiq-providers-amazon/pull/449#issuecomment-402229384 _"If you sort by name..."_) the point of sorting instance types not by name, but by AWS given list. Here is example, why I'm querying not only their API, but their repo too:
```ruby
%w[m3.medium m3.large m3.xlarge m3.2xlarge].sort
=> ["m3.2xlarge", "m3.large", "m3.medium", "m3.xlarge"]
```

Fryguy:
From a pure-yaml perspective, I think that

```
%w[m3.medium m3.large m3.xlarge m3.2xlarge].sort
=> ["m3.2xlarge", "m3.large", "m3.medium", "m3.xlarge"]
```

is fine.  The important thing is consistency, so that future runs sort the same way, even if you have to rebuild the file from scratch.

> Although I'm not sure, will it be used somewhere in UI or not.

It will, but by the time it does it will have gone through EMSRefresh -> stored in a database (thus losing the order) -> UI retrieves from the database and sorts again for presentation.

Fryguy:
> One note: as I see you still don't understand (#449 (comment) "If you sort by name...") the point of sorting instance types list not by name, but by AWS given list. Here is example, why I'm querying not only their Price List API, but their repo too:

I don't understand, and the example does not make it clear.  I'm not sure why 2 sources of information matters when there is only a single "key" in the yaml in the end.

lib/tasks_private/instance_types.rake
Fryguy: > It complicates code and readability, and (maybe less important, but...) it will decrease speed and increase memory usage.

That should not be a concern, unless you are saying it will take *hours* and use *gigs* of memory.  This is a one-off execution.

> Types list order can't be sorted alphabetically (it leads to bad end user experience), that's why I'm taking types list from AWS GitHub repo, to make it consistent with AWS lists.

I don't understand the bad end user experience.  The sorting in the file is strictly for diff purposes and the end user will never see that sorting.  Unless you mean the dev running this rake task as the end-user?

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,191 @@
...
+    out_file = data_dir.join('aws_instance_types.yml')
 
-  def format_instance_miq(i)
-    {}.tap do |o|
-      # x1.16large manually fixed to x1.16xlarge
-      # i3.16large manually fixed to i3.16xlarge
-      %i(family instance_store_size instance_store_volumes vpc_only enhanced_networking).each do |key|
-        o[key] = i[key]
+    # weird cache logging issue workaround
+    I18n.backend = I18n.backend.backend unless Rails.initialized?
```
Fryguy: Really curious about this...never seen this before...if you found it via the internet can you include a link to it in the comments for future readers?

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,191 @@
...
-  instances.each do |i|
-    i.deep_symbolize_keys!
-    miq_format = format_instance_miq(i)
-    if i[:generation] == 'current'
-      available[i[:instance_type]] = miq_format
-    else
-      previous[i[:instance_type]] = miq_format
+    # prevent global namespace and memory pollution
+    def isolated
+      yield
```
Fryguy: Seems this method is no longer necessary, unless you have some future plans for it?

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,191 @@
...
+    def http_get(uri, headers = {})
+      LOGGER.info("Making request to #{uri}")
 
-  # new discontinued from current_deprecated
-  discontinued_count += current_deprecated.count
-  current_deprecated.values.each do |instance|
-    instance[:disabled] = true
-    output_instances << instance
-  end
+      response = Net::HTTP.start(uri.host, :use_ssl => true) do |http|
```
Fryguy: I thought elsewhere I mentioned using Octokit as opposed to Net::Http (unless you are grabbing non github content?

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,191 @@
...
-    f.puts(new_instance_types_file)
-  end
+    def get_gh_file(file_path, sha: nil)
+      if sha.blank?
+        commits = get_gh_data('commits', :path => file_path, :page => 1, :per_page => 1)
+        sha = commits[0]['sha']
+        LOGGER.debug("Latest commit SHA of '#{file_path}' file is #{sha}")
+      end
+
+      GH_CACHE.fetch("#{file_path}.#{sha}".tr('/', '.')) do
```
Fryguy: Can you explain the reason for the cache?  I don't understand why it's necessary.

lib/tasks_private/lib/aws_instance_data_parser.rb
```diff
@@ -0,0 +1,336 @@
+# frozen_string_literal: true
+
+class AwsInstanceDataParser
```
Fryguy: This class feel overly complex with all of the `concerning`, `memoizable`, a special helper method for memoization, hardcoded constants, delayed parsing, etc.  Can you explain why this is all needed.  In the back of my head, it feels like an awful lot of code when the goal is to take one-or-more jsons and transform to a different json.  That feels like it would be at most it would be a 100 line method, so I feel like I'm missing something.

spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
```diff
@@ -147,7 +147,7 @@ def table_counts_from_api
       :custom_attribute              => custom_attributes_count,
       :disk                          => disks_count,
       :firewall_rule                 => firewall_rules_count,
-      :flavor                        => 103,
+      :flavor                        => 114,
```
Fryguy: How did we gain 11 flavors?  Was this just the count that was missing since the last update, or are those discontinued ones?

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,191 @@
...
-  current_available = ManageIQ::Providers::Amazon::InstanceTypes::AVAILABLE_TYPES.dup
-  output_instances = []
-  current_available.keys.each do |type|
-    available_instance = available[type]
-    next unless available_instance
-    output_instances << available_instance
-    current_available.delete(type)
-    available.delete(type)
-  end
+    # using python sdk, as it contains more historical data than ruby-sdk
```
Fryguy: Does this require something to be installed locally?

lib/tasks_private/instance_types.rake
Fryguy: No, but if other changes are being made, I would suggest remove it

Fryguy:
Here is a gist with a quick generator for the yaml based on the original hashes, along with a diff based on what your code generates (I had to tweak your sorting code a little because even though it looks like it preserves the original order, it doesn't).

https://gist.github.com/Fryguy/a12ab407e83a513d08d5ad6281e88dcf

I recommend that the "base" yaml extracted from the Ruby file is committed first, and then your changes are just a regular diff on top of that.

lib/tasks_private/instance_types.rake
AlexanderZagaynov: I don't want to introduce a new gem for a simple `get` purpose

lib/tasks_private/instance_types.rake
AlexanderZagaynov: It's primary usage is for `AwsProductsDataCollector`, because those files are too big, but permanent in time (new prices/positions - new price list).

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: For readability and maintainability.

spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
AlexanderZagaynov: For example: https://aws.amazon.com/blogs/aws/amazon-ec2-update-additional-instance-types-nitro-system-and-cpu-options/

lib/tasks_private/instance_types.rake
AlexanderZagaynov: No

lib/tasks_private/instance_types.rake
AlexanderZagaynov: No, I found the solution with debugging.
Without: `I18n::InvalidLocale: :en is not a valid locale`

AlexanderZagaynov:
@Fryguy about "diff with original data": here you are: https://github.com/AlexanderZagaynov/manageiq-providers-amazon/compare/3a9a45211d9374470c5051b87e0b1187cdcc30ac...GH-0396
However, you can see the same blended result, because originally (https://github.com/ManageIQ/manageiq-providers-amazon/pull/449/commits/83c26b32d2748ea1b0935103ec624c8d2c9d0f71#diff-6239e0700a502ffce9e787dd8034668aL8 - don't forget to press "load diff") it was also sorted in a meaningful way, not just with `sort`

lib/tasks_private/lib/aws_instance_data_parser.rb
Fryguy: By hardcoded I constants I meant things like https://github.com/ManageIQ/manageiq-providers-amazon/pull/449/files#diff-c1dcfc36a8dc66b1c22b3401ea962b35R43, where there are specific types hardcoded as opposed to discovered from the raw data.  However, it might be a necessary evil if the raw data sources don't have that association.

spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
Fryguy: So you are saying those are new instances since our last update of the yaml file?  I'm more asking because when I diffed the lists before your PR and after, I didn't see 11 new flavors in the yml file.

Fryguy:
I see...so, the trouble with AlexanderZagaynov/manageiq-providers-amazon@3a9a452...GH-0396 is that the diff also contains different sorting, so it's really difficult to interpret.  Instead, perhaps, sort the file first, commit that, then commit your changes on top of that, so the diff is as minimal as possible.  That's essentially what I have in https://gist.github.com/Fryguy/a12ab407e83a513d08d5ad6281e88dcf

lib/tasks_private/instance_types.rake
Fryguy: @AlexanderZagaynov If you're just doing a simple get, then I recommend using `open-uri`.  It's part of stdlib, and greatly simplifies the code.  FWIW, I don't understand the aversion to adding a new gem, when the purpose of the gem makes the code clearer, even if it achieves ultimately the same thing.

lib/tasks_private/instance_types.rake
Fryguy: @AlexanderZagaynov Yeah, after I made this comment, I ran it locally and saw the size of those files...they are enormous.

lib/tasks_private/instance_types.rake
AlexanderZagaynov: Since GitHub now isn't involved at all, we don't need that gem.
I believe `open-uri` can be a bad idea: https://sakurity.com/blog/2015/02/28/openuri.html
And I'm always avoiding such things just as pattern / good habit.

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: Yes, some data is missing there. 
BTW, maybe we don't need some fields at all?
I have no idea where they used, maybe you can point me on?

AlexanderZagaynov:
@Fryguy Here is the old-new diff: https://github.com/AlexanderZagaynov/manageiq-providers-amazon/compare/beeb0bf...GH-0396#diff-d8dde9392fe3a31bd3f5580c14571a1c

AlexanderZagaynov:
New flavors:
- c5d.18xlarge
- c5d.2xlarge
- c5d.4xlarge
- c5d.9xlarge
- c5d.large
- c5d.xlarge
- m5d.12xlarge
- m5d.24xlarge
- m5d.2xlarge
- m5d.4xlarge
- m5d.large
- m5d.xlarge

Total flavors count: 119, was 107 (+12)
Including deprecated: 33, was 33 (+0)
Including discontinued: 5, was 4 (+1)

New discontinued is `i3.metal`. That's because it's a bare type, and I'm filtering them out.

spec/models/manageiq/providers/amazon/aws_refresher_spec_counts.rb
AlexanderZagaynov: See https://github.com/ManageIQ/manageiq-providers-amazon/pull/449#issuecomment-404568411

lib/tasks_private/instance_types.rake
```diff
@@ -1,215 +1,94 @@
...
-  def self.names
-    instance_types.keys
-  end
-end
-EOT
+    ## save data
 
-  File.open(ManageIQ::Providers::Amazon::Engine.root.join('instance_types.rb'), 'w') do |f|
-    f.puts(new_instance_types_file)
+    out_file.write(types_data.to_yaml.each_line.map(&:rstrip).join("\n") << "\n")
```
Fryguy: I would avoid the line stripping...let `to_yaml` do what it does and just write that.  As long as there is consistency, autogenerated files, IMO, should be as close to raw as possible.

So this line simplifies to `outfile.write(types_data.to_yaml)`

lib/tasks_private/lib/aws_instance_data_parser.rb
```diff
@@ -0,0 +1,316 @@
...
+  def physical_processor
+    product_data['physicalProcessor']
+  end
+
+  def cpu_clock_speed
+    product_data['clockSpeed']
+  end
+
+  def cpu_arches
+    CPU_ARCHES.fetch(product_data['processorArchitecture']) do |cpu_arch|
```
Fryguy: I found a simpler way to do this....

```ruby
CPU_ARCHES => {32 => "i386", 64 => "x86_64"}.freeze

product_data['processorArchitecture'].scan(/\d+/).map { |bits| CPU_ARCHES[bits] }
```

lib/tasks_private/lib/aws_instance_data_parser.rb
```diff
@@ -0,0 +1,316 @@
...
+      :vpc_only                => vpc_only?,
+    }.freeze
+    @unknown_values.freeze
+    @parsed = true
+  end
+
+  def save_unknown(value, attribute_name = nil, nils: nil)
+    attribute_name ||= caller_locations(1..1).first.label # use a bit of magic
+    (@unknown_values[attribute_name.to_sym] ||= Set.new) << value
+    nils ? Array.new(nils, nil) : nil
```
Fryguy: I don't really understand the purpose of this method.  I think its purpose is to deal with post-parse-presentation of unexpected things, and then returning an Array of 3 nils (not sure shy that's needed).

lib/tasks_private/lib/aws_instance_data_parser.rb
Fryguy: It's quite possible we don't need a lot of the fields...you'd have to search the parser for what is used out of the instance types data.  When I first built the file, we wanted everything, and since I had to manually scrape anyway, it was in the same place, so I just added it in, because we didn't know what would be used.  Ultimately we used *some* of them, but there are things we definitely don't use.  I would be ok with dropping the unused things.

Fryguy:
> @Fryguy Here is the old-new diff: AlexanderZagaynov/manageiq-providers-amazon@beeb0bf...GH-0396diff-d8dde9392fe3a31bd3f5580c14571a1c

Link doesn't work for me...not sure why.

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: `save_unknown` used in after `||` part of assignments. e.g. `something = we_have || we_dont_have_and_we_need_to_remember_that`, so I return `nil` to set the variable(s)

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: This way we will not be able to control if that field contains something unexpected.

lib/tasks_private/instance_types.rake
AlexanderZagaynov: rubocop complains on right spaces

AlexanderZagaynov:
@Fryguy shortened link to `diff`: https://goo.gl/B2FxJQ

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: In that case I propose to do that in a separate PR, as it can take much time, and those 2 BZs will still stay unclosed.

lib/tasks_private/instance_types.rake
Fryguy: autogenerated files should be ignored by rubocop...we just put it in the exclusion list

lib/tasks_private/lib/aws_instance_data_parser.rb
Fryguy: So you prefer the hardcoded-ness of it?

lib/tasks_private/lib/aws_instance_data_parser.rb
Fryguy: Oh I see, so it's for learning about new things.  Hmmm... that's interesting.

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: Yes, this way we will not be blind when Amazon'll introduce something new here.

Fryguy:
Some comments on that diff (I can't comment directly, so inline)

- Prefer puttin hvm before paravirtual for backward consistency (and it's alphabetical)

```diff
-  - :hvm
   - :paravirtual
+  - :hvm
```

- memory is losing precision...is that intentional?

```diff
-  :memory: 1825361100.8
+  :memory: 1825361100
```

- description and family changed, which is end-user apparent, and is possibly an issue

```diff
-  :description: C1 High-CPU Medium
+  :description: C1 Compute optimized Medium
...
-  :description: C1 High-CPU Extra Large
+  :description: C1 Compute optimized XL
...
-  :description: C3 High-CPU Double Extra Large
+  :description: C3 Compute optimized 2XL
...
-  :family: Compute Optimized
+  :family: Compute optimized
```

- network type changed in a handful of cases:

```diff
-  :network_performance: :up_to_10_gbps
+  :network_performance: :up_to_10_gigabit
```

lib/tasks_private/lib/aws_instance_data_parser.rb
```diff
@@ -0,0 +1,316 @@
...
+          save_unknown(instance_type, :instance_type, :nils => 3)
+        )
+      ).freeze
+  end
+
+  def parsed_storage
+    @parsed_storage ||= begin
+      volumes, size, type =
+        if ebs_only?
+          Array.new(3, nil)
```
Fryguy: This can be simplified without resorting to creating temporary arrays...  if you have extra vars on the left hand side of an assignment, Ruby will just set the rest to nil, so you can just put `nil` here, and it's enough

```ruby
[2] pry(main)> a = b = c = 1
=> 1
[3] pry(main)> a, b, c = nil
=> nil
[4] pry(main)> [a, b, c]
=> [nil, nil, nil]
```

lib/tasks_private/lib/aws_instance_data_parser.rb
```diff
@@ -0,0 +1,316 @@
...
+
+  def intel_avx?
+    product_data['intelAvxAvailable'] == 'Yes' ||
+      !!(processor_features =~ INTEL_AVX_REGEXP) ||
+      CPU_FEATURES[physical_processor]&.include?(:avx)
+  end
+
+  def intel_avx2?
+    product_data['intelAvx2Available'] == 'Yes' ||
+      !!(processor_features =~ INTEL_AVX2_REGEXP) ||
```
Fryguy: Instead of `!!(foo =~ REGEXP)`, you can also do `REGEXP.match?(foo)`, which reads a little nicer.

Ladas:
@AlexanderZagaynov @Fryguy can we fetch the flavors from the API as @djberg96 writes in https://github.com/ManageIQ/manageiq-providers-amazon/issues/397#issuecomment-406022861 ?

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: Yes, I told you so there: https://github.com/ManageIQ/manageiq-providers-amazon/pull/449#discussion_r202402641

However, regardless the current context, I would like to mention that implicit initialization can be bad sometimes:
```ruby
a, b, c = 1
puts [a, b, c].map(&:inspect)

=>

1
nil
nil
```

lib/tasks_private/lib/aws_instance_data_parser.rb
AlexanderZagaynov: `Regexp` doesn't have such method:
```
//.methods.grep /match/
=> [:match]
```

AlexanderZagaynov:
@Fryguy 

> memory is losing precision...is that intentional?

yes. `1000.5` bytes doesn't makes sense, right?

> description and family changed

`description` was generated by 3rd party: https://github.com/powdahound/ec2instances.info/blob/master/scrape.py#L665
I'm not even sure, it is used / or it is some key field.

`family` - I'm taking it directly from AWS price list data, as you can see - the only difference - is a second word capitalization, which I believe should also cause no issues.

@martinpovolny can you say how this instance types list used in UI? Which fields are mandatory, and how/where they're used?

@Ladas 

> can we fetch the flavors ...

https://github.com/ManageIQ/manageiq-providers-amazon/issues/397#issuecomment-406250918

bronaghs:
@AlexanderZagaynov - they are displayed as Flavors in the UI. "Instance Types" is the AWS terminology, "Flavors" is the ManageIQ terminology.
A drop down of flavors will be displayed in the instance provisioning dialogs.

AlexanderZagaynov:
@Fryguy @Ladas
As per our discussion with @bronaghs this PR only moves the data to the yaml file, and updates it with a new flavors. I've described further plans in the top comment.
Data updates: https://goo.gl/m9JzCd
@Fryguy I responded to your question about `description` and other values here: https://github.com/ManageIQ/manageiq-providers-amazon/pull/449#issuecomment-406310448

Fryguy:
I believe there is a way to ignore `db/fixtures/aws_instance_types.yml` for the bot to avoid the flood...I *think* you just edit the .yamllint file at the root.

AlexanderZagaynov:
@Fryguy ok, I moved that change from that PR to the current one

.yamllint
```diff
@@ -11,3 +11,4 @@ rules:
     indent-sequences: false
   line-length:
     max: 120
+  trailing-spaces: false
```
Fryguy: Instead of adding a rule that ignores trailing spaces across the board, I think it's preferable to just add the aws_instance_types.yml to the ignore list above, since it's autogenerated.

db/fixtures/aws_instance_types.yml
```diff
@@ -0,0 +1,4055 @@
+---
+c1.medium:
```
Fryguy: What is this sorted by?  As mentioned earlier, having it sort by name will allow future diffs to be a lot more readable.

app/models/manageiq/providers/amazon/instance_types.rb
```diff
@@ -1,2594 +1,29 @@
...
-    "h1.4xlarge"  => {
-      :name                    => "h1.4xlarge",
-      :family                  => "Storage optimized",
-      :description             => "H1 Quadruple Extra Large",
-      :memory                  => 64.0.gigabytes,
-      :vcpu                    => 16,
-      :ebs_only                => false,
-      :instance_store_size     => 4000.0.gigabytes,
-      :instance_store_volumes  => 2,
-      :architecture
```
Fryguy: No need for the `== true`

Fryguy: Also don't need the `== true`

Fryguy: I don't really understand the need to deep_freeze all over the place, but I'll not let it stop the PR.

Fryguy:
@AlexanderZagaynov Some comments above.  Additionally, please squash old commits as they no longer make sense giving the feature is not being added (feature added then removed is not good for the commit history).  Can you also edit the original PR post to remove the "Closes #396 and #397" and other details no longer relevant now that this PR has a different meaning?

app/models/manageiq/providers/amazon/instance_types.rb
Fryguy: I have no idea why Github thinks I commented here :/   I commented on the lines that say `== true`

Fryguy: And here I commented on the line that says `.each_value(&:freeze).freeze`

bronaghs:
@AlexanderZagaynov  -  no need to backport since this is a not a fix for a blocker bug.
I would prefer not to have the `freeze` all over the place, it isnt necessary.

db/fixtures/aws_instance_types.yml
AlexanderZagaynov: yes, exactly, it is sorted by name

app/models/manageiq/providers/amazon/instance_types.rb
AlexanderZagaynov: done

AlexanderZagaynov: done

AlexanderZagaynov: I'm freezing that, because it is a constant. I believe that constant should surely be constant.

.yamllint
AlexanderZagaynov: done

AlexanderZagaynov:
@Fryguy @bronaghs done on remarks.
However, I didn't squashed commits entirely, but removed unnecessary ones, and organized rest in a meaningful steps.
P.S. Instance types data update diff: https://goo.gl/p2qFKq

Fryguy:
LGTM.  @bronaghs Not sure about the description changes as those are user-facing, but if you are good with those, then this is good to merge.

PR #452, opened 2018-06-08
Fix missing MiqServer in Authenticator spec

agrare:
The new audit_failure method in authenticator/base raises an evm event
on auth failure which targets the MiqServer.my_server.  In the amazon
authenticator spec this was nil which was causing the audit_failure
method to throw an exception trying to get class and id from the target.

https://bugzilla.redhat.com/show_bug.cgi?id=1318353


djberg96:
For reference: https://github.com/ManageIQ/manageiq/pull/17508

👍

PR #453, opened 2018-06-11
[WIP] Use ansible power actions

Ladas:
Use ansible power actions

Use simple ansible role for doing start&stop of a VM

TODO:
- [ ] fill out all the needed into about the role, right now it's basically what comes from galaxy-init 

Depends on:
- [ ] https://github.com/ManageIQ/manageiq/pull/17564
- [ ] https://github.com/ManageIQ/manageiq-api/pull/398

Examples running action from rails c
-----------------------

```ruby
# Start VM
pp ManageIQ::Providers::Amazon::CloudManager.first.vms.find_by_name("ladas_ansible_test_2__0").raw_start
```

```ruby
# Stop VM
pp ManageIQ::Providers::Amazon::CloudManager.first.vms.find_by_name("ladas_ansible_test_2__0").raw_stop
```

```ruby
# Create security group
pp ManageIQ::Providers::Amazon::NetworkManager::SecurityGroup.raw_create_security_group(ManageIQ::Providers::Amazon::CloudManager.first.network_manager, {"vpc_id" => ManageIQ::Providers::Amazon::CloudManager.first.cloud_networks.first.ems_ref, "security_group_description" => "ladas_test_changed", "security_group_name"=> "ladas_test", "security_group_rules" => [{:proto => 'tcp', :from_port => 22, :to_port => 22, :cidr_ip => '0.0.0.0/0'}]})
```

Examples with API
-------------------------

```
# getting a token
curl --user admin:smartvm -i -X GET -H "Accept: application/json" http://localhost:3001/api/auth

# set the token env var
TOKEN=xyz

# Get the ID of a network manager, in my case it's 4, use it in 'providers/4' api

# Create security group, modify the name, otherwise it just modifies the existing sc
curl -i -X POST -H "Accept: application/json" -H "X-Auth-Token: $TOKEN" http://localhost:3001/api/providers/4/security_groups --data '{ "action" : "create", "resource" : {"vpc_id":"vpc-ff49ff91","security_group_description":"ladas_test_changed","security_group_name":"ladas_test5","security_group_rules":[{"proto":"tcp","from_port":22,"to_port":22,"cidr_ip":"0.0.0.0/0"}]}}'

# Get VM ID
VM_ID=xyz # get some AWS vm id from MIQ DB, e.g. 
ManageIQ::Providers::Amazon::CloudManager.first.vms.find_by_name("ladas_ansible_test_2__0").id

# Invoke special start vm
curl -i -X POST -H "Accept: application/json" -H "X-Auth-Token: $TOKEN" http://localhost:3001/api/instances/$VM_ID --data '{ "action" : "special_start" }'

# Invoke special stop
curl -i -X POST -H "Accept: application/json" -H "X-Auth-Token: $TOKEN" http://localhost:3001/api/instances/$VM_ID --data '{ "action" : "special_stop" }'
```


Ladas:
@Fryguy @agrare @gtanzillo so I've just placed the content under ansible/ dir, then it follows https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#directory-layout

then I get the path using engine root https://github.com/ManageIQ/manageiq-providers-amazon/pull/453/files#diff-235184baadf210fc9a4d037d8ffa1c37R82

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -71,6 +71,18 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
...
+  def ansible_env_vars
+    {
+      'EC2_ACCESS_KEY' => default_authentication.userid,
+      'EC2_SECRET_KEY' => default_authentication.password,
+      'EC2_REGION'     => provider_region,
+    }
+  end
+
+  def ansible_root
+    ManageIQ::Providers::Amazon::Engine.root.join("ansible")
```
agrare: I think we talked about having this be `Engine.root.join("content/ansible")`

Ladas: ok, I can do that

Ladas: done

Fryguy:
Is this just PoC at the model level or is this your expected implementation?  I'm curious on the performance aspect of using ansible runner over aws-sdk directly.

Ladas:
@Fryguy right, for now, it is just a POC we can use for demo/examples. I think we could go this way implementing CRUD actions on objects (if it will ever be an action item)

Speed wise, I think most of the modules do at least 2 requests, 1 for trying to fetch the resource first. But I would need to dig into the module to verify (or capture a VCR). Although it should be like +-1s difference, so nothing too horrible.

miha-plesko:
@Ladas @agrare is this a valid approach? Can we perform integration on Nuage as well taking this PR as example?

Ladas:
@miha-plesko We are almost there. The code in models is how it should look like from the backend side. 

From the ansible side of view, the format is just recommended ansible galaxy format, so you can start there. We want it in content/ansible, so it's usable in embeded ansible too, but we have to do some changes for that (that is why I have it in content_tmp)

And we are trying to refine the API approach and the UI is being finished. 

I think you can safely work on the ansible galaxy roles (we will allow requirements.yml, so the roles code can live only in galaxy) and the models. The rest should be finished and documented in 1-2 weeks.

agrare:
@Ladas can this be closed now that we have documentation and an example in the dummy_provider?

PR #456, opened 2018-06-21
Persister: optimized InventoryCollection definitions

slemrmartin:
Optimizations and simplification of InventoryCollection definitions

**Issue**: https://github.com/ManageIQ/manageiq/issues/17396
- ((**depends on**): https://github.com/ManageIQ/manageiq/pull/17621

What is possible is moved to core


slemrmartin:
cc @agrare

PR #460, opened 2018-07-09
[WIP] Toolbar overrides

martinpovolny:
Do not merge. This is demo.

### Demo toolbar plugin from provider repos running ansible actions

Requires (merged)
 * https://github.com/ManageIQ/manageiq-ui-classic/pull/4267 
 * https://github.com/ManageIQ/manageiq-ui-classic/pull/4262 -- here is some more explanation

Includes: https://github.com/ManageIQ/manageiq-providers-amazon/pull/462

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,31 @@
+{
+  "presets": [
+    [
+      "env",
+      {
+        "targets": {
+          "browsers": [
+            "> 1%",
+            "last 2 versions",
+            "Firefox ESR",
+            "IE 10",
+            "IE 11"
+          ]
+        }
+      }
+    ],
+    "react",
+    "es2015",
+  ],
+  "plugins": [
+    "transform-class-properties",
+    "transform-export-extensions",
+    "transform-object-rest-spread",
+    "transform-object-assign"
+  ],
+  "env": {
+    "test": {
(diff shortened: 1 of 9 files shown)


app/toolbar_overrides/ems_cloud_center.rb
```diff
@@ -0,0 +1,33 @@
+module ManageIQ
+  module Providers
+    module Amazon
+      module ToolbarOverrides
+        class EmsCloudCenter < ::ApplicationHelper::Toolbar::Override
```
Fryguy: move this to app/<foo>/manageiq/providers/amazon/toolbar_overrides/ems_cloud_center.rb

...where <foo> could be any of the autoload directories...I'm thinking maybe helpers or a new directory?  We can put whatever we want really...if it's a "new" directory, it just needs to be in application.rb in core, but that's easy.  What is the structure you use for toolbar stuff in ui-classic?

martinpovolny: sure, can move it from 'app/toolbar_overrides' to where you wrote. will give it a shot

in ui-classic the toolbars live under `app/helpers/application_helper/toolbar/` each toolbar has it's class

so in the provider repos I'd like to have `<something>/toolbar_overrides/` and then each overrride in it's own class

PR #461, opened 2018-07-09
refresh S3 target at rearmost

AlexanderZagaynov:


AlexanderZagaynov:
@Ladas @tumido Tom have concern about such sorting.
Let's discuss it here?

tumido:
Yes, I do! :laughing: 

I just think moving the `StorageManager` at the end of the queue, every time a job is picked, is a bad idea. Imagine a situation when and `CloudManager` and `NetworkManager` refreshes are scheduled repetitively. So before `NetworkManager` finishes, there's already a `CloudManager` in the queue. In such situation the `StorageManager` might never get picked up. And this scheduling race condition can happen by event triggered targeted refreshes, or click-spamming the refresh button, etc..

Scenario:

1.   Initial refresh queue:
```
[0] Storage
[1] Cloud
[2] Network
```
2. `Storage` would be moved to the end and `Cloud` gets picked up
```
[0] Network
[1] Storage
```
3. `Network` gets picked up
```
[0] Storage
```
4. At this point (before next job selection), a new full `Cloud` refresh arrived, so the `Storage` is pushed the the end again and `Cloud` is picked
```
[0] Storage
[1] Cloud
```
5. A full `Cloud` refresh schedules a `Network` refresh, so the `Network` is added to the queue and `Storage` has to be jumped again 
```
[0] Storage
[1] Network
```

Result:
The `Storage` refresh might never happen.

That's the concern! :smiley:

Ladas:
@tumido that is a valid concern. @AlexanderZagaynov I think the best will be to make a dedicated refresh worker for s3 again

AlexanderZagaynov:
@Ladas @tumido please review

app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
```diff
@@ -12,6 +12,14 @@ def save_inventory(ems, target, _hashes_or_persister)
     EmsRefresh.queue_refresh(ems.ebs_storage_manager) if target.kind_of?(ManageIQ::Providers::BaseManager)
   end
 
+  def preprocess_targets
```
Ladas: this should not be needed? If we never queue it as a part of cloud manager?

juliancheal: I'm not sure I understand what this bit is for.

agrare: Yeah I don't think this is needed because by changing `combined_managers` this worker won't ever dequeue a refresh for the S3 manager

agrare:
@AlexanderZagaynov you might have already but if not can you test this with a full evmserverd to make sure the queueing/workers bit works as expected?

juliancheal:
@AlexanderZagaynov I assume you've ran this on your machine. Can you show what workers are running, when you run a full refresh.

app/models/manageiq/providers/amazon/cloud_manager/refresh_worker.rb
```diff
@@ -14,7 +14,10 @@ def queue_name_for_ems(ems)
     private
 
     def combined_managers(ems)
-      [ems].concat(ems.child_managers)
+      child_managers = ems.child_managers.reject do |child_ems|
+        child_ems.kind_of?(ManageIQ::Providers::Amazon::StorageManager::S3)
```
agrare: I'm okay with this for now, but wonder if we could have a more general way of solving this by adding a method on `ext_management_system` like `#combine_with_parent_manager?`

AlexanderZagaynov: I guess it requires more DRYing, since that method presents (exactly the same code) in other providers.
- https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager/refresh_worker.rb#L16
- https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/refresh_worker.rb#L16

And I don't think it should be moved in an `ExtManagementSystem` because its only used there.

It can be probably moved here:
- https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/refresh_worker.rb

And be overriden in subclasses as needed.

agrare: Yeah makes sense to put this in the BaseManager::RefreshWorker

juliancheal:
@AlexanderZagaynov that's great to see S3 as a solo worker 👍

app/models/manageiq/providers/amazon/cloud_manager/refresher.rb
Ladas: @AlexanderZagaynov can we remove this method (it shouldn't be needed)

AlexanderZagaynov: @Ladas Have you seen that comment? https://github.com/ManageIQ/manageiq-providers-amazon/pull/461#issuecomment-409240480
Looks like it processes all the targets without current code.

AlexanderZagaynov:
@agrare I believe `ManagerRefresh::InventoryCollectionDefault` removal was definitely not caused by a current PR. It was possibly removed by some other commit. And I'm wondering how that was able to pass unnoticed.

Ladas:
@AlexanderZagaynov I'll send a PR that moves the spec to the correct location, with correct described class

agrare:
@AlexanderZagaynov it is because we removed InventoryCollectionDefault from core a while ago so it started failing here.

We should not just remove this test, but rather move it to another location and in another PR not this one

AlexanderZagaynov:
@Ladas @agrare I reverted my latest commit with that failing spec removal.
Are you going to merge current PR now or after Ladas's PR will be merged?

agrare:
@AlexanderZagaynov after the test fix is merged we will close & reopen this and when it is green I will merge it

AlexanderZagaynov:
@agrare ok, got it. please notice that since you've merged related core repo PR - it can possibly cause an issue shown on a first picture above, without current PR merge.

djberg96:
@AlexanderZagaynov @agrare @Ladas Is this something you feel is worth backporting?

PR #462, opened 2018-07-09
[WIP] Add JS in general, amazon-security-group component

himdel:
The component comes from https://github.com/ManageIQ/manageiq-ui-classic/pull/4181,
the rest of the JS config is just the needed bits from ui-classic, and dependencies.

Differences from the original:

* (js) `API` comes from `window.API`, not imported directly
* (js) `addReact` comes from `ManageIQ.component.addReact`, not imported directly (should probably change this in ui-classic too)
* (css) not included via the asset pipeline, use webpack (=> gaprindashvili/no)
* (props) use `recordId` instead of `providerId` in the wrapper (toolbar interface)
* (props) make `recordId` string, not number (bigint)

---

Meant to go together with https://github.com/ManageIQ/manageiq-ui-classic/pull/4267 and https://github.com/ManageIQ/manageiq-providers-amazon/pull/460 .. and https://github.com/ManageIQ/manageiq-ui-classic/pull/4268 .. and https://github.com/ManageIQ/manageiq-ui-classic/pull/4262

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,31 @@
+{
+  "presets": [
+    [
+      "env",
+      {
+        "targets": {
+          "browsers": [
+            "> 1%",
+            "last 2 versions",
+            "Firefox ESR",
+            "IE 10",
+            "IE 11"
+          ]
+        }
+      }
+    ],
+    "react",
+    "es2015",
+  ],
+  "plugins": [
+    "transform-class-properties",
+    "transform-export-extensions",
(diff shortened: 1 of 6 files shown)


himdel:
cc @martinpovolny

himdel:
Updated to use `recordId` instead of `providerId` (props), to match https://github.com/ManageIQ/manageiq-ui-classic/pull/4267

PR #466, opened 2018-07-18
Add more public image filters

djberg96:
To help reduce the number of public images collected this adds a couple extra parameters, and suggests an additional one (owner-id) to help restrict it even further.

The equivalent CLI command would be this: `aws ec2 describe-images --filters "Name=owner-id,Values=xxx" "Name=image-type,Values=machine" "Name=is-public,Values=true" "Name=state,Values=available"`

Note that this won't significantly reduce the number of collected images without the `owner-id` filter, but at least we have it, and the users can see the suggestion for how to heavily reduce the results. In our local dev environment the number was reduced from over 100k to just 14 once I set the owner-id filter, for example.


Fryguy:
For the yamllint issue, arrays should be aligned with the key entry:

```yaml
:some_key:
- array_val1
- array_val2
```

is what is expected as opposed to

```yaml
:some_key:
  - array_val1
  - array_val2
```

Ladas:
@djberg96 this is not covered by a VCR, so manual test was ok? (we could do at least stubbed spec, that would at least test that the params are valid?). And how does this change the amount of images fetched (e.g. for the us-east, N.Virginia with 108k)

PR #472, opened 2018-08-02
Deal with amazon filter limit

Ladas:
Deal with amazon filter limit. First every query using filter will do max 200 items at a time, if more targets are requested, we will do multiple queries. (it's a hard limit in AWS, so requesting more will lead to 500 error)

Then, do not fetch all current images as extra. This can slowdown
refresh significantly, e.g. if we enable and then disable all public
images collection (which would lead to fetching e.g. 100k images 
by 200 batches). We will keep fetching only images used by instances,
that were not included in current public/private image filters, so
it can't happen we have instance without image.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1612086


app/models/manageiq/providers/amazon/inventory/collector.rb
```diff
@@ -99,4 +99,13 @@ def health_check_members(load_balancer_name)
   def health_check_members_refs(load_balancer_name)
     @health_check_members_refs.try(:[], load_balancer_name) || []
   end
+
+  def max_filter_size
+    200
```
agrare: I assume this isn't some constant in the aws-sdk gem?

app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
```diff
@@ -74,10 +76,9 @@ def extra_image_references
     # public_images
     return @extra_image_references if @extra_image_references
 
-    db_image_refs        = Set.new(manager.miq_templates.pluck(:ems_ref))
     instances_image_refs = Set.new(instances.map { |x| x["image_id"] })
     api_images_refs      = Set.new((private_images + shared_images + public_images).map { |x| x["image_id"] })
 
-    @extra_image_references = ((db_image_refs + instances_image_refs) - api_images_refs).to_a
+    @extra_image_references = (instances_image_refs - api_images_refs).to_a
```
agrare: So will this never pull in public images which are not in use by one of the instances?  Does that match with the use-case of provisioning from a public image?

app/models/manageiq/providers/amazon/inventory/collector.rb
Ladas: haven't found that, that would be nice :-)

app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
Ladas: we will still pull in public images, if those are allowed, byt the specified filter

This is if the public images were not collected even then (either public images are disabled, or filter misses them). So we do have an old BZ saying that we always want image, because without image, we won't see OS info). 

So without this code, if you delete all Vms using the public image, we will disconnect it. Previously it would stay, but that is probably not possible to make in an efficient way.

agrare: Ahh okay I misunderstood, :+1: thanks

simaishi:

```
$ git log -1
commit 83cbdd26dbf0bc6e1d4891025c9ab2fc91866bcd
Date:   Fri Aug 3 10:51:50 2018 -0400

    Merge pull request #472 from Ladas/deal_with_amazon_filter_limit
    
    Deal with amazon filter limit
    
    (cherry picked from commit 2b48af4225eb978b44c8ef6b65893a7b9d2aa395)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1651347
```

PR #476, opened 2018-08-20
remove bare metal instance

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1590288#c6

There was a bare metal instance in the virtual instance types list. Now it removed.


cben:
Travis error is missing log/ dir, fixed in manageiq master, close the PR and re-open should fix

bronaghs:
thanks @cben

PR #479, opened 2018-08-27
Add in missing user creation audit event.

juliancheal:
Recieved from this PR https://github.com/ManageIQ/manageiq/pull/17852

### DIFF
--- a/spec/models/authenticator/amazon_spec.rb
+++ b/spec/models/authenticator/amazon_spec.rb
@@ -375,11 +375,16 @@ describe Authenticator::Amazon do
           expect(authenticate).to eq(123)
         end
 
-        it "records two successful audit entries" do
+        it "records three successful audit entries" do
           expect(AuditEvent).to receive(:success).with(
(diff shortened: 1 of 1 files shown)


juliancheal:
@agrare tests are passing locally, but I'm not sure where `userNameType ` in the message
```
message => "User creation successful for User: userNameType with ID: #{username}"
```

is coming from :(

agrare:
@juliancheal looks like it is coming from here
```
   83:     def update_user_attributes(user, username, amazon_user)
=> 84:       user.userid = username
   85:       user.name   = amazon_user.name
   86:     end
   87: 
   88:     private
(byebug) amazon_user
#<Aws::IAM::User name="userNameType">
```

agrare:
I think we can improve the stubbed user to give it a better username, merging this for now to get the repo back to green.

juliancheal:
@agrare ah are we getting that from the user stubbing? I see, that explains why I hadn't seen it I was testing the original PR with actual amazon log in.

PR #480, opened 2018-08-27
Inventory Builder inherits from core

slemrmartin:
common functionality extracted to `ManageIQ::Providers::Inventory::Builder`


slemrmartin:
@Ladas Amazon is a bit special with its builder, but it's still worth to inherit from core. Please review

app/models/manageiq/providers/amazon/inventory.rb
```diff
@@ -2,4 +2,22 @@ class ManageIQ::Providers::Amazon::Inventory < ManageIQ::Providers::Inventory
...
+  def self.default_manager_name
+    "CloudManager"
+  end
+
+  def self.parser_classes_for(ems, target)
+    case target
+    when ManagerRefresh::TargetCollection
+      [ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager,
+       ManageIQ::Providers::Amazon::Inventory::Parser::NetworkManager,
+       ManageIQ::Providers::Amazon::Inventory::Parser::StorageManager::Ebs]
```
agrare: I wonder if we could do something generally with child_manager_types on ExtManagementSystem or something...

PR #482, opened 2018-08-29
avoid blank names by using uids

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1531117

Resource name tag can be not defined or a blank string.
Resource ID will be used in such case.

My approach:
1) Change all `get_from_tags` methods to make them return `nil` instead of blank strings, 
to make `||` code behave as expected.
```
'' || 'uid' => ''
''.presence || 'uid' => 'uid'
```
2) Find all `name` code references, and fix those, where it can have any sense.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/refresh_parser.rb
@@ -303,7 +303,7 @@ class ManageIQ::Providers::Amazon::CloudManager::RefreshParser < ManageIQ::Provi
     new_result = {
       :type                   => ManageIQ::Providers::Amazon::CloudManager::OrchestrationStack.name,
       :ems_ref                => uid,
-      :name                   => stack.name,
+      :name                   => stack.name.presence || uid,
       :description            => stack.description,
       :status                 => stack.stack_status,
       :status_reason          => stack.stack_status_reason,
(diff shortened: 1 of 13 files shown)


AlexanderZagaynov:
@Ladas @bronaghs please review

djberg96:
Interestingly, other database vendors treat blank strings as nulls but I guess Postgres doesn't. I was thinking that maybe instead of fiddling with the the parsers, we create a custom handler in the model itself. That way we only have to change it in one place.

Ladas:
@djberg96 better to have it in parser, there are strategies bypassing models (since creating AR object just to assign hash is pretty expensive)

djberg96:
@Ladas Sorry, I'm not following. Why would putting this in orchestration_stack.rb be a bad idea?

```
  def name=(value)
    super(name.presence || uid)
  end
```

Ladas:
@djberg96 right, what I am saying is, if you'll use batch saving with :use_ar_object => false, this will not be called. So it's better to have the transformation logic in refresh. 

(since only refresh owns that data, it doesn't make much sense to have it in models anyway)

djberg96:
I guess my question is why we ignore AR for batch saving, seems like a pretty huge pitfall in general.

Ladas:
@djberg96 well there is the :use_ar_object, so you can choose ignore or not ignore. In general, we go from business logic in models, since nobody else than refresh can set these things. And the models are slow (and in 99% of cases, it doesn't do anything, it's just hash -> AR record -> same hash -> DB)

djberg96:
Well, I've made my case. If you think we need to worry about batch saves for orchestration templates in Amazon, then ok. Otherwise, I think my proposal is better.

AlexanderZagaynov:
@Ladas so, should it be merged, or I need to something else?

simaishi:
Hammer backport details:

```
$ git log -1
commit a5dd7c4a01e5f65f30ec478288b12a2eda8f2468
Date:   Tue Sep 25 20:45:17 2018 +0200

    Merge pull request #482 from AlexanderZagaynov/BZ-1531117_name_tag
    
    avoid blank names by using uids
    
    (cherry picked from commit 4197be8d1dee08a18c87483cf5ada03bebb6f5c1)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1531117
```

PR #483, opened 2018-08-31
Collect AWS service catalog items

Ladas:
Collect AWS service catalog items.

For refresh it needs role:
ServiceCatalogAdminReadOnlyAccess

TODO:
- [x] specs

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/17933


Ladas:
@gtanzillo @gmcculloug @syncrou not sure if we should require the admin access? Or what permissions you would envision we would have.

Also it looks like I can only see provisioned instances of what my account provisioned, I haven't found any admin view of all provisioned instances.

app/models/manageiq/providers/amazon/inventory/collector.rb
```diff
@@ -68,6 +68,10 @@ def aws_cloud_formation
     @aws_cloud_formation ||= manager.connect(:service => :CloudFormation)
   end
 
+  def aws_service_catalog
+    @aws_service_catalog = manager.connect(:service => :ServiceCatalog)
```
agrare: Should be `@aws_service_catalog ||=`

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -342,4 +344,64 @@ def key_pairs
       )
     end
   end
+
+  def service_offerings
+    collector.service_offerings.each do |service_offering|
```
agrare: What is the order of magnitude of the number of service offerings?  Hundreds/thousands/etc...
Doing API calls in a loop over a collection sucks :/

Ladas: not sure, but there'll be few of those. Only way to make it work nicely is multi-thread + targeted refresh (if possible)

app/models/manageiq/providers/amazon/inventory/collector.rb
Ladas: done

Issue #465, opened 2018-07-18
Upgrade AWS SDK library to version 3


djberg96:
We are currently using version 2 of the aws-sdk for Ruby. Version 3 has been out for a few years now, so we should probably upgrade.

AlexanderZagaynov:
@djberg96 I had to make changes in 3 gems. Now I need to test that somehow on Travis.
Problem that those gems have cross-dependencies, and can't be tested in a usual way.
They all should be tested on update branches of each other. Maybe you have some idea, how to do that?
Maybe we can create `aws_sdk_v3` branches in all of those MiQ repos, and point dependencies to it,
then test everything against those?

@tumido you did similar change on Google, maybe you can help?

And next thing I'm thinking about - is that since v3 gem was made modular, maybe we can just pick up only needed sub-gems? Need to do more research.

jerryk55:
@AlexanderZagaynov someone has to be sure that this change does not impact the SSA functionality  of amazon_ssa_support.

AlexanderZagaynov:
@jerryk55 can current specs suite provide such confidence?

jerryk55:
Not entirely sure.  Probably should rope @hsong-rh into the discussion.  He did most of the work on that gem...

hsong-rh:
@AlexanderZagaynov Current specs only use mocked data to test SSA logics, may not find aws sdk APIs problems, if there is any. It's better to run a real SSA job with V3 gem.

AlexanderZagaynov:
@hsong-rh frankly speaking, I have no idea how to do that :)
can you point me on?

AlexanderZagaynov:
Here is the logs from local testing. They have several weird errors, I'm going to investigate them.
[miq-core_aws-sdk_v3.log](https://github.com/ManageIQ/manageiq-providers-amazon/files/2405703/miq-core_aws-sdk_v3.log)
[miq-amazon_aws-sdk_v3.log](https://github.com/ManageIQ/manageiq-providers-amazon/files/2405704/miq-amazon_aws-sdk_v3.log)
[miq-ssa_aws-sdk_v3.log](https://github.com/ManageIQ/manageiq-providers-amazon/files/2405706/miq-ssa_aws-sdk_v3.log)
[miq-pending_aws-sdk_v3.log](https://github.com/ManageIQ/manageiq-providers-amazon/files/2405707/miq-pending_aws-sdk_v3.log)

[...]

PR #488, opened 2018-09-21
update aws-sdk gem to v3

AlexanderZagaynov:
This is a part of global update `aws-sdk` gem to `v3`

Together with:
- https://github.com/ManageIQ/manageiq/pull/19411
- https://github.com/ManageIQ/manageiq-gems-pending/pull/449
- https://github.com/ManageIQ/amazon_ssa_support/pull/44

It should fix https://github.com/ManageIQ/manageiq-providers-amazon/issues/465


manageiq-providers-amazon.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_dependency("aws-sdk", ["~>2.9.7"])
+  s.add_dependency("aws-sdk", ["~> 3"])
```
Fryguy: Same comment regarding the gem version, and also remove all of the parens and square brackets (go for consistency with the lines below)

AlexanderZagaynov:
Hi @Fryguy ! Done.

manageiq-providers-amazon.gemspec
AlexanderZagaynov: done

AlexanderZagaynov:
@NickLaMuro If you mean I should revert / adopt those commits - I'll do. 
There is several testing issues anyway, need to change more.

AlexanderZagaynov:
@NickLaMuro that was fixed in https://github.com/aws/aws-sdk-ruby/pull/1760, thanks to @Fryguy :)

https://github.com/aws/aws-sdk-ruby/blob/22f10bed9c6cfb3317b9fc684c124168c5dca71e/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb#L262

So, I'm simply removing that patch entirely.

AlexanderZagaynov:
@miq-bot add_label gaprindashvili/yes
@miq-bot add_label hammer/yes

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
```diff
@@ -200,9 +200,19 @@ def get_network_ports
   end
 
   def get_ec2_floating_ips_and_ports
-    instances = @aws_ec2.instances.select { |instance| instance.network_interfaces.blank? }
-    process_collection(instances, :network_ports) { |instance| parse_network_port_inferred_from_instance(instance) }
-    process_collection(instances, :floating_ips) { |instance| parse_floating_ip_inferred_from_instance(instance) }
+    instances = @aws_ec2.instances
+    instances = instances.select do |instance|
+      interfaces = instance.network_interfaces
+      interfaces.blank? || interfaces.size.zero?
```
Fryguy: I don't understand this... `.blank?` (normally) checks for empty collections, so is there a time when the are 0 items, and `blank?` returns `false`?

Fryguy:
Aside from my comment above, this LGTM.  @NickLaMuro not sure why you have a request change on this, but what do you think now?

NickLaMuro:
@Fryguy was mostly using that as a reminder that the PRs that I merged to patch `aws-sdk` have been reverted:

https://github.com/ManageIQ/manageiq-providers-amazon/pull/488/commits/11f987fcddf87d184a9673b5318a78defec445ca
https://github.com/ManageIQ/manageiq-providers-amazon/pull/488/commits/1879e5cc8303d27bdb78ed1c17d77bd9adf856be

Which they have.

But also, that [I was suggesting](https://github.com/ManageIQ/manageiq-providers-amazon/pull/488#pullrequestreview-157727920) that we wait to confirm the tests failed when the other PRs have been fixed, addressed and merged, since those should be merge-able without requiring other PRs to work.  Reason being is that test is in there to confirm we don't have a regression since the fix ( #432 ) really didn't provide a spec of it's own to validate.

Since it seems that `amazon_ssa_support` also has a `manageiq-gems-pending` dependency, and merging these separately would basically cause developers to not be able to bundle for a period of time, waiting to do the above might not be possible.

* * *

Anyway, I will remove my ❌ since it doesn't seem like there is an easy solution for this.

app/models/manageiq/providers/amazon/network_manager/refresh_parser.rb
AlexanderZagaynov: The reason is that its not kind of `Enumerable`. 
This fix is intentional, without it there was few related issues.

AlexanderZagaynov:
@NickLaMuro other repos were updated as well. Check the base issue, it has all the links: https://github.com/ManageIQ/manageiq-providers-amazon/issues/465

NickLaMuro:
@AlexanderZagaynov I am referring to why Travis is currently failing:

```
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies.......
Bundler could not find compatible versions for gem "aws-sdk":
  In Gemfile:
    manageiq-gems-pending (> 0) was resolved to 0.1.0, which depends on
      aws-sdk (~> 2.9.7)
    manageiq-providers-amazon was resolved to 0.1.0, which depends on
      aws-sdk (~> 3.0)
```

To get this to pass prior to merging this (in CI), and do what I suggested in the previous comment would probably have some undesirable affects for the rest of the team in the mean time.

Again, I removed my "Request changes" review.

djberg96:
@AlexanderZagaynov What's the word?

AlexanderZagaynov:
> @AlexanderZagaynov What's the word?
in a month :)

djberg96:
@AlexanderZagaynov I think you'll need to update the gemspec, too.

AlexanderZagaynov:
@djberg96 you mean this: https://github.com/ManageIQ/manageiq-providers-amazon/pull/488/files#diff-19237fe8b6348694e23d6c2579276b6d ?

djberg96:
@AlexanderZagaynov Oh, I didn't see it. I guess it's gems-pending that's the culprit.

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -89,7 +88,6 @@ def connection_rescue_block
 
     def translate_exception(err)
       require 'aws-sdk'
-      require 'patches/aws-sdk-core/seahorse_client_net_http_pool_patch'
```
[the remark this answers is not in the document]
AlexanderZagaynov: done

AlexanderZagaynov:
All tests are green on my machine with patched versions of all gems mentioned here.

AlexanderZagaynov:
@Fryguy @NickLaMuro @agrare please review

agrare:
Will merge when the cross-repo test is green https://travis-ci.org/agrare/manageiq-cross_repo/builds/600907361

PR #500, opened 2018-11-16
Always set vm_or_template.connection_state

agrare:
Instead of defaulting to nil always set vm_or_template.connection_state


agrare:
@Ladas can you take a look?  This is the last PR for this.

PR #520, opened 2019-03-25
Update supported ruby versions on Travis

juliancheal:
Update supported ruby versions on Travis


agrare:
Looks like Hakiri job just failed

PR #522, opened 2019-03-29
collect metrics from CloudWatch Agent

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1684525

Gather data with the new metric names from CloudWatch Agent
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Install-CloudWatch-Agent.html

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
@@ -24,15 +24,15 @@ class ManageIQ::Providers::Amazon::CloudManager::MetricsCapture < ManageIQ::Prov
         'MemoryUtilization', 'mem_used_percent', # Linux
         'Memory % Committed Bytes In Use',       # Windows
       ].freeze,
-      :calculation           => ->(stat, _) { stat },
+      :calculation           => ->(*stats, _) { stats.compact.mean },
       :vim_style_counter_key => "mem_usage_absolute_average",
     }.freeze,
     {
       :amazon_counters       => [
         'SwapUtilization', 'swap_used_percent', # Linux
-        'Paging File % Usage',                  # Windows
+        'Paging File % Usage',                  # Windows TODO: calc is wrong for more than one paging file (AZ)
       ].freeze,
-      :calculation           => ->(stat, _) { stat },
+      :calculation           => ->(*stats, _) { stats.compact.mean },
       :vim_style_counter_key => "mem_swapped_absolute_average",
     }.freeze,
   ].freeze
@@ -130,11 +130,13 @@ class ManageIQ::Providers::Amazon::CloudManager::MetricsCapture < ManageIQ::Prov
 
     counters_by_id              = {target.ems_ref => VIM_STYLE_COUNTERS}
     counter_values_by_id_and_ts = {target.ems_ref => counter_values_by_ts}
+
     return counters_by_id, counter_values_by_id_and_ts
   end
 
   def counter_values_by_timestamp(metrics_by_counter_name)
     counter_values_by_ts = {}
+
     COUNTER_INFO.each do |i|
       timestamps = i[:amazon_counters].collect do |c|
         metrics_by_counter_name[c].keys unless metrics_by_counter_name[c].nil?
@@ -143,6 +145,7 @@ class ManageIQ::Providers::Amazon::CloudManager::MetricsCapture < ManageIQ::Prov
       # If we are unable to determine if a datapoint is a 1-minute (detailed)
       #   or 5-minute (basic) interval, we will throw it away.  This includes
       #   the very first interval.
+
       timestamps.each_cons(2) do |last_ts, ts|
         interval = ts - last_ts
         next unless interval.in?(INTERVALS)
@@ -151,38 +154,47 @@ class ManageIQ::Providers::Amazon::CloudManager::MetricsCapture < ManageIQ::Prov
         value   = i[:calculation].call(*metrics, interval)
 
         # For (temporary) symmetry with VIM API we create 20-second intervals.
+
         (last_ts + 20.seconds..ts).step_value(20.seconds).each do |inner_ts|
           counter_values_by_ts.store_path(inner_ts.iso8601, i[:vim_style_counter_key], value)
         end
       end
     end
+
     counter_values_by_ts
   end
 
   def metrics_by_counter_name(cloud_watch, counters, start_time, end_time)
     metrics_by_counter_name = {}
+
     counters.each do |c|
-      metrics = metrics_by_counter_name[c.metric_name] = {}
+      metrics = metrics_by_counter_name[c.metric_name] ||= {}
+      options = c.to_hash.merge(:statistics => %w[Average].freeze, :period => 60).freeze
 
       # Only ask for 1 day at a time, since there is a limitation on the number
       #   of datapoints you are allowed to ask for from Amazon Cloudwatch.
       #   http://docs.amazonwebservices.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html
+
       (start_time..end_time).step_value(1.day).each_cons(2) do |st, et|
+        opts = options.merge(:start_time => st, :end_time => et).freeze
+
         statistics, = Benchmark.realtime_block(:capture_counter_values) do
-          options = {:start_time => st, :end_time => et, :statistics => ["Average"], :period => 60}
-          cloud_watch.client.get_metric_statistics(c.to_hash.merge(options)).datapoints
+          data = cloud_watch.client.get_metric_statistics(opts)
+          data.datapoints
         end
 
         statistics.each { |s| metrics[s.timestamp.utc] = s.average }
       end
     end
+
     metrics_by_counter_name
   end
 
   def get_counters(cloud_watch)
     counters, = Benchmark.realtime_block(:capture_counters) do
-      filter = [{:name => "InstanceId", :value => target.ems_ref}]
-      cloud_watch.client.list_metrics(:dimensions => filter).metrics.select { |m| m.metric_name.in?(COUNTER_NAMES) }
+      filter = [{ :name => "InstanceId", :value => target.ems_ref }]
+      data = cloud_watch.client.list_metrics(:dimensions => filter)
+      data.metrics.select { |m| m.metric_name.in?(COUNTER_NAMES) }
     end
     counters
   end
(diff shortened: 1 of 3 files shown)


AlexanderZagaynov:
@djberg96 @Fryguy @Ladas please review

app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
```diff
@@ -124,11 +130,13 @@ def perf_capture_data_amazon(cloud_watch, start_time, end_time)
 
     counters_by_id              = {target.ems_ref => VIM_STYLE_COUNTERS}
     counter_values_by_id_and_ts = {target.ems_ref => counter_values_by_ts}
+
     return counters_by_id, counter_values_by_id_and_ts
   end
 
   def counter_values_by_timestamp(metrics_by_counter_name)
     counter_values_by_ts = {}
+
```
agrare: @AlexanderZagaynov I don't disagree with these extra newlines (well not all of them at least) but they're completely unrelated to the rest of this PR.

Can you remove any unrelated whitespace changes?

app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
```diff
@@ -145,38 +154,47 @@ def counter_values_by_timestamp(metrics_by_counter_name)
...
     counter_values_by_ts
   end
 
   def metrics_by_counter_name(cloud_watch, counters, start_time, end_time)
     metrics_by_counter_name = {}
+
     counters.each do |c|
-      metrics = metrics_by_counter_name[c.metric_name] = {}
+      metrics = metrics_by_counter_name[c.metric_name] ||= {}
```
agrare: So are some of these counters updating different counters?

app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
```diff
@@ -145,38 +154,47 @@ def counter_values_by_timestamp(metrics_by_counter_name)
...
+      options = c.to_hash.merge(:statistics => %w[Average].freeze, :period => 60).freeze
```
agrare: Is this refactoring?  It looks like all the same keys are all still here

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
```diff
@@ -58,18 +58,26 @@
   end
 
   context 'counters present' do
-    let(:vm_name) { 'i-091369dd8a8d470ab' }
     let(:ems) { FactoryBot.create(:ems_amazon_with_vcr_authentication, :provider_region => 'eu-central-1') }
+    let(:data) { vm.perf_collect_metrics('realtime').last.first.last }
```
agrare: Maybe break this up into more variables so it is more descriptive?  the `.last.first.last` is really opaque

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
```diff
@@ -58,18 +58,26 @@
...
-      with_vcr_data { vm.perf_collect_metrics('realtime').last.first.last.first.last }
+    shared_examples 'available' do
+      it { is_expected.to have_key('mem_usage_absolute_average') }
+      it { is_expected.to have_key('mem_swapped_absolute_average') }
     end
 
-    it { is_expected.to have_key('mem_usage_absolute_average') }
-    it { is_expected.to have_key('mem_swapped_absolute_average') }
+    context 'from linux old way' do
+      subject { with_vcr_data { data.first.last } }
```
agrare: So this ends up being `vm.perf_collect_metrics('realtime').last.first.last.first.last`?

app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
AlexanderZagaynov: Metric can have a different set of dimensions, <br>depending on CloudWatch Agent settings:

1 dimension sets


In my case, when they were changed:
*_(CloudWatch keeps 1 minute interval metrics for 2 weeks)_*

2 memory 1
2 memory 2
2 memory 3


And dimensions can have different values too:

3 pagefile 1
3 pagefile 2

AlexanderZagaynov: My thoughts:
- It makes no practical sense to collect same metric with different dimension sets at once (collect all dimensions in the same set instead, and aggregate it in required ways later). And I'm not sure, that's even possible. So, I'm not counting this case as worth handling.
- In case when settings were changed, we have two or more row series like:
```
config1 ->
  time1, dimension1, dimension2, value1
  time2, dimension1, dimension2, value2
config2 ->
  time3, dimension1,             value3
  time4, dimension1,             value4
  time5, dimension1,             value5
```
we just need to concatenate them together.

AlexanderZagaynov: - For Paging File(s) situation is bit more complicated: this counter provided for each file separately. Here we have two edge cases:
  - There is only one Paging File (which is most common scenario I believe), within 1 minute interval we have 2 or more values - we just take average of them.
  - There is more than one Paging File - to count common average we need to know each file's size, which is complicated within existing code design. Because this should be really rare case, I just left a note about that: https://github.com/ManageIQ/manageiq-providers-amazon/blob/9162f0482a9a5461d2784d61bdd0375cfbddc87b/app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb#L33
I have thoughts about refactoring the whole metrics collection code, at least for AWS and Azure, which should make it more flexible and effective, and solve such issues too, but right now I want to be simple.

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
AlexanderZagaynov: Result of `perf_collect_metrics` looks like:
```
[
  { target.ems_ref => VIM_STYLE_COUNTERS },
  { 
    target.ems_ref => {
      timestamp => {
        counter_name => value,
        ...
      },
      timestamp... => { ... }
    }
  }
]
```
In the result we need to get: 
1. second (`== last`) hash from array, 
2. first (`== single`) key-value pair in that hash, 
3. and last element of that pair (`== value`).

Which will give us hash of `timestamp => { counter_name => value }` pairs.

I have no better idea to make it more descriptive, than to insert comment with that schema above.
Maybe you can propose something else?

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
AlexanderZagaynov: Exactly. Need to see values from one of the data points in the timestamped series.

app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
AlexanderZagaynov: Yes, just refactoring, I took repeated steps out of the cycle below.

app/models/manageiq/providers/amazon/cloud_manager/metrics_capture.rb
AlexanderZagaynov: done

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
agrare: For readability I would recommend the following:

```ruby
let(:metrics) do
  # This is more obvious what is being returned than just seeing `.last`
  _vim_counters, amazon_counters = vm.perf_collect_metrics('realtime')
  amazon_counters
end
```

Then you can use the vm to dereference the hash instead of `.first.last`
```ruby
timestamps = metrics[vm.ems_ref] # This gets you the timestamps
```

This way it is more obvious what the structure of the data is, trust me if you just see `get_metrics.last.first.last` it is not obvious what is going on.

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
agrare: >> So this ends up being vm.perf_collect_metrics('realtime').last.first.last.first.last?
>
> Exactly.

That's even more confusing :smile:

So assuming we start with something like above where we have a `let(:metrics)` then here we'd have
```
>> timestamps = metrics[vm.ems_ref].keys
=> [timestamp1, timestamp2]
>> counters = metrics[vm.ems_ref][timestamps.first].keys
=> [counter1, counter2]
>> value = metrics[vm.ems_ref][timestamps.first][counters.first]
```

Giving the intermediate values names makes it easier for the reader to see what is going on.

agrare: Or you could do what you did [down there](https://github.com/ManageIQ/manageiq-providers-amazon/pull/522/files#diff-ffbdff4705cd8ab74bed2c08ee3d8d9cR75) and just use the timestamp value since its from a vcr..

simaishi:
Hammer backport details:

```
$ git log -1
commit 17f10144799b8e0fea19eb7c473c7f222f04db72
Date:   Mon Apr 1 17:57:59 2019 -0400

    Merge pull request #522 from AlexanderZagaynov/BZ-1684525_new_metrics_agent
    
    collect metrics from CloudWatch Agent
    
    (cherry picked from commit c5118c3f27eeed4af327da1cdb3fd1322a87d779)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1695628
```

simaishi:
@AlexanderZagaynov Can this be `gaprindashvili/yes`?

simaishi:

```
$ git log -1
commit af340e15606ea9471a89e877ea501ab05723b7af
Date:   Mon Apr 1 17:57:59 2019 -0400

    Merge pull request #522 from AlexanderZagaynov/BZ-1684525_new_metrics_agent
    
    collect metrics from CloudWatch Agent
    
    (cherry picked from commit c5118c3f27eeed4af327da1cdb3fd1322a87d779)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1695655
```

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
AlexanderZagaynov: changed there: https://github.com/ManageIQ/manageiq-providers-amazon/pull/525

spec/models/manageiq/providers/amazon/cloud_manager/metrics_capture_spec.rb
AlexanderZagaynov: changed there: https://github.com/ManageIQ/manageiq-providers-amazon/pull/525

PR #529, opened 2019-04-15
Return not set timeout in dialog as nil instead of 0.

lfu:
To pass the validation on timeoutInMinutes that the value must be greater than or equal to 1.

https://bugzilla.redhat.com/show_bug.cgi?id=1698439

@miq-bot add_label bug
cc @gmcculloug

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/orchestration_service_option_converter.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/orchestration_service_option_converter.rb
@@ -3,13 +3,13 @@ class ManageIQ::Providers::Amazon::CloudManager::OrchestrationServiceOptionConve
   private_constant :REGEX_TAGS, :REGEX_TAGS
 
   def stack_create_options
-    timeout = @dialog_options['dialog_stack_timeout']
+    timeout = @dialog_options['dialog_stack_timeout'].to_i
     policy_body, policy_url = parse_policy(@dialog_options['dialog_stack_policy'])
 
     stack_options = {
       :parameters         => stack_parameters,
       :on_failure         => @dialog_options['dialog_stack_onfailure'],
-      :timeout_in_minutes => timeout.blank? ? nil : timeout.to_i,
+      :timeout_in_minutes => timeout.positive? ? timeout : nil,
       :notification_arns  => parse_multiple_lines(@dialog_options['dialog_stack_notifications']),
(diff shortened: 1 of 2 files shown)


gmcculloug:
cc @bzwei

bzwei:
I believe we are handling the value when there is no UI input in the dialog. What does `@dialog_options['dialog_stack_timeout']` return - `""`, `nil`, `"null"`, `0` or `"0"`?

gmcculloug:
@bzwei The field would not exist in the hash and therefore would be `nil`.  In this case we have an empty text field that is marked as integer type and is coming through a `0`.

app/models/manageiq/providers/amazon/cloud_manager/orchestration_service_option_converter.rb
```diff
@@ -3,13 +3,13 @@ class ManageIQ::Providers::Amazon::CloudManager::OrchestrationServiceOptionConve
...
   def stack_create_options
-    timeout = @dialog_options['dialog_stack_timeout']
+    timeout = @dialog_options['dialog_stack_timeout'].to_i
     policy_body, policy_url = parse_policy(@dialog_options['dialog_stack_policy'])
 
     stack_options = {
       :parameters         => stack_parameters,
       :on_failure         => @dialog_options['dialog_stack_onfailure'],
-      :timeout_in_minutes => timeout.blank? ? nil : timeout.to_i,
+      :timeout_in_minutes => timeout.zero? ? nil : timeout,
```
bzwei: should we have `timeout > 0 ? nil :timeout` to be safer?

gmcculloug: Yes, but I think it would be: `timeout > 0 ? timeout : nil`

simaishi:
Hammer backport details:

```
$ git log -1
commit 98c27003a85bfb16dc8aff44a8cb66765bfd6cae
Author: Greg McCullough <gmccullo@redhat.com>
Date:   Mon Apr 15 16:40:25 2019 -0400

    Merge pull request #529 from lfu/dialog_timeout_1698439
    
    Return not set timeout in dialog as nil instead of 0.
    
    (cherry picked from commit 64bbea94a457dfc3769281af37d72ef2a88cc097)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1702070
```

PR #532, opened 2019-05-09
new regions and instance types

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1705565
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1703474

### DIFF
--- a/app/models/manageiq/providers/amazon/regions.rb
+++ b/app/models/manageiq/providers/amazon/regions.rb
@@ -11,6 +11,18 @@ module ManageIQ
         ManageIQ::Providers::Amazon::Engine.root.join('db/fixtures/aws_regions.yml')
       ).each_value(&:freeze).freeze
 
+      # https://github.com/aws/aws-sdk-ruby/blob/5fe5795e8910bb667996dfc75e4f16b7e69e3980/gems/aws-partitions/partitions.json#L11
+      ORDINARY_REGIONS_REGEXP = /^(us|eu|ap|sa|ca)\-\w+\-\d+$/
+
+      # https://docs.aws.amazon.com/general/latest/gr/rande.html - see quotes below
+      SPECIAL_REGIONS = begin
+        atypical_regions = [
+          'ap-northeast-3', # "To request access to the Asia Pacific (Osaka-Local) Region, contact..."
+          'ap-east-1',      # "you must manually enable before you can use..."
+        ].freeze
+        REGIONS.keys.select { |name| atypical_regions.include?(name) || name !~ ORDINARY_REGIONS_REGEXP }
+      end.freeze
+
       def self.regions
         additional_regions = Hash(Settings.ems.ems_amazon.try!(:additional_regions)).stringify_keys
         disabled_regions   = Array(Settings.ems.ems_amazon.try!(:disabled_regions))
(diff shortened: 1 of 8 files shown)


AlexanderZagaynov:
I know :) I'm in the middle of fetching more instance types currently.

agrare:
@AlexanderZagaynov @Ladas don't want to disrupt this PR because this is just a normal update, but would like to investigate using the static regions definition in the aws gem instead of keeping our own copy.

Caching `Aws::Partitions.defaults["partitions"].flat_map { |p| p["regions"].keys }` could be a suitible replacement for `YAML.load_file(ManageIQ::Providers::Amazon::Engine.root.join('db/fixtures/aws_regions.yml')).each_value(&:freeze).freeze` and we wouldn't need to update our own copy of a yaml file

AlexanderZagaynov:
@agrare no, it will be depending on the bundled gem version.
And I'm going to use API queries after update to `v3` gem.

agrare:
> no, it will be depending on the bundled gem version (e.g. can be outdated).

Exactly, when we want to update the regions we would just bump the gem version

> And I'm going to use API queries after update to v3 gem.

Even better

AlexanderZagaynov:
> Exactly, when we want to update the regions we would just bump the gem version

currently we use old `v2` version, and it can be theoretically discontinued
P.S. I've checked right now, and your example returns only ordinary regions.

> > And I'm going to use API queries after update to v3 gem.
> Even better

so, that's the solution :)

agrare:
It looks like that work might have stalled?  I don't mind continuing to do this in the interim but the interim is getting pretty long :)

AlexanderZagaynov:
> It looks like that work might have stalled? I don't mind continuing to do this in the interim but the interim is getting pretty long :)

that's another question )))

agrare:
Looking forward to it :+1:

Ladas:
Yeah going forward, we should use flavors info from the API, like we do in the Topological Inventory https://github.com/Ladas/topological_inventory-collector-amazon/blob/f28f7a572cffbe736bf21967f9ea9e6971947560/lib/topological_inventory/amazon/collector/pricing.rb#L21

And regions can be fetched from this API call https://github.com/ManageIQ/topological_inventory-amazon/blob/8912f2954e666222edcc88899e68a703961e312b/lib/topological_inventory/amazon/collector.rb#L69

So lets talk to Loic and just do this :-)

AlexanderZagaynov:
@miq-bot add_label hammer/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit 7ff93bc57f16a12310884ca57f213cc66fdd284e
Date:   Thu May 9 21:44:41 2019 -0400

    Merge pull request #532 from AlexanderZagaynov/BZ-1705565_new_regions_and_instance_types
    
    new regions and instance types
    
    (cherry picked from commit 9d1567e0816e583493924a13b4419b109d23d3a9)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1703474
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1728708
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1703461
```

PR #535, opened 2019-05-14
undisable all regions

AlexanderZagaynov:
From @Fryguy:
> I don't think we disable anything upstream, so we can let the community try it.

From @simaishi:
> We'll be enabling all regions for upstream and disabling untested regions for downstream.


agrare:
Hey @AlexanderZagaynov what is the rationale behind removing these?

AlexanderZagaynov:
@agrare I've forward you email discussion :)

agrare:
@AlexanderZagaynov okay can you put that in the description?  No reason why the rationale needs to be secret

PR #548, opened 2019-07-25
Cleanup legacy refresh parsers and settings

agrare:
I'm not aware of anytime that we've had to disable graph refresh or targeted refresh to resolve an issue, and keeping both code bases around is more likely to cause issues with inconsistencies.

Remove the config to select graph or legacy refresh, only allow graph refresh, delete old code.


agrare:
@Ladas I don't see settings for `allow_targeted_refresh` in the Network or Storage Managers, do they not support it or is it not necessary to have that setting on child managers?

Ladas:
@agrare I believe that option isn't necessary, because there is 1 targeted refresh under cloud that refreshes everything.

agrare:
@Ladas perfect, taking out of WIP then

PR #551, opened 2019-08-08
Add JSON-Schema create-params and a common Validate Credentials API

agrare:
Add a set of parameters required by the UI to drive their data-driven-forms and implement the common validate_credentials API instead of raw_connect.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -123,10 +123,10 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
     {
       :title  => "Configure AWS",
       :fields => [
-        {:component => "text-field", :name => "role",                   :type => "hidden", :initialValue => "ec2"},
-        {:component => "text-field", :name => "region",                 :label => "Region"},
-        {:component => "text-field", :name => "endpoints.ec2.username", :label => "Access Key"},
-        {:component => "text-field", :name => "endpoints.ec2.password", :label => "Secret Key", :type => "password"}
+        {:component => "text-field", :name => "role",              :type => "hidden", :initialValue => "ec2"},
+        {:component => "text-field", :name => "region",            :label => "Region"},
+        {:component => "text-field", :name => "access_key",        :label => "Access Key"},
+        {:component => "text-field", :name => "secret_access_key", :label => "Secret Key", :type => "password"}
       ]
     }
   end
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -124,12 +124,6 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
       :title  => "Configure AWS",
       :fields => [
         {
-          :component    => "text-field",
-          :name         => "role",
-          :type         => "hidden",
-          :initialValue => "ec2"
-        },
-        {
           :component  => "text-field",
           :name       => "region",
           :label      => "Region",
@@ -138,14 +132,14 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
         },
         {
           :component  => "text-field",
-          :name       => "access_key",
+          :name       => "endpoints.default.access_key",
           :label      => "Access Key",
           :isRequired => true,
           :validate   => [{:type => "required-validator"}]
         },
         {
           :component  => "text-field",
-          :name       => "secret_access_key",
+          :name       => "endpoints.default.secret_access_key",
           :label      => "Secret Key",
           :type       => "password",
           :isRequired => true,
@@ -153,12 +147,12 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
         },
         {
           :component => "text-field",
-          :name      => "proxy_uri",
+          :name      => "endpoints.default.proxy_uri",
           :label     => "Proxy URI"
         },
         {
           :component => "text-field",
-          :name      => "assume_role",
+          :name      => "endpoints.default.assume_role",
           :label     => "Assume Role"
         }
       ]
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -119,6 +119,18 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
     Settings.ems.ems_amazon.blacklisted_event_names
   end
 
+  def self.params_for_create
+    {
+      :title  => "Configure AWS",
+      :fields => [
+        {:component => "text-field", :name => "role",         :type => "hidden", :initialValue => "ec2"},
+        {:component => "text-field", :name => "region",       :label => "Region"},
+        {:component => "text-field", :name => "ec2_username", :label => "Access Key"},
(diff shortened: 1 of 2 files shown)


agrare:
cc @skateman

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -119,6 +119,24 @@ def self.default_blacklisted_event_names
     Settings.ems.ems_amazon.blacklisted_event_names
   end
 
+  def self.provider_create_params
+    {
+      self.name => params_for_create
+    }
+  end
```
skateman: Ok, now I get it, I don't think we actually need this. I can simply work with the one below. Or is there any special reason for this format?

agrare: Nope I was just trying to maintain the same format that would be returned by calling it on the base class.

skateman: Nah, it might make sense to have it on the `ExtManagementSystem` but not here.

agrare: :+1:

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -51,6 +51,23 @@ module ClassMethods
...
+    #   "endpoints" => {
+    #     "ec2" => {
+    #       "access_key" => "",
+    #       "secret_access_key" => "",
+    #     }
+    #   }
+    # }
+    def validate_credentials(args)
+      region = args["region"]
+      ec2_endpoint = args.dig("endpoints", "ec2")
```
Fryguy: dig could return nil if a layer is missing, but the next line doesn't check for nil before going deeper.

Fryguy:
Liking this approach!

app/models/manageiq/providers/amazon/manager_mixin.rb
agrare: :+1: good call

skateman:
I like it too, but RSpec doesn't :confused: what about @hyperkid123?

Hyperkid123:
The schema looks good. Is there supposed to be some validation?

skateman:
@Hyperkid123 that's coming in a separate PR [here](https://github.com/ManageIQ/manageiq/pull/19120), I'll expose it in an API PR soon.

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -119,6 +119,18 @@ def self.default_blacklisted_event_names
...
 
+  def self.params_for_create
+    {
+      :title  => "Configure AWS",
+      :fields => [
+        {:component => "text-field", :name => "role",         :type => "hidden", :initialValue => "ec2"},
+        {:component => "text-field", :name => "region",       :label => "Region"},
+        {:component => "text-field", :name => "ec2_username", :label => "Access Key"},
+        {:component => "text-field", :name => "ec2_password", :label => "Secret Key", :type => "password"}
+      ]
```
skateman: Having this flat structure might not be enough if you're expecting the data in the validation method in a nested form. Here all the items in the DDF schema are on the same level, but below you wrap everything except of `region` around an `endpoint`.

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -51,6 +51,25 @@ module ClassMethods
...
+    # args:
+    # {
+    #   "region" => "",
+    #   "endpoints" => {
+    #     "ec2" => {
+    #       "access_key" => "",
+    #       "secret_access_key" => "",
+    #     }
+    #   }
+    # }
```
skateman: The current DDF schema won't send the data in this form, but:
```ruby
{
  "region" => "",
  "access_key" => "",
  "secret_access_key" => ""
}
```

martinpovolny:
I do not think that we need the schema to be the DDF schema. Rather I'd have something simple that we can use to generate the final DDF schema with some whistles and bells.

Wdyt, @skateman, @Hyperkid123?

app/models/manageiq/providers/amazon/cloud_manager.rb
rvsia: For nested values you can write something like `endpoints.ec2.username` and DDF will send it as 
```jsx
endpoints: {
   ec2: {
      username: `myname`
 }
}
```

agrare: Okay great, yeah I'm not very familiar with DDF so was looking to you guys for how to get the structure back that we're looking for.

agrare: Okay how does that look?

Hyperkid123: Nesting like this is great. We can go even step further and add `authentication.default.userid` to match the actual API request payload (if there is any). But we can add prefixes easily later on.

Hyperkid123:
@martinpovolny i am not really against using something different but is there as long as we use something explicit. I would like to avoid having something ambiguous (like the mozzila schema).

We should probably consider if this is purely for UI purposes or is there any other reason why would want this? If there is i would consider thinking about something else. Otherwise i don't see a reason why not to use ddf.

app/models/manageiq/providers/amazon/cloud_manager.rb
agrare: Okay there's only one endpoint so we can just flatten out the structure, good to know how to nest under endpoints for providers like ovirt though :+1:

agrare:
@Hyperkid123 how to do optional parameters in DDF?  I'd like to add proxy-uri and assume-role but they're not required

Hyperkid123:
Do you mean optional in a way that the fields are there but they are not required or if someone chooses some option from select for instance, then another field appears?

If you mean the first case: By default all fields are option (as in they are not required). If you wan to add required you must add validator and additional properties for the component to add visual representation (for required the *).

example: 
```javascript
{
  component: 'text-field',
  name: 'field',
  label: 'Validated field!',
  isRequired: true,
  validate: [
    {
      type: 'required-validator',
    }
  ]
}
```
http://data-driven-forms.surge.sh/renderer/validators#requiredvalidator

agrare:
Oh okay awesome thanks @Hyperkid123

Hyperkid123:
@agrare yes. Currently we have just a limited amount of data types but we can add some if we need: http://data-driven-forms.surge.sh/renderer/data-types

What data-types are we talking about? Or do you want to just check the if the string is URL for instance? For that we have a number of different validators we can use.

Hyperkid123:
@agrare All values are strings bu default. There is no data-type URL in JS, but we can add url validator: http://data-driven-forms.surge.sh/renderer/validators#urlvalidators

There is a lot of params for URL validator like port, host, if ip address is valid, etc. Let me know what url format you want and we can add it.

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -119,6 +119,46 @@ def self.default_blacklisted_event_names
...
+          :name      => "endpoints.default.proxy_uri",
+          :label     => "Proxy URI"
+        },
+        {
+          :component => "text-field",
+          :name      => "endpoints.default.assume_role",
+          :label     => "Assume Role"
+        }
+      ]
+    }
```
Fryguy: Minor, but I feel this should be a constant, which after that I don't have a preference if it's exposed as a method as well and/or just the constant.

Fryguy:
Regarding the validation it would be really sweet if we had a DFF -> Hash validation class on the Ruby side.  Unfortunately, you'd need a Ruby implementation of each validator, but this would get us server-side validation (to help with direct-api calls) in addition to the client-side validation.

PR #553, opened 2019-08-23
We no longer disable regions in the settings

juliancheal:
This rake task now just updates the regions yaml file.


juliancheal:
@miq-bot assign @agrare

PR #554, opened 2019-09-03
add metal to names and remove dedicated

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1724686
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1724631

PR #556, opened 2019-09-04
[IVANCHUK] fix types count

AlexanderZagaynov:
fix for https://github.com/ManageIQ/manageiq-providers-amazon/pull/554#issuecomment-527903227


AlexanderZagaynov:
@agrare do you know what was changed? I see that there is three possible values in this branch.

agrare:
Hey @AlexanderZagaynov I took care of this over in https://github.com/ManageIQ/manageiq-providers-amazon/pull/557
We were adding a constant value to the number of flavors in some cases but just checking if graph refresh is used is better.

AlexanderZagaynov:
@agrare got it, thanks

PR #565, opened 2019-10-02
Fix rspec expect taking a lambda as an argument

agrare:
Rspec is not allowing expect to take a lambda as an argument anymore


djberg96:
Pretty sure I read that rspec 4 will mandate that the outer `describe` be`RSpec.describe`. Any chance of updating that while we're here?

Fryguy:
@djberg96 It's unrelated to this change, so should probably be a different PR.  In fact, we can probably pull_request_blaster_outer that change.

simaishi:
Travis is failing in ivanchuk branch as well, adding `ivanchuk/yes`

PR #570, opened 2019-10-25
Optionally symlink a core repo in bin/setup

agrare:
If a MANAGEIQ_REPO env var is passed to bin/setup then symlink that repo
to spec/manageiq instead of cloning a new copy.  This makes it possible
to test cross-repo changes with a plugin.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,9 +1,17 @@
 #!/usr/bin/env ruby
 require 'pathname'
+require 'fileutils'
 
 gem_root = Pathname.new(__dir__).join("..")
+spec_manageiq = gem_root.join("spec/manageiq")
 
-unless gem_root.join("spec/manageiq").exist?
+if ENV.key?("MANAGEIQ_REPO")
+  manageiq_repo = Pathname.new(ENV["MANAGEIQ_REPO"])
+  puts "== Symlinking spec/manageiq to #{manageiq_repo}"
+
+  FileUtils.rm_rf(spec_manageiq.expand_path)
+  FileUtils.ln_s(manageiq_repo.expand_path, spec_manageiq.expand_path)
+elsif !spec_manageiq.exist?
   puts "== Cloning manageiq sample app =="
   system "git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 spec/manageiq"
 end


bin/setup
```diff
@@ -4,8 +4,13 @@ require 'pathname'
 gem_root = Pathname.new(__dir__).join("..")
 
 unless gem_root.join("spec/manageiq").exist?
-  puts "== Cloning manageiq sample app =="
-  system "git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 spec/manageiq"
+  if ENV.key?("MANAGEIQ_REPO")
+    puts "== Symlinking #{ENV["MANAGEIQ_REPO"]} to spec/manageiq"
+    system "ln -s #{ENV["MANAGEIQ_REPO"]} spec/manageiq"
```
Fryguy: If the symlink already exists, then what happens?  Wondering if this should force the symlink if it already exists.  Then you don't need the "cleanup" line in the manageiq-cross_repo.

bin/setup
```diff
@@ -4,8 +4,13 @@ require 'pathname'
 gem_root = Pathname.new(__dir__).join("..")
 
 unless gem_root.join("spec/manageiq").exist?
-  puts "== Cloning manageiq sample app =="
-  system "git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 spec/manageiq"
+  if ENV.key?("MANAGEIQ_REPO")
+    puts "== Symlinking #{ENV["MANAGEIQ_REPO"]} to spec/manageiq"
```
Fryguy: I think the wording here is backwards... Should it be 

puts "== Symlinking spec/manageiq to #{ENV["MANAGEIQ_REPO"]}"

bin/setup
agrare: > If the symlink already exists, then what happens? 

If the symlink already exists then this wouldn't touch it because of the `unless gem_root.join("spec/manageiq").exist?`

> Wondering if this should force the symlink if it already exists. Then you don't need the "cleanup" line in the manageiq-cross_repo.

Hm, do you mean if an ENV["MANAGEIQ_REPO"] is passed we should blow away what is currently at `spec/manageiq`, or in all cases blow it away?

I'd rather leave the "normal developer" mode of operations here the same, otherwise we'd be `git clone`'ing a lot so if it exists already then just leave it.

We could overwrite whatever is there if `ENV["MANAGEIQ_REPO"]` is passed, that's kind of an implicit "I'm being called by cross_repo-tests" though since I doubt anyone would call `bin/setup` with that env var, they'd just symlink it once and leave it.

bin/setup
agrare: You're right, fixed

bin/setup
Fryguy: > Hm, do you mean if an ENV["MANAGEIQ_REPO"] is passed we should blow away what is currently at spec/manageiq, or in all cases blow it away?

If `ENV["MANAGEIQ_REPO"]`, then I think it replaces the current symlink, else we do what it did before...so something like...

```ruby
if ENV.key?("MANAGEIQ_REPO")
  # ln -sf
elsif !gem_root.join("spec/manageiq").exist?
  # git clone
end
```

agrare: :+1: works for me, I'll update the core generator PR with this

PR #576, opened 2019-11-14
Rely on the default implementation of .ems_class

carbonin:
Requires https://github.com/ManageIQ/manageiq/pull/19516


carbonin:
Trying to figure out what to do with the `ManageIQ::Providers::Amazon::AgentCoordinatorWorker`.

Looks like this and `MiqEmsRefreshCoreWorker` are the only exceptions to the `parent` implementation.

carbonin:
Okay, I think I got it working. Thoughts @Fryguy @agrare @hsong-rh?

PR #579, opened 2019-11-19
Add worker kill priority to the AgentCoordinatorWorker

carbonin:
This is needed in order to seed the MiqWorkerType model


carbonin:
Adding WIP as we debate how priority is handled in ManageIQ/manageiq#19536

PR #580, opened 2019-12-12
Add a method to create an AWS EMS from DDF params

agrare:


app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -152,13 +166,26 @@ def self.params_for_create
...
           :component => "text-field",
-          :name      => "endpoints.default.assume_role",
+          :name      => "endpoints.default.service_account",
           :label     => "Assume Role"
         }
       ]
     }.freeze
   end
 
+  def self.create_from_params(params)
```
skateman: Would be nice if this method would exist on the top-level `Provider` class and it would be able to select the right one based on the `type` attribute of the passed `params`. This way we would not need any provider-specific logic in the API.

skateman: But that's not in the context of this PR

agrare: Yeah I'm playing around with maybe this can be completely generic if we setup the params properly, or at least do the bulk of it generically and the providers only override anything weird that they do

skateman:
I created an API PR that exposes this call under certain conditions. Sent it further for MM for testing.

skateman:
@hyperkid123 is getting a LoadErrot with this PR when calling an `OPTIONS` request on the `/api/providers`:
```
LoadError (Unable to autoload constant ManageIQ::Providers::Amazon::StorageManager::Ebs, expected /home/mmarosi/manageiq/manageiq-providers-amazon/app/models/manageiq/providers/amazon/storage_manager/ebs.rb to define it):

```

agrare:
@skateman it is defined there though isn't it?

```
$ grep 'ManageIQ::Providers::Amazon::StorageManager::Ebs' app/models/manageiq/providers/amazon/storage_manager/ebs.rb
class ManageIQ::Providers::Amazon::StorageManager::Ebs < ManageIQ::Providers::StorageManager
```

Do you get the same error with a rails console?

skateman:
No, but I can't even reproduce the problem with CURL. @hyperkid123, can you give us more of the context? The request URL and the exact params maybe?

Hyperkid123:
here is the request:
```
Request URL: http://localhost:3000/api/providers
Request Method: POST
Status Code: 400 Bad Request
Remote Address: [::1]:3000
Referrer Policy: no-referrer-when-downgrade
```
Request headers (removed cookie and token)
```
accept: */*
Accept-Encoding: gzip, deflate, br
accept-language: en-US,en;q=0.9,cs;q=0.8
Connection: keep-alive
Content-Length: 197
content-type: text/plain;charset=UTF-8
DNT: 1
Host: localhost:3000
Origin: http://localhost:3000
Referer: http://localhost:3000/ems_cloud/new
Sec-Fetch-Mode: cors
Sec-Fetch-Site: same-origin
User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.88 Safari/537.36
```

Payload:
```js
{
  zone: "default",
  name: "Foo",
  type: "ManageIQ::Providers::Amazon::CloudManager",
  provider_region: "Foo",
  endpoints: {default: {userid: "Foo", password: "Foo"}},
  default: {userid: "Foo", password: "Foo"},
  zone_name: "default",
  ddf: true
}

```

skateman:
That's fine, but I'm curious about the OPTIONS request you mentioned.

Hyperkid123:
> That's fine, but I'm curious about the OPTIONS request you mentioned.

So the options request was throwing the autoload issue. It did that twice yesterday but it is no happening right now.

skateman:
Also reproduced the problem randomly with a forced code reload in the area:
```
LoadError (Unable to autoload constant ManageIQ::Providers::Amazon::StorageManager::Ebs, expected /home/skateman/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/manageiq-providers-amazon-120d359a407a/app/models/manageiq/providers/amazon/storage_manager/ebs.rb to define it):
[----] F, [2020-01-14T11:09:57.437535 #37521:312bd3c] FATAL -- :   
[----] F, [2020-01-14T11:09:57.437611 #37521:312bd3c] FATAL -- : activesupport (5.1.7) lib/active_support/dependencies.rb:512:in `load_missing_constant'
activesupport (5.1.7) lib/active_support/dependencies.rb:202:in `const_missing'
activesupport (5.1.7) lib/active_support/inflector/methods.rb:271:in `const_get'
activesupport (5.1.7) lib/active_support/inflector/methods.rb:271:in `block in constantize'
activesupport (5.1.7) lib/active_support/inflector/methods.rb:267:in `each'
activesupport (5.1.7) lib/active_support/inflector/methods.rb:267:in `inject'
activesupport (5.1.7) lib/active_support/inflector/methods.rb:267:in `constantize'
activesupport (5.1.7) lib/active_support/inflector/methods.rb:312:in `safe_constantize'
activesupport (5.1.7) lib/active_support/core_ext/string/inflections.rb:77:in `safe_constantize'

```

Hyperkid123:
@agrare @skateman submit endpoint is working. We will now work with @skateman to create the correct UI schema.

skateman:
@agrare we're okay with the changes, just

skateman:
@agrare we're okay with the changes, just the schema needs a little revisit to look as it looks in the current implementation, but I'll do that. Do you need a *Seal of Approval&trade;* right now, or would you like to also implement the `edit` feature?

agrare:
@skateman if you want to update the schema lets get this in and I'll follow up with an edit method

skateman:
@agrare @Fryguy can we get this in?

PR #582, opened 2019-12-19
[WIP] Collect instance types from the API

agrare:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1781155


app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
```diff
@@ -4,7 +4,7 @@ def instances
   end
 
   def flavors
-    ManageIQ::Providers::Amazon::InstanceTypes.all
+    hash_collection.new(aws_ec2.client.describe_instance_types[:instance_types])
```
djberg96: Any filters we should apply here? Do we want to cap the results? Just checking:

https://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html#describe_instance_types-instance_method

agrare: No need to cap the results, but we do need to handle paging generally.  That is out of scope of this PR since that is a general issue with all of these API calls.  I opened an issue to track that, https://github.com/ManageIQ/manageiq-providers-amazon/issues/583

djberg96: So then it just becomes `hash_collection.new(aws_ec2.client.describe_instance_types.flat_map(&:instance_types))`

agrare:
Superseded by https://github.com/ManageIQ/manageiq-providers-amazon/pull/614

PR #586, opened 2020-02-04
Make the DDF fields more similar to the current form's layout + i18n

skateman:
This is a followup based on discussions with @hyperkid123 and earlier in Mahwah with @Fryguy and @agrare. I removed the `name` and `zone` fields as they will be provided by the add provider form in order to decrease redundancy among forms. We can still reverse out from this solution, but in any case, the ordering of fields should and will be a little different. Common stuff has to go first, we can't implement jumping inconsistencies in DDF nicely.

I also added i18n calls on the strings, so @mzazrivec can consume them.

**Old form:**
Screenshot from 2020-02-04 20-39-59

**New form:**
Screenshot from 2020-02-05 20-47-16

Note that the bottom screenshot doesn't show the `name`, `type` and `zone` fields in the correct order, because I'm using the schema from https://github.com/ManageIQ/manageiq-ui-classic/pull/6402.

There are still stuff missing:
* [x] Validation button
* [x] ~~Dynamically~~ loading the values for the region
* [x] The endpoint URL field's name
* [x] The proxy field that was in the DDF schema but not in the old implementation

@miq-bot add_reviewer @agrare 
@miq-bot add_reviewer @Fryguy

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -122,55 +122,61 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
 
   def self.params_for_create
     @params_for_create ||= {
-      :title  => "Configure AWS",
       :fields => [
         {
-          :component  => "text-field",
-          :name       => "name",
-          :label      => "Name",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
-        },
-        {
-          :component  => "text-field",
-          :name       => "zone_name",
-          :label      => "Zone",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
-        },
-        {
-          :component  => "text-field",
+          :component  => "select-field",
           :name       => "provider_region",
-          :label      => "Region",
+          :label      => _("Region"),
           :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
+          :validate   => [{:type => "required-validator"}],
+          :options    => ManageIQ::Providers::Amazon::Regions.all.sort_by { |r| r[:description] }.map do |region|
+            {
+              :label => region[:description],
+              :value => region[:name]
+            }
+          end
         },
         {
-          :component  => "text-field",
-          :name       => "endpoints.default.userid",
-          :label      => "Access Key",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
+          :component => 'sub-form',
+          :name      => 'endpoints',
+          :title     => _("Endpoint"),
+          :fields    => [
+            {
+              :component              => 'validate-provider-credentials',
+              :name                   => 'aws-endpoint-valid',
+              :validationDependencies => %w[zone_name provider_region],
+              :fields                 => [
+                {
+                  :component => "text-field",
+                  :name      => "endpoints.default.url",
+                  :label     => _("Endpoint URL"),
+                },
+                {
+                  :component => "text-field",
+                  :name      => "endpoints.default.service_account",
+                  :label     => _("Assume role ARN"),
+                },
+                {
+                  :component  => "text-field",
+                  :name       => "endpoints.default.userid",
+                  :label      => _("Access Key ID"),
+                  :helperText => _("Should have privileged access, such as root or administrator."),
+                  :isRequired => true,
+                  :validate   => [{:type => "required-validator"}]
+                },
+                {
+                  :component  => "text-field",
+                  :name       => "endpoints.default.password",
+                  :label      => _("Secret Access Key"),
+                  :type       => "password",
+                  :isRequired => true,
+                  :validate   => [{:type => "required-validator"}]
+                },
+              ],
+            },
+          ],
         },
-        {
-          :component  => "text-field",
-          :name       => "endpoints.default.password",
-          :label      => "Secret Key",
-          :type       => "password",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
-        },
-        {
-          :component => "text-field",
-          :name      => "endpoints.default.proxy_uri",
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -122,55 +122,48 @@ def self.default_blacklisted_event_names
...
-          :label      => "Access Key",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
+          :component => 'sub-form',
+          :name      => 'endpoints',
+          :title     => _("Endpoints"),
+          :fields    => [
+            {
+              :component => "text-field",
+              :name      => "<TBD>", # TODO: maybe proxy uri? (default_url in the old world)
```
skateman: This is a big big :question: for me @agrare :confused:

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -122,55 +122,48 @@ def self.default_blacklisted_event_names
...
-          :name       => "endpoints.default.password",
-          :label      => "Secret Key",
-          :type       => "password",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
-        },
-        {
-          :component => "text-field",
-          :name      => "endpoints.default.proxy_uri",
-          :label     => "Proxy URI"
```
skateman: Do we need this, is this the same as the `default_url` from the right? @agrare

app/models/manageiq/providers/amazon/cloud_manager.rb
skateman: This is the stuff I was able to find, it is being sent as an argument to `validate_credentials_task`:
```ruby
          uri = URI.parse(WEBrick::HTTPUtils.escape(params[:default_url]))
          [
            user,
            password,
            :EC2,
            params[:provider_region],
            ems.http_proxy_uri,
            true,
            uri,
            {:assume_role => params[:default_assume_role].presence},
          ]
```

So I guess it becomes the 7th argument of [`raw_connect`](https://github.com/ManageIQ/manageiq-providers-amazon/blob/01feee535cc84e6da97590ae1b31145416ea8815/app/models/manageiq/providers/amazon/manager_mixin.rb#L79) which is in the provider.

agrare: Okay so that's if you have a different endpoint from the normal one e.g. gov cloud?  I'm wondering how that's different from using a proxy

skateman: I don't even know what gov cloud is :sweat_smile: so good question.

agrare: Well if we support it now we shouldn't change that, its just weird :)

skateman: Ok, how should I name the field? So your validation/add endpoints can handle it.

agrare: It should just be called `"url"` since that's what it is called on the endpoint table

agrare: Okay more confused, we don't pass that in when we use the instance method [`#connect`](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/manager_mixin.rb#L20-L33) so even if it validated with it, it would never be used (?!?) (:scream:)

app/models/manageiq/providers/amazon/cloud_manager.rb
agrare: I believe that this is different from the default_url, we need to figure out when you add an amazon provider from the UI today with these fields filled in, where in the database do those fields go?

app/models/manageiq/providers/amazon/cloud_manager.rb
skateman: `endpoints.default.url`

app/models/manageiq/providers/amazon/cloud_manager.rb
skateman: As the old UI doesn't use this, I'm just omitting it for now.

Hyperkid123:
@skateman there is a component for async credentials validation prepared in mamage IQ. We can take a look in the office. We might have to adjust it a bit to enable string based definition of the validation action but that should be ok

skateman:
I went through the dynamic fields that could be possibly loaded through the API and the only one is the `region` in a few cloud providers. I don't think it's worth it, we can just statically pass the region data into the schema.

agrare:
Travis is actually green its just not updated here, @skateman just remove WIP when you want to have this merged

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -122,55 +122,53 @@ def self.default_blacklisted_event_names
...
-        },
-        {
-          :component  => "text-field",
-          :name       => "endpoints.default.userid",
-          :label      => "Access Key",
+          :label      => _("Region"),
           :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
+          :validate   => [{:type => "required-validator"}],
+          :options    => ManageIQ::Providers::Amazon::Regions.all.sort_by { |r| r[:description] }.map do |region|
```
agrare: :+1: nice

skateman:
@agrare I'm still figuring out the validation part, it might need some additional schema changes.

skateman:
I think this is done, what do you think @Hyperkid123? Can you test this with your PR?

skateman:
I was able to create a provider with this schema! :tada:

Hyperkid123:
@skateman sorry I was on PTO. It makes me really happy that this works!

skateman:
Me too Martin, me too!

Issue #587, opened 2020-02-07
ManagerMixin::ClassMethods::verify_credentials always returns true


skateman:
The [`verify_credentials`](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/manager_mixin.rb#L76) method always returns `true`, even if the credentials are invalid.

If this is fixed, I think it should also return a more complex result containing the validation result and a textual message if an error occured.

@miq-bot assign @agrare 
Parent issue: https://github.com/ManageIQ/manageiq/issues/18818

skateman:
@miq-bot add_label bug

skateman:
@agrare I leave it up to you, basically the only error that can happen is that the credentials are invalid, right? :sweat_smile:

PR #588, opened 2020-02-07
Fix verify_credentials always returning true

agrare:
The verify_credentials method wasn't checking that the AWS connection
was able to execute any commands.

Fixes https://github.com/ManageIQ/manageiq-providers-amazon/issues/587

### DIFF
--- a/app/models/manageiq/providers/amazon/manager_mixin.rb
+++ b/app/models/manageiq/providers/amazon/manager_mixin.rb
@@ -73,7 +73,7 @@ module ManageIQ::Providers::Amazon::ManagerMixin
       )
       secret_access_key = MiqPassword.try_decrypt(secret_access_key)
 
-      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, :assume_role => assume_role)
+      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, validate = true, :assume_role => assume_role)
     end
 
     def raw_connect(access_key_id, secret_access_key, service, region,


app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -73,7 +73,11 @@ def verify_credentials(args)
       )
       secret_access_key = MiqPassword.try_decrypt(secret_access_key)
 
-      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, :assume_role => assume_role)
+      begin
+        !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, validate = true, :assume_role => assume_role)
+      rescue
```
djberg96: `validate = true`? Is it just `true` or is it supposed to be `:validate => true`?

skateman: Nope, as of [here](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/manager_mixin.rb#L80), but maybe the allocation is unnecessary and it can be just `true`.

agrare: Just for readability, seeing just true as an arg doesn't really mean anything without going to the method description and counting args

agrare: This is why I like kwargs for stuff like this

skateman: We might don't want the exception handling here, testing ...

skateman: Yeah, better without the handling, the task fails and the exception message is actually helpful.

agrare: :+1: done

skateman:
Good to go after green @gtanzillo

agrare:
@gtanzillo this is green now, looks like we had a sporadic failure that should be fixed by https://github.com/ManageIQ/manageiq-providers-amazon/pull/590

PR #590, opened 2020-02-07
Require aws-sdk-core in aws spec helper

agrare:
We are going through Aws.config to look for services without requiring anything to pull in Aws (which is defined in aws-sdk-core).


djberg96:
👍

PR #599, opened 2020-02-25
Implement a method that allows editing with incoming DDF params

skateman:
Created analogously with `self.create_from_params`. Related API PR: https://github.com/ManageIQ/manageiq-api/pull/753

@miq-bot assign @agrare


skateman:
@agrare we might need additional sanitizing based on the schema, I'm not sure about that. But if we do it, we need to implement it outside the provider's repo universally and call it here (and maybe for creation too). What do you think?

PR #603, opened 2020-03-03
Remove invalid partials in Amazon specs

djberg96:
This removes a couple invalid partials that causes errors if strict partial validation is enabled. In this case, objects are stubbing methods that they don't actually implement. With these removed, the specs still pass.


djberg96:
@miq-bot add_label test, refactoring

PR #607, opened 2020-03-23
Remove the overridden persister initialize

agrare:
There was a pointless overridden Persister#initialize method and manager_uuids for key_pairs that are already added by the builder


djberg96:
@agrare I see this has a *collector* argument, whereas the base class doesn't. Is it not used? Just double checking.

agrare:
@djberg96 yes it does but that isn't used anywhere in the persister anymore

djberg96:
👍

PR #608, opened 2020-03-23
remove inventory object refresh

d-m-u:
just what it says on the tin

@miq-bot add_label cleanup
@miq-bot assign @agrare

### DIFF
--- a/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
+++ b/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
@@ -12,25 +12,21 @@ module AwsRefresherSpecCommon
 
   ALL_GRAPH_REFRESH_SETTINGS = [
     {
-      :inventory_object_refresh => true,
       :inventory_collections    => {
         :saver_strategy => "default",
       },
     }, {
-      :inventory_object_refresh => true,
       :inventory_collections    => {
         :saver_strategy => "batch",
         :use_ar_object  => true,
       },
     }, {
-      :inventory_object_refresh => true,
       :inventory_collections    => {
         :saver_strategy => "batch",
         :use_ar_object  => false,
       },
     }, {
       :inventory_object_saving_strategy => :recursive,
-      :inventory_object_refresh         => true
     }
   ].freeze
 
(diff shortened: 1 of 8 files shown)


spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -78,11 +74,6 @@ def assert_common
     assert_network_router
     assert_relationship_tree
     assert_specific_labels_on_vm
-
-    if options.inventory_object_refresh
-      assert_specific_service_offerings
-      assert_specific_service_instances
-    end
```
agrare: We probably want to test both of these since `options.inventory_object_refresh` should always be true now right?

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -880,18 +865,7 @@ def assert_specific_vm_on_cloud_network_public_ip
     expect(v.network_ports.first.fixed_ip_addresses).to match_array([@ip2.fixed_ip_address])
     expect(v.network_ports.first.ipaddresses).to match_array([@ip2.fixed_ip_address, @ip2.address])
     expect(v.ipaddresses).to match_array([@ip2.fixed_ip_address, @ip2.address])
-
-    if options.inventory_object_refresh
-      expect(v.operating_system).to(
-        have_attributes(
-          :product_name => "linux_redhat",
-        )
-      )
```
agrare: Same here

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -902,7 +876,7 @@ def assert_specific_vm_on_cloud_network_public_ip
       have_attributes(
         :config_version       => nil,
         :virtual_hw_version   => nil,
-        :guest_os             => options.inventory_object_refresh ? "linux_redhat" : nil,
+        :guest_os             => nil,
```
agrare: This would be `"linux_redhat"` right?  If ` options.inventory_object_refresh` is true

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -880,18 +867,8 @@ def assert_specific_vm_on_cloud_network_public_ip
...
-          :product_name => "linux_redhat",
-        )
-      )
-    else
-      # Old refresh can't fetch the public image and there fore also operating_system
-      expect(v.operating_system).to be_nil
-    end
-
+    expect(v.operating_system).to(have_attributes(:product_name => "linux_redhat"))
+    expect(v.operating_system).to be_nil
```
agrare: I don't think this can be both `{:product_name => "linux_redhat"}` _and_ `nil`...

PR #614, opened 2020-04-02
Use V3 interface to get instance types

djberg96:
Meant to replace https://github.com/ManageIQ/manageiq-providers-amazon/pull/582

This updates the Amazon provider so that it uses the V3 api to collect available instance information instead of using a local, hand edited .yaml file.

Some notes on specific fields:

* name - This is now `instance_type`.
* description - There doesn't appear to be a description field, so I'm just re-using the `instance_type`.
* enabled - As far as I can tell the API will only return enabled images, so I've defaulted this to true.
* cloud_subnet_required - Removed for now, I don't see a way to glean that information currently.
* cpus - Same as before, but takes a little more digging to get to it.
* cpu_cores - Previously this was hard coded at 1. Now we actually get this information from the API.
* memory - Same as before, but takes a little more digging to get to it.
* supports_32_bit - Same as before...
* supports_64_bit - Same as before...
* supports_hvm - There is no field for this, but as per the documentation, all instance types support HVM AMI's, so this was hard coded to true.
* supports_paravirtual - There is no field for this, but we know from the documentation that any current instance type does not support paravirtual, and that only a subset of the non-current instance types do, so I use a helper method to set that.
* block_storage_based_only - There's no specific field for this, but we can check the supported storage types. If there's anything other than 'ebs' then it's false.
* ephemeral_disk_size - Here I'm using the total disk size, keeping in mind that there could be more than one disk.
* ephemeral_disk_count - Here I'm just checking the size of the disk array. TODO: Actually add up the various `count` fields instead.
* publicly_available - Defaults to true since as far as I can tell the API only returns what's publicly available. There is no field for this.
* vpc_only - I'm making a best guess atm based on what information I can find, but the API doesn't provide it.

See https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-types.html for more information, including example output.

Also see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/virtualization_types.html

~~Other comments: I'm not use the `HashCollection` wrapper because I'm not really seeing the purpose of it, so for now I'm just using the results from the V3 API directly.~~

Draft for now until I get the specs worked out, and a couple fields verified.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1781155


Fryguy:
You had said hand edited, but if I recall there a generator for it, no?  I would think some of that information could be pulled from there.

agrare:
@Fryguy I think the generator is only for regions https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/lib/tasks_private/regions.rake#L4 I don't see one for instance_types

Fryguy:
Ok thanks, I could have sworn we had it for instance types 😕

agrare:
Yeah I really thought we did also, thought AlexZ added region and instance generators at the same time.

djberg96:
@Fryguy @agrare I think the last conversation about it was here: https://github.com/ManageIQ/manageiq-providers-amazon/pull/532

Fryguy:
Not sure, but I think some of the removed fields may be used elsewhere, such as in provisioning.

Fryguy:
Also, how are we going to deal with legacy/discontinued types?  Those are important because when Amazon discontinues an instance_type that doesn't mean that existing instances are removed; real instances can return an instance type that may be discontinued, and we need to account for that.

agrare:
> Also, how are we going to deal with legacy/discontinued types

I think we can "build" a legacy flavor based on the instance's instance_type data.  We don't need to have all of the details because we aren't going to be provisioning new instances with these

djberg96:
> Not sure, but I think some of the removed fields may be used elsewhere, such as in provisioning.

The only removed field was `vpc_only`, which I've added back in. However, near as I can tell it doesn't affect provisioning. I saw nothing for that field explicitly in the UI classic code, and fiddling with the provisioning screen in the UI didn't seem to change whether or not the cloud subnet field was mandatory.

@himdel or @ZitaNemeckova, is there any connection between that field and the UI that I'm missing?

I'm having a hard time finding any sort of consistent matrix that provides that information. I've found stuff like https://mindmajix.com/aws-ec2-instance-types that shows some,  but https://docs.amazonaws.cn/en_us/vpc/latest/userguide/how-it-works.html#how-it-works-subnet would lead me to believe that any current flavor is vpc-only, though it's possible I'm misinterpreting article.

I realize we were pulling from https://raw.githubusercontent.com/powdahound/ec2instances.info/master/www/instances.json before, but it's not clear to me where that information is coming from.

So, what I have now is a best guess. If you have a better way please let me know.

himdel:
As far as I can tell, `cloud_subnet_required` is only displayed as a column in the Flavors list, somewhere in the UI,
and as `{:label => _('Cloud Subnet Required'), :value => @record.cloud_subnet_required?}` in the textual summary, but that's a model method (with a question mark).

There is no UI logic based on any of the two. There may be validation logic though.

Fryguy:
Yeah I was thinking logic at the provisioning layer for calls like available_flavors.  @gmcculloug / @tinaafitz might know better.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -317,29 +317,51 @@ def hardware_disks(persister_hardware, instance, flavor)
...
+  end
+
+  # This logic is cobbled together from various online sources. The V3 API does not include
+  # such information, but most current flavors are vpc-only from what I can gather.
+  #
+  def vpc_only?(flavor)
+    if %w[a1 c1 c3 r3].include?(flavor['instance_type'].split('.').first.downcase)
+      return false
+    end
+    true
```
Fryguy, line 358: Think this whole method can just be

```
def vpc_only?(flavor)
   %w[a1 c1 c3 r3].exclude?(flavor['instance_type'].split('.').first.downcase)
end
```

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -317,29 +317,51 @@ def hardware_disks(persister_hardware, instance, flavor)
...
+    return false unless %w[c1 c3 hs1 m1 m3 m2 t1].include?(flavor['instance_type'].split('.').first.downcase)
+
+    true
+  end
+
+  # This logic is cobbled together from various online sources. The V3 API does not include
+  # such information, but most current flavors are vpc-only from what I can gather.
+  #
+  def vpc_only?(flavor)
+    if %w[a1 c1 c3 r3].include?(flavor['instance_type'].split('.').first.downcase)
```
Fryguy, line 355: Minor, but I think this `.split('.').first.downcase` bit should be moved into a private method to make it clearer what it's doing (and also because it's used up above in the supports_paravirtual method as well).  I'm thinking something like:

```ruby
private def instance_family(flavor)
  flavor.split('.').first.downcase
end
```

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -317,29 +317,51 @@ def hardware_disks(persister_hardware, instance, flavor)
...
+        :memory                   => flavor.dig('memory_info', 'size_in_mi_b')&.megabytes,
+        :supports_32_bit          => flavor.dig('processor_info', 'supported_architectures').include?('i386'),
+        :supports_64_bit          => flavor.dig('processor_info', 'supported_architectures').include?('x86_64'),
+        :supports_hvm             => true,
+        :supports_paravirtual     => supports_paravirtual?(flavor),
+        :block_storage_based_only => flavor.dig('supported_root_device_types').any? { |device| device != 'ebs' },
+        :ephemeral_disk_size      => flavor.dig('instance_storage_info', 'total_size_in_gb')&.gigabytes || 0,
+        :ephemeral_disk_count     => flavor.dig('instance_storage_info', 'disks')&.size || 0,
+        :publicly_available       => true,
+        :cloud_subnet_required    => vpc_only?(flavor)
```
Fryguy, line 335: Can you move this back up under block_storage_based_only where it was previously?

Fryguy:
> However, near as I can tell it doesn't affect provisioning.

@djberg96 Did you search for `cloud_subnet_required`? I'm seeing it used in placement automation

```
/Users/jfrey/dev/manageiq-content/content/automate/ManageIQ/Cloud/VM/Provisioning/Placement.class/__methods__/best_fit_amazon.rb:41:                if flavor.try(:cloud_subnet_required)
```

gmcculloug:
@Fryguy I don't see `cloud_subnet_required` referenced in any of the provisioning workflows which would match @djberg96 test results.  As you found it is referenced from the `best_fit_amazon` method in Automate.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
djberg96: Done.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
djberg96: Done.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
djberg96: Done.

AlexanderZagaynov:
Instance types fetching task lives here https://github.com/AlexanderZagaynov/manageiq-providers-amazon/pull/2/commits/67eee8d1762346e08128ef3b4783b11cf629bcd9

djberg96:
@AlexanderZagaynov thanks!

djberg96:
@AlexanderZagaynov Wow, that takes a while. I ultimately ran into this:

```
D, [2020-07-02T11:59:25.831648 #37405] DEBUG -- : Cache generate: /Users/dberger/Dev/manageiq-djberg96/tmp/aws_cache/products_data/AB2/890/AmazonEC2.offers_data.20200129004120
rake aborted!
SocketError: Failed to open TCP connection to pricing.us-east-1.amazonaws.com:443 (getaddrinfo: nodename nor servname provided, or not known)
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:116:in `block (3 levels) in offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:114:in `block (2 levels) in offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:113:in `each'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:113:in `map'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:113:in `block in offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:111:in `offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:61:in `block in parse!'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:60:in `parse!'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:42:in `products_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:38:in `result'
lib/tasks_private/instance_types.rake:131:in `block (3 levels) in <top (required)>'
lib/tasks_private/instance_types.rake:37:in `isolated'
lib/tasks_private/instance_types.rake:124:in `block (2 levels) in <top (required)>'
/Users/dberger/.rbenv/versions/2.6.5/bin/bundle:23:in `load'
/Users/dberger/.rbenv/versions/2.6.5/bin/bundle:23:in `<main>'

Caused by:
SocketError: getaddrinfo: nodename nor servname provided, or not known
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:116:in `block (3 levels) in offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:114:in `block (2 levels) in offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:113:in `each'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:113:in `map'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:113:in `block in offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:111:in `offers_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:61:in `block in parse!'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:60:in `parse!'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:42:in `products_data'
/Users/dberger/Dev/manageiq-providers-amazon-djberg96/lib/aws_products_data_collector.rb:38:in `result'
lib/tasks_private/instance_types.rake:131:in `block (3 levels) in <top (required)>'
lib/tasks_private/instance_types.rake:37:in `isolated'
lib/tasks_private/instance_types.rake:124:in `block (2 levels) in <top (required)>'
/Users/dberger/.rbenv/versions/2.6.5/bin/bundle:23:in `load'
/Users/dberger/.rbenv/versions/2.6.5/bin/bundle:23:in `<main>'
Tasks: TOP => aws:extract:instance_types
(See full trace by running task with --trace)
```

AlexanderZagaynov:
It says it can't connect to amazonaws.com, is that site available in your network?
Here is the pricing API docs which might be helpful: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/using-ppslong.html
And the link I gave you was a backup of the very first version. Since that time I did some refactoring, but anyway, it was never merged, and stayed only in my repo. I'll try to search my archives this weekend, and give you the latest one, which I used to update that yaml file manually.

djberg96:
@AlexanderZagaynov It must have been temporary because I tried again later, though then I got errors about nil descriptions. I would definitely appreciate any updates you have!

AlexanderZagaynov:
Hi @djberg96, apologies for being late.
Looks like I found the most recent version I've used for periodic yaml updates.
https://github.com/ManageIQ/manageiq-providers-amazon/compare/master...AlexanderZagaynov:tasks/instance_types

djberg96:
I never did quite get this finished as it was never completely clear how some of the fields should be translated. I'll leave the branch out there in case anyone wants to pick it up.

PR #615, opened 2020-04-03
Use `zone_id` instead of `zone_name` in the DDF form

skateman:
We can directly send the `zone_id` instead of sending the `zone_name` and looking it up explicitly when handling the API payload. 

@miq-bot assign @agrare 


djberg96:
👍

PR #618, opened 2020-04-22
Update Floating IP target and event collection

djberg96:
Currently classic EIP's in AWS are not getting picked up by targeted refresh. This is because not all AWS floating IP resources have an `allocationId`. Specifically, a "classic" elastic IP will not have an `allocationId`. But, they do have an IP address.

Consequently, I've updated the targeted refresh collection and event parser to look at both `publicIp` and `allocationId` fields. I felt it was best to look for both, since in some cases the `ems_ref` could be either one since the parser looks for both.

~~The only downside to this is that I could no longer use filters, since the AWS SDK treats multiple filters as an AND rather than an OR, and I could see no way to do that within the SDK API. As a result, I'm filtering with pure Ruby after the fact.~~

Edit: Switched to multiple calls with single filters.

I've also updated the event target parser specs, which I think resolve the TODO items that lsmola had mentioned.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1746860

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
@@ -139,12 +139,12 @@ class ManageIQ::Providers::Amazon::Inventory::Collector::TargetCollection < Mana
   def floating_ips
     return [] if references(:floating_ips).blank?
 
-    multi_query(references(:floating_ips)) do |refs|
-      results = aws_ec2.client.describe_addresses.flat_map(&:addresses).select { |addr|
-        refs.include?(addr.allocation_id) || refs.include?(addr.public_ip)
-      }
 
-      hash_collection.new(results).all
+    multi_query(references(:floating_ips)) do |refs|
+      hash_collection.new(
+        aws_ec2.client.describe_addresses(:filters => [{:name => 'allocation-id', :values => refs}]).flat_map(&:addresses) +
+        aws_ec2.client.describe_addresses(:filters => [{:name => 'public-ip', :values => refs}]).flat_map(&:addresses)
+      ).all
     end
   end
 
--- a/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
@@ -140,9 +140,11 @@ class ManageIQ::Providers::Amazon::Inventory::Collector::TargetCollection < Mana
     return [] if references(:floating_ips).blank?
 
     multi_query(references(:floating_ips)) do |refs|
-      hash_collection.new(
-        aws_ec2.client.describe_addresses(:filters => [{:name => 'public-ip', :values => refs}]).flat_map(&:addresses)
-      ).all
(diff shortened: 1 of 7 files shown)


app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -140,9 +140,11 @@ def floating_ips
     return [] if references(:floating_ips).blank?
 
     multi_query(references(:floating_ips)) do |refs|
-      hash_collection.new(
-        aws_ec2.client.describe_addresses(:filters => [{:name => 'allocation-id', :values => refs}]).flat_map(&:addresses)
-      ).all
+      results = aws_ec2.client.describe_addresses.flat_map(&:addresses).select { |addr|
+        refs.include?(addr.allocation_id) || refs.include?(addr.public_ip)
```
agrare, line 144: @djberg96 could you do two filters, one for allocation-id and one for public-ip?

Something like
```
aws_ec2.client.describe_addresses(:filters => [{:name => 'allocation-id', :values => refs}]).flat_map(&:addresses) +
  aws_ec2.client.describe_addresses(:filters => [{:name => 'public-ip', :values => refs}]).flat_map(&:addresses)
```

djberg96: Sure could, though now we're making two calls instead of one, though it would potentially be a memory savings. If you prefer that approach, though, I can make it.

agrare: Yeah just thinking we will only have a hand full of target refs and possibly thousands/tens-of-thousands of total addresses so to bring them all back vs filtering them on the aws side

djberg96: @agrare ok, done

simaishi:
@djberg96 can this be `ivanchuk/yes`?

djberg96:
@simaishi I think it'll be ok. @agrare any objections?

agrare:
:+1: fine by me

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit a1b9403eba0efb5e8fae361826caca325b7f90b8
Date:   Thu Apr 23 08:39:01 2020 -0400

    Merge pull request #618 from djberg96/floating_ip_target_collection

    Update Floating IP target and event collection

    (cherry picked from commit ba32611b728bd7c915151635e682474c125f8c0d)

    https://bugzilla.redhat.com/show_bug.cgi?id=1746860
```

simaishi:
Jansa backport details:

```
$ git log -1
commit 7502c00625a93ad7212dd451c4a8b526a530b4a9
Date:   Thu Apr 23 08:39:01 2020 -0400

    Merge pull request #618 from djberg96/floating_ip_target_collection

    Update Floating IP target and event collection

    (cherry picked from commit ba32611b728bd7c915151635e682474c125f8c0d)

    https://bugzilla.redhat.com/show_bug.cgi?id=1746860
```

Issue #619, opened 2020-04-24
Are relationships between cloud_networks and network_routers correct?


djberg96:
After digging into the Amazon network provider, I'm confused about the way the relationships are setup between `cloud_networks`, `network_routers` and `cloud_subnets`.

To see what I'm talking about, I've created a VPC (dberger-vpc), two subnets (dberger-subnet1 and dberger-subnet2) and two network_routers (dberger-route1 and dberger-route2). Both subnets and both routers are associated with dberger-vpc. However, both subnets are associated with dberger-route1, and none with dberger-route2.

There are two curious things that are currently happening. First, even though both routers are associated with the VPC on the AWS side, only one of them - the one associated with the subnets - actually has a relationship with the cloud network on the ManageIQ side:

```
cloud_network = ManageIQ::Providers::Amazon::NetworkManager::CloudNetwork.find_by(:name => "dberger-vpc")
cloud_network.network_routers # => only dberger-route1
```

Second, even though dberger-route1 can be seen via `cloud_network.network_routers`, the reverse is not true.

```
router = ManageIQ::Providers::Amazon::NetworkManager::NetworkRouter.find_by(:name => "dberger-route1")
router.cloud_network # => nil
```

Note that within the AWS console, the association with the VPC is definitely there (and you can see the vpc-id using the rest API).

I think I know where/why this is happening within the code, but I wanted to double check as to whether or not this was a deliberate design decision before I try to update it.

djberg96:
Looking at the core `CloudNetwork` model, even if I associate the router with the network, it won't pick up all the associated routers because of this:

`has_many :network_routers, -> { distinct }, :through => :cloud_subnets`

But adding this will at least create the inverse relationship between the router and cloud network:

`:cloud_network    => persister.cloud_networks.lazy_find(network_router['vpc_id'])`

djberg96:
Fixed with #621

PR #621, opened 2020-04-29
Establish relationship directly between cloud networks and network routers

djberg96:
The short version is that there's currently a one-way relationship between cloud networks and network routers in the AWS provider. This addresses that.

See https://github.com/ManageIQ/manageiq-providers-amazon/issues/619 for details.


agrare:
@djberg96 can you add a spec test to check the new relationship link?

djberg96:
@agrare Updated, and also removed an outdated refresh type checker that was causing the network router specs to never run.

agrare:
@djberg96 the specs failures look related to this change

djberg96:
Hm, I thought they were passing. Looking...

djberg96:
@agrare Interestingly, if I run `bundle exec rspec spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/targeted_refresh_scope_spec.rb` directly, they pass.

agrare:
> @agrare Interestingly, if I run bundle exec rspec spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/targeted_refresh_scope_spec.rb directly, they pass.

That's because that's not the spec that is failing :rofl:

djberg96:
@agrare Yup I updated the comment. But, they both run `assert_vpc`, only in the one that's failing it doesn't find the router. I'm guessing I need a stub somewhere.

Issue #620, opened 2020-04-27
AWS_EC2_Instance_DELETE events are not always received


john-dupuy:
I have a simple test case that will, 

1) Provision an instance on EC2
2) Delete the instance
3) Look for the `AWS_EC2_Instance_DELETE` event on the timeline of the instance

Out of three runs of this test, only 1 received the `AWS_EC2_Instance_DELETE` and it was not associated with the instance name. When running this manually, I will check the DB for the events, from those 3 test runs, the events received look like: 
```
vmdb_production=# select id, event_type, vm_name, source, timestamp from event_streams where source = 'AMAZON';
 id  |           event_type            |        vm_name         | source |        timestamp        
-----+---------------------------------+------------------------+--------+-------------------------
 567 | AWS_EC2_NetworkInterface_CREATE |                        | AMAZON | 2020-04-27 12:26:03.122
 568 | AWS_EC2_VPC_UPDATE              |                        | AMAZON | 2020-04-27 12:26:03.098
 569 | AWS_EC2_Subnet_UPDATE           |                        | AMAZON | 2020-04-27 12:26:03.175
 570 | AWS_EC2_SecurityGroup_UPDATE    |                        | AMAZON | 2020-04-27 12:26:03.16
 571 | AWS_EC2_Volume_CREATE           |                        | AMAZON | 2020-04-27 12:26:03.143
 572 | AWS_EC2_Instance_CREATE         | test-event-vcqj-jdupuy | AMAZON | 2020-04-27 12:26:03.201
 573 | AWS_EC2_Instance_DELETE         |                        | AMAZON | 2020-04-27 12:28:19.663
 574 | AWS_EC2_SecurityGroup_UPDATE    |                        | AMAZON | 2020-04-27 12:28:19.696
 575 | AWS_EC2_Subnet_UPDATE           |                        | AMAZON | 2020-04-27 12:28:19.713
 576 | AWS_EC2_VPC_UPDATE              |                        | AMAZON | 2020-04-27 12:28:19.679
 583 | AWS_EC2_Volume_CREATE           |                        | AMAZON | 2020-04-27 13:16:44.214
 584 | AWS_EC2_NetworkInterface_CREATE |                        | AMAZON | 2020-04-27 13:16:44.187
 585 | AWS_EC2_Instance_CREATE         | test-event-wwf4-jdupuy | AMAZON | 2020-04-27 13:16:44.274
[...]

PR #633, opened 2020-06-24
Set the monitor_running event to unblock the main thread

agrare:
The do_before_work_loop [starts the event monitor](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/event_catcher/runner.rb#L59).

This [spins up the thread](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/event_catcher/runner.rb#L155-L169) and [waits for it to be started](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/event_catcher/runner.rb#L171).  The amazon event catcher only sets the started event to true [once an event is raised](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/runner.rb#L11).

This means if you try to stop the event catcher before an event occurs you have to SIGKILL it as it will never shutdown cleanly.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/runner.rb
@@ -6,9 +6,9 @@ class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Runner < ManageIQ
   end
 
   def monitor_events
+    event_monitor_running
     event_monitor_handle.poll do |event|
       _log.debug { "#{log_prefix} Received event #{event["messageId"]}" }
-      event_monitor_running
       @queue.enq event
     end
   ensure


jrafanie:
This fixes the issues I saw in https://github.com/ManageIQ/manageiq/pull/20290 where we'd do `event_catcher#kill` on an amazon catcher and it would stick around for 10 minutes, the stopping timeout in pods.  @carbonin suggested I try SIGTERM locally and indeed SIGTERM was seemingly being ignored on my amazon event catcher.  

@agrare  tracked this down to where we [trap interrupts](https://github.com/ManageIQ/manageiq/blob/9600648fdbd0803809742fbf33f25c69b2c99f06/app/models/miq_worker/runner.rb#L420) and [later react to them once do_work completes](https://github.com/ManageIQ/manageiq/blob/9600648fdbd0803809742fbf33f25c69b2c99f06/app/models/miq_worker/runner.rb#L270).  Since `do_work` calls `start_event_monitor` in the base event catcher and `start_event_monitor` won't start until the first event is received, it never gets the exit in the event of a tiny setup with very few events... it couldn't be more obvious now...

djberg96:
@agrare Don't suppose there's a way to spec this, is there?

agrare:
@djberg96 the existing tests for Worker Runners are pretty hacky and have to stub a bunch of stuff to not actually start up a worker, so there isn't an easy way to spec this

agrare:
For reference this is what the VMware event catcher does: https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/infra_manager/event_catcher/runner.rb#L27-L29

djberg96:
@agrare Could there be a connection between this and https://github.com/ManageIQ/manageiq-providers-amazon/issues/620 ?

agrare:
Hm that's a good question, my first reaction was no because once an event is caught and yielded to that block it would set the event that the monitor thread was started.  There could be some weirdness though in the fact that the main runner thread isn't in the do_work loop where it actually processes events until that event is set.

simaishi:
Jansa backport details:

```
$ git log -1
commit dd5052771d9ab30b215be05dbc92dac6c3ef60fa
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Thu Jun 25 12:09:17 2020 -0400

    Merge pull request #633 from agrare/event_monitor_never_sets_monitor_running

    Set the monitor_running event to unblock the main thread

    (cherry picked from commit 3d9f0905fbb5561ccf452be25496759385a71fd8)
```

PR #634, opened 2020-07-13
Remove unused workers and worker config

agrare:
The NetworkManager has an event_catcher.rb and a metrics_collector_worker.rb which aren't used (don't have runners and not require_nested), and the config/settings.yml has network and storage ebs refresh_worker config which don't have workers.


djberg96:
@agrare is it possible these are used depending on the configuration settings?

agrare:
No because the MiqWorkerTypes don't include these and there are no runners:

```
>> pp MiqWorkerType.worker_class_names.select { |n| n.start_with?("ManageIQ::Providers::Amazon") }
   (0.8ms)  SELECT "miq_worker_types"."worker_type" FROM "miq_worker_types"
["ManageIQ::Providers::Amazon::AgentCoordinatorWorker",
 "ManageIQ::Providers::Amazon::CloudManager::EventCatcher",
 "ManageIQ::Providers::Amazon::CloudManager::MetricsCollectorWorker",
 "ManageIQ::Providers::Amazon::CloudManager::RefreshWorker",
 "ManageIQ::Providers::Amazon::StorageManager::S3::RefreshWorker"]
```

It appears that Ladas intended to remove these in https://github.com/ManageIQ/manageiq-providers-amazon/commit/15db06f9a423af55afed7267448b615507703225 but left some files around

PR #642, opened 2020-08-03
[WIP] Instance types renewal update

djberg96:
These were some tasks that @AlexanderZagaynov had created before he left. Until we can get switched over to v3 we can use this to renew the AWS instance types.

I'm not sure if want to implement this or wait for v3 so I've marked it WIP, but I'd at least like to have this available.


lib/tasks_private/lib/aws_products_data_collector.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
+
+# https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/using-ppslong.html
+
+class AwsProductsDataCollector
+  OFFERS_HOSTNAME = 'https://pricing.us-east-1.amazonaws.com'
+
+  cattr_accessor :cache, :instance_writer => false do
+    cache_dir = Rails.root.join('tmp', 'aws_cache', 'products_data')
```
agrare, line 9: Could you use a `Tmpdir` that you wouldn't have to have a separate cleanup rake task?

djberg96:
Something's wrong. The expected flavor numbers keep shifting.

djberg96:
Never did work out the issue, closing. I'll leave the branch alive in case anyone else ever wants to pick this up.

PR #647, opened 2020-09-08
Fix bug where new regions are not reflected in the list

Fryguy:
While technically the providers are static, they can be prevented or exposed via Settings.  Since we were memoizing a dynamic value, subsequent changes were not reflected.


agrare (approved):

agrare:
We do not use DDF forms for provider creation on jansa so this isn't necessary

PR #648, opened 2020-09-09
Prefer to not memoize at all for dynamic params

Fryguy:
This PR also leverages the DDF.find_field method for cleaner specs.

jansa/yes? since #647 is jansa/yes?


agrare (approved):

agrare:
We do not use DDF forms for provider creation on jansa so this isn't necessary

PR #652, opened 2020-09-23
Keep the parent manager's tenant_id in sync with its child managers

agrare:
When syncing child managers' attributes with the parent we should include the tenant_id.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -93,16 +93,19 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
     if network_manager
       network_manager.zone_id         = zone_id
       network_manager.provider_region = provider_region
+      network_manager.tenant_id       = tenant_id
     end
 
     if ebs_storage_manager
       ebs_storage_manager.zone_id         = zone_id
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -93,16 +93,19 @@ def ensure_managers_zone_and_provider_region
     if network_manager
```
agrare, line 93: I don't like the name `ensure_managers_zone_and_provider_region` since as we keep more attributes in sync the method will become comically long.  `sync_child_managers` or something?

djberg96: Could we move those callbacks into the individual managers somehow?

agrare: I don't think we can have a callback on the NetworkManager fire when the CloudManager changes

simaishi:
Jansa backport details:

```
$ git log -1
commit 81e912c1262c041382b4844898de91b03362bb97
Date:   Tue Sep 29 09:42:53 2020 -0400

    Merge pull request #652 from agrare/sync_tenant_id_child_managers

    Keep the parent manager's tenant_id in sync with its child managers

    (cherry picked from commit 3ef13358f74a17b426a1e54cb2c649aaa4a97c8d)
```

PR #657, opened 2020-10-08
Declared DDF schema for cloud volume creation and editing

skateman:
This is an initial Cloud Volume form schema for the fields that are specific to the amazon-provided cloud volumes. I couldn't find any place where the possible types were declared, so I created a constant that now holds them.

The idea is to call this method from the API when an options request hits the `/api/cloud_volumes` endpoint with a valid `ems_id`. I am not entirely sure if the method should take the `ems_id` as an argument or I should do a lookup in the API before calling it, but this is how it works in the provider forms as well.

@miq-bot add_reviewer @agrare 

Parent issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/7334

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
@@ -113,7 +113,7 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
           :id         => 'size',
           :label      => _('Size (in bytes)'),
           :type       => 'number',
-          :step       => 1024 * 1024 * 1024,
+          :step       => 1.gigabytes,
           :isRequired => true,
           :validate   => [{:type => 'required'}],
           :condition  => {
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
@@ -2,6 +2,14 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
   supports :create
   supports :snapshot_create
 
+  CLOUD_VOLUME_TYPES = {
+    :gp2      => N_('General Purpose SSD (GP2)'),
+    :io1      => N_('Provisioned IOPS SSD (IO1)'),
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -96,6 +104,74 @@ def provider_object(connection = nil)
...
+            :when => 'volume_type',
+            :is   => 'io1',
+          }
+        },
+        {
+          :component => 'select',
+          :name      => 'cloud_volume_snapshot_id',
+          :id        => 'cloud_volume_snapshot_id',
+          :label     => _('Base Snapshot'),
+          :options   => ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolumeSnapshot.where(:ems_id => ems_id).map do |cvs|
```
agrare, line 155: @skateman we can't use the same ems_id to query for availability zones in the [CloudManager :point_up:](https://github.com/ManageIQ/manageiq-providers-amazon/pull/657/files#diff-89fd56220c0624c072f49be1ec3b9b06153377a98664a770ed5723413b18216eR115) and in the `StorageManager`.

We do delegate availability zones to the parent cloud manager so I think what we should do here is:
```
storage_manager = ManageIQ::Providers::Amazon::StorageManager::Ebs.find_by(:id => id)
storage_manager.availability_zones # this will delegate to the parent cloud_manager automatically
storage_manager.cloud_volume_snapshots
```

agrare (changes requested):

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -96,6 +104,131 @@ def provider_object(connection = nil)
...
+  def self.params_for_create(ems)
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'size',
+          :id         => 'size',
+          :label      => _('Size (in bytes)'),
+          :type       => 'number',
+          :step       => 1024 * 1024 * 1024,
```
agrare, line 116: `1.gigabyte` ?

skateman: :rofl: yes

agrare (approved):

PR #658, opened 2020-10-13
Add fields for smartstate docker credentials into the provider form

skateman:
@agrare noticed that I forgot to add the [smartstate docker credentials endpoint](https://github.com/ManageIQ/manageiq-ui-classic/pull/2525/files) into the amazon form, probably because the old form was had a bug and it never showed these fields :confused: 

Screenshot from 2020-10-13 09-45-01

I will need some help with the validation and if it there are other fields as well...

@miq-bot add_label bug


agrare:
@skateman I'm not sure if we can/ever did validate these

agrare:
AgentCoordinator does `@ems.authentications.find_by(:authtype => "smartstate_docker")` and after removing the validation from this that works.  Remove the validation and LGTM

agrare (approved):

agrare:
@skateman is this still WIP?

PR #673, opened 2020-12-16
Fix linter issues

djberg96:
This PR cleans up the last of the rubocop linter issues that I see locally. Specifically:

* The `provider_model` singleton method isn't actually private, so moves it out of the private block.
* Disables the `RescueException` warning (I don't see us changing that without risking breakage).
* Changes something that looks like a comparison (even though it isn't) in `userdata_payload` into more readable code.
* Simple whitespace fix (whitespace between method name and parens) that was causing a group expression warning.
* Disabled a specific useless assignment warning, along with a reference to why we did that.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/provision_workflow.rb
@@ -47,14 +47,14 @@ class ManageIQ::Providers::Amazon::CloudManager::ProvisionWorkflow < ManageIQ::P
     allowed_ci(:availability_zones, [:cloud_network, :cloud_subnet, :security_group], targets.map(&:id))
   end
 
+  def self.provider_model
+    ManageIQ::Providers::Amazon::CloudManager
+  end
+
   private
 
   def dialog_name_from_automate(message = 'get_dialog_name')
     super(message, {'platform' => 'amazon'})
-  end
-
-  def self.provider_model
-    ManageIQ::Providers::Amazon::CloudManager
   end
 
   def security_group_to_availability_zones(src)
--- a/app/models/manageiq/providers/amazon/manager_mixin.rb
+++ b/app/models/manageiq/providers/amazon/manager_mixin.rb
@@ -76,8 +76,7 @@ module ManageIQ::Providers::Amazon::ManagerMixin
       # Pull out the password from the database if a provider ID is available
       secret_access_key ||= find(args["id"]).authentication_password('default')
 
-      # rubocop:disable Lint/UselessAssignment (see #588)
-      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, validate = true, :assume_role => service_account)
+      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, validate = true, :assume_role => service_account) # rubocop:disable Lint/UselessAssignment (see #588)
     end
 
     def raw_connect(access_key_id, secret_access_key, service, region,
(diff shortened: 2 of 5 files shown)


app/models/manageiq/providers/amazon/cloud_manager/provision_workflow.rb
```diff
@@ -47,16 +47,16 @@ def allowed_availability_zones(_options = {})
     allowed_ci(:availability_zones, [:cloud_network, :cloud_subnet, :security_group], targets.map(&:id))
   end
 
+  def self.provider_model
+    ManageIQ::Providers::Amazon::CloudManager
+  end
+
```
agrare, line 53: @djberg96 can you tell if this was intended to be private?  If so I'd prefer either using `private_class_method` or `class << self; private` over making it public

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -76,7 +76,7 @@ def verify_credentials(args)
       # Pull out the password from the database if a provider ID is available
       secret_access_key ||= find(args["id"]).authentication_password('default')
 
-      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, validate = true, :assume_role => service_account)
+      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, validate = true, :assume_role => service_account) # rubocop:disable Lint/UselessAssignment (see #588)
```
agrare, line 79: Lets pull `validate = true` up instead of adding rubocop:disable comments
```
validate = true

raw_connect(..., validate)
```

I just don't want to pass `true` as an argument since that's not very descriptive

agrare: Or we can pull the validate logic out of `raw_connect` and instead do in `verify_credentials`:
```
connection = raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, :assume_role => service_account)
!!validate_connection(connection)
```

I actually like that better, raw_connect just returns the connection.  It is up to verify_credentials to...verify the credentials

app/models/manageiq/providers/amazon/cloud_manager/provision_workflow.rb
djberg96: Based on https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_provision_virt_workflow.rb#L361-L365 I would say it was meant to be public.

app/models/manageiq/providers/amazon/manager_mixin.rb
djberg96: Ok, updated.

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -76,7 +76,8 @@ def verify_credentials(args)
       # Pull out the password from the database if a provider ID is available
       secret_access_key ||= find(args["id"]).authentication_password('default')
 
-      !!raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, validate = true, :assume_role => service_account)
+      connection = raw_connect(access_key, secret_access_key, :EC2, region, proxy_uri, :assume_role => service_account)
+      !!validate_connection(connection)
```
agrare, line 80: Future refactoring but I think we can remove the `verify` logic from `raw_connect` now since that looked like the only caller (we should verify it was the only caller with verify=true)

agrare (approved):

PR #675, opened 2021-01-12
Move logger initialization and configuration into plugins

Fryguy:
Part of https://github.com/ManageIQ/manageiq/pull/20950

@agrare Please review.

### DIFF
--- a/lib/manageiq/providers/amazon/engine.rb
+++ b/lib/manageiq/providers/amazon/engine.rb
@@ -13,6 +13,14 @@ module ManageIQ
         def self.plugin_name
           _('Amazon Provider')
         end
+
+        def self.init_loggers
+          $aws_log ||= Vmdb::Loggers.create_logger("aws.log")
+        end
+
+        def self.apply_logger_config(config)
+          Vmdb::Loggers.apply_config_value(config, $aws_log, :level_aws)
+        end
       end
     end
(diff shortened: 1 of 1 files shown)


lib/manageiq/providers/amazon/engine.rb
```diff
@@ -13,6 +13,14 @@ def self.vmdb_plugin?
         def self.plugin_name
           _('Amazon Provider')
         end
+
+        def self.init_loggers
+          $aws_log ||= Vmdb::Loggers.create_logger("aws.log")
```
Fryguy, line 18: The reason I have `||=` here is that something else could initialize the logger first.  For example, in core, FileDepotS3 needs $aws_log, so we might need to initialize in core anyway.  Similarly, multiple providers might use $fog_log, and each would try to initialize.

NickLaMuro: Do we have to be concerned about this being initialized in multiple threads/processes at all?

NickLaMuro (approved):

lib/manageiq/providers/amazon/engine.rb
Fryguy: I don't believe so.

agrare (approved):

PR #679, opened 2021-01-29
Add a basic container manager for Amazon EKS

agrare:
Add an `Amazon::ContainerManager` which subclasses the main `Kubernetes::ContainerManager` to support Amazon Web Service's Elastic Kubernetes Service (https://aws.amazon.com/eks/)

EKS uses STS or IAM for authentication (https://docs.aws.amazon.com/eks/latest/userguide/managing-auth.html) but after that it is a standard Kubernetes API.  This means that we simply have to add methods to acquire an STS token with a user's access_key/secret_access_key and return that as the bearer token and the base Kubernetes code takes over.

The STS tokens are valid for ~15 minutes which is sufficient for operations and the tokens stay live during a `watch` operation

In order to get an STS token you have to give the eks cluster name, I've chosen to put this in the uid_ems column rather than the name column but I'm open to other suggestions.

TODO:
- [x] Fetch an STS token in raw_connect
- [x] Merge `#provider_region` and `#uid_ems` into instance-level connect_options
- [x] Full refresh
- [x] ~~Streaming refresh / watches (Might tackle this in a follow-up)~~
- [x] `.verify_credentials`
- [x] Update DDF forms
- [x] VCR Refresher specs

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/422

Cross-Repo: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/293

### DIFF
--- a/app/models/manageiq/providers/amazon/container_manager.rb
+++ b/app/models/manageiq/providers/amazon/container_manager.rb
@@ -43,4 +43,11 @@ class ManageIQ::Providers::Amazon::ContainerManager < ManageIQ::Providers::Kuber
     )
   end
   private_class_method :sts_presigned_url
+
+  def connect_options(options = {})
+    super.merge(
+      :region_name  => provider_region,
+      :cluster_name => uid_ems
+    )
+  end
 end
--- a/app/models/manageiq/providers/amazon/container_manager.rb
+++ b/app/models/manageiq/providers/amazon/container_manager.rb
@@ -11,4 +11,36 @@ class ManageIQ::Providers::Amazon::ContainerManager < ManageIQ::Providers::Kuber
   def self.description
     @description ||= "Amazon EKS".freeze
   end
+
+  def self.kubernetes_auth_options(options)
+    auth_options = {}
+
+    # If a user has configured a service account we should use that
+    auth_options[:bearer_token] = options[:bearer] if options[:bearer].present?
+
+    # Otherwise request a token from STS with an access_key+secret_access_key
+    auth_options[:bearer_token] ||= sts_eks_token(options[:username], options[:password], options[:region_name], options[:cluster_name])
+
+    auth_options
+  end
+
+  def self.sts_eks_token(access_key, secret_access_key, region_name, cluster_name)
+    url = sts_presigned_url(access_key, secret_access_key, region_name, cluster_name)
+
+    "k8s-aws-v1.#{Base64.urlsafe_encode64(url).sub(/=+$/, "")}"
+  end
+  private_class_method :sts_eks_token
+
+  def self.sts_presigned_url(access_key, secret_access_key, region_name, cluster_name)
+    require "aws-sdk-sts"
+    sts_client = Aws::STS::Client.new(
+      :credentials => Aws::Credentials.new(access_key, secret_access_key),
+      :region      => region_name
+    )
+
+    Aws::STS::Presigner.new(:client => sts_client).get_caller_identity_presigned_url(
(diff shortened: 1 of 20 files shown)


config/settings.yml
```diff
@@ -87,6 +87,12 @@
   :ec2_network:
     :inventory_collections:
       :saver_strategy: batch
+  :eks:
+    :refresh_interval: 15.minutes
+    :streaming_refresh: false
```
agrare, line 92: NOTE: disabled until I copy over the watch_thread inventory collector and persister classes.

app/models/manageiq/providers/amazon/container_manager.rb
```diff
@@ -0,0 +1,14 @@
+class ManageIQ::Providers::Amazon::ContainerManager < ManageIQ::Providers::Kubernetes::ContainerManager
+  require_nested :EventCatcher
+  require_nested :EventParser
+  require_nested :Refresher
+  require_nested :RefreshWorker
+
+  def self.ems_type
+    @ems_type ||= "eks".freeze
```
Fryguy, line 8: Moving forward, if we could have sensible defaults for the various settings keys, then this can act as the anchor key, and all of the rest can front from it.  In fact, maybe it's better to have this at the engine.rb

agrare: Well the engine is for the whole plugin where this is per manager

Fryguy: Ah yeah true...I still would love to specify less overrides in the child objects though.  So yeah, perhaps the ems_type stays here, but still becomes an anchor.

agrare: Oh yeah :100: %

agrare:
```
>> ManageIQ::Providers::Amazon::ContainerManager.raw_connect('57AB61E8FC45.gr7.us-east-1.eks.amazonaws.com', 443, {:username => "ACCESS_KEY", :password => "SECRET_KEY", :region_name => "us-east-1", :cluster_name => "eks-te
st", :ssl_options => {:verify_ssl => 0}}).discover
=> true
```

:tada:

app/models/manageiq/providers/amazon/container_manager.rb
```diff
@@ -0,0 +1,105 @@
...
+    # Otherwise request a token from STS with an access_key+secret_access_key
+    auth_options[:bearer_token] ||= sts_eks_token(options[:username], options[:password], options[:region_name], options[:cluster_name])
+
+    auth_options
+  end
+
+  def self.sts_eks_token(access_key, secret_access_key, region_name, cluster_name)
+    url = sts_presigned_url(access_key, secret_access_key, region_name, cluster_name)
+
+    "k8s-aws-v1.#{Base64.urlsafe_encode64(url).sub(/=+$/, "")}"
```
agrare, line 33: Ruby equivalent of https://github.com/aws/aws-cli/blob/develop/awscli/customizations/eks/get_token.py#L97-L98

agrare:
@miq-bot cross-repo-tests including https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/422

agrare:
Okay this depends on https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/422 because the instance-level connect params don't match the class-level ones for username/password.

Cross-repo tests for this are green https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/422

chessbyte (approved):
Cross-repo tests are green.  Merging.

PR #684, opened 2021-02-15
Parse flavors from the collector not the persister

agrare:
By parsing flavors using `persister.flavors.find()` we're assuming that the flavor either exists in the persister cache (aka full refresh) or already exists in the database (targeted refresh).

If someone deletes the "unknown" flavor from the database then the `persister.flavors.find("unknown")` call will return a `nil` which we're assuming can never happen in the instances parser.

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
@@ -16,6 +16,20 @@ class ManageIQ::Providers::Amazon::Inventory::Collector::TargetCollection < Mana
     target.manager_refs_by_association.try(:[], collection).try(:[], :name).try(:to_a).try(:compact) || []
   end
 
+  def flavors
+    return @flavors_hashes if @flavors_hashes
+
+    all_instance_types = ManageIQ::Providers::Amazon::InstanceTypes.instance_types
+
+    # Only include flavors which are referenced by the targeted instances
+    targeted_instance_types = instances.collect { |instance| instance["instance_type"] }.uniq.compact
+
+    instance_types = all_instance_types.values_at(*targeted_instance_types)
+    instance_types << all_instance_types["unknown"] if instance_types.any?(nil)
+
+    @flavors_hashes = instance_types.compact
+  end
+
   def instances
     return [] if references(:vms).blank?
     return @instances_hashes if @instances_hashes
--- a/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
@@ -19,15 +19,9 @@ class ManageIQ::Providers::Amazon::Inventory::Collector::TargetCollection < Mana
   def flavors
     return @flavors_hashes if @flavors_hashes
 
-    all_instance_types = ManageIQ::Providers::Amazon::InstanceTypes.instance_types
-
     # Only include flavors which are referenced by the targeted instances
-    targeted_instance_types = instances.collect { |instance| instance["instance_type"] }.uniq.compact
-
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -273,11 +273,11 @@ def instance_hardware(persister_instance, instance, flavor)
       :bitness              => architecture_to_bitness(instance['architecture']),
       :virtualization_type  => instance['virtualization_type'],
       :root_device_type     => instance['root_device_type'],
-      :cpu_sockets          => flavor[:cpus],
+      :cpu_sockets          => flavor[:vcpu],
```
agrare, line 276: Before these were using the result of `persister.flavors.find(instance_type)` which was returning the "parsed" flavor.

The flavors returned by the collector (from db/fixtures/aws_instance_types.yml) have slightly different keys

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -229,8 +229,8 @@ def instances
       status = instance.fetch_path('state', 'name')
       next if collector.options.ignore_terminated_instances && status.to_sym == :terminated
 
-      # TODO(lsmola) we have a non lazy dependency, can we remove that?
-      flavor = persister.flavors.find(instance['instance_type']) || persister.flavors.find("unknown")
```
agrare, line 233: `persister.flavors.find("unknown")` <- this is the issue if the `"unknown"` flavor gets deleted it returns `nil` and doesn't rebuild it as part of the targeted refresh.

If we include the flavors in the targeted refresh then it doesn't matter what gets deleted they'll be re-created properly

app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb
```diff
@@ -16,6 +16,20 @@ def name_references(collection)
...
+    all_instance_types = ManageIQ::Providers::Amazon::InstanceTypes.instance_types
+
+    # Only include flavors which are referenced by the targeted instances
+    targeted_instance_types = instances.collect { |instance| instance["instance_type"] }.uniq.compact
+
+    instance_types = all_instance_types.values_at(*targeted_instance_types)
+    instance_types << all_instance_types["unknown"] if instance_types.any?(nil)
+
+    @flavors_hashes = instance_types.compact
+  end
```
Fryguy, line 31: Not for this PR, but I wonder if it's cleaner to just have the instance_types Hash return "unknown" as a default value.  Then I think this code simplifies to:

```suggestion
  def flavors
    return @flavors_hashes if @flavors_hashes

    # Only include flavors which are referenced by the targeted instances
    targeted_instance_types = instances.collect { |instance| instance["instance_type"] }

    @flavors_hashes = ManageIQ::Providers::Amazon::InstanceTypes.instance_types.values_at(*targeted_instance_types).uniq
  end
```

agrare: Done

Fryguy:
@agrare This PR seems to have broken the core spec/models/provider_tag_mapping_spec.rb

cc @jrafanie

Fryguy:
Backported to `kasparov` in commit 8b2c4ee5fcc6ea0b4abcca2f277009fca48c15fe.

```text
commit 8b2c4ee5fcc6ea0b4abcca2f277009fca48c15fe
Date:   Thu Feb 18 14:06:12 2021 -0500

    Merge pull request #684 from agrare/fix_refresh_nil_flavor_exception
    
    Parse flavors from the collector not the persister
    
    (cherry picked from commit ad195a307c2f78600f06c2b92c62bf7660577990)
```

PR #686, opened 2021-02-17
Fix AgentCoordinatorWorker sync_workers

agrare:
The AgentCoordinatorWorker runs 1 copy if there are any valid amazon providers in the zone and the smartproxy role is enabled.

It was pretending to be a PerEmsWorker and overriding the desired_queue_names method but didn't override the instance queue_name so sync_workers was constantly getting:

```
ManageIQ::Providers::Amazon::AgentCoordinatorWorker.sync_workers Workers are being synchronized: Current: [\"ems_0\"], Desired: [\"ems_agent_coordinator\"]"
```

Instead of trying to be a PerEmsWorker simply make it a standard worker that checks for any valid EMSs before saying it has the required_role.

### DIFF
--- a/app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
+++ b/app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
@@ -1,17 +1,21 @@
 class ManageIQ::Providers::Amazon::AgentCoordinatorWorker < MiqWorker
+  include MiqWorker::ReplicaPerWorker
+
   require_nested :Runner
 
-  include PerEmsWorkerMixin
-
   self.required_roles = ['smartproxy']
+  self.workers        = 1
+
+  def self.has_required_role?
+    super && all_valid_ems_in_zone.any?
+  end
+
+  def self.all_valid_ems_in_zone
+    ems_class.where(:zone_id => MiqServer.my_server.zone.id).select { |e| e.enabled && e.authentication_status_ok? }
+  end
 
   def self.ems_class
     ManageIQ::Providers::Amazon::CloudManager
-  end
-
-  def self.desired_queue_names
-    # All cloud managers will share the same agent coordinator
-    super.any? ? ["ems_agent_coordinator"] : []
   end
 
   def self.kill_priority
--- a/spec/models/manageiq/providers/amazon/agent_coordinator_worker_spec.rb
+++ b/spec/models/manageiq/providers/amazon/agent_coordinator_worker_spec.rb
@@ -5,13 +5,13 @@ describe ManageIQ::Providers::Amazon::AgentCoordinatorWorker do
     end
   end
 
-  describe ".desired_queue_names" do
+  describe ".has_required_role?" do
+    before        { ServerRole.seed }
     let!(:server) { EvmSpecHelper.create_guid_miq_server_zone[1] }
-
-    let(:zone) { server.zone }
+    let(:zone)    { server.zone }
 
     it "returns an empty array if no providers are created" do
-      expect(described_class.desired_queue_names).to eq([])
+      expect(described_class.has_required_role?).to be_falsy
     end
 
     context "with an ems" do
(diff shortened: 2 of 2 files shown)


agrare:
cc @jrafanie

app/models/manageiq/providers/amazon/agent_coordinator_worker.rb
```diff
@@ -1,17 +1,21 @@
...
+  def self.has_required_role?
+    super && all_valid_ems_in_zone.any?
   end
 
-  def self.desired_queue_names
-    # All cloud managers will share the same agent coordinator
-    super.any? ? ["ems_agent_coordinator"] : []
+  def self.all_valid_ems_in_zone
+    ems_class.where(:zone_id => MiqServer.my_server.zone.id).select { |e| e.enabled && e.authentication_status_ok? }
+  end
```
agrare, line 15: Maybe some follow-up refactoring but I could see pulling this out into a `ProviderWorkerMixin` since it is in `PerEmsWorkerMixin` and the `PerEmsTypeWorkerMixin` (and this is neither of those)

agrare: It was also duplicated by the VimBroker, actually the Broker had a lot of similarities to this worker type.  Run only one if there are any valid EMSs of a specific type in the zone.

agrare: https://github.com/ManageIQ/manageiq/pull/21060

jrafanie (approved):
LGTM, any idea how far back this needs to be backported?

agrare:
That's a great question, I think _allllll_ the way back since this was overriding desired_queue_names from the start

chessbyte:
I would only backport it to Jansa and Kasparov, at most - as I doubt we will be releasing errata for more than that.

agrare:
So interestingly something about running `evmserverd` is setting the `queue_name` of this worker to `agent_coordinator_worker`.

If I run simulate_queue_worker the queue_name on the worker record is `nil` but when I run it with `rake evm:start` it is "correct"
Still given the fact that this isn't a queue worker that is just a hack for "run one of these please"

jrafanie:
> So interestingly something about running `evmserverd` is setting the `queue_name` of this worker to `agent_coordinator_worker`.

I posted in the other issue but it's more relevant here...

yeah, I'm not sure but it looks like somehow it's storing the `ems_0` in the queue_name.  

I think Azure event catcher and refresh workers are also `PerEmsWorkerMixin` as the base manager catcher/refresh is and their queue_name is `ems_2` below:

```

vmdb_production=# select id, type, queue_name from miq_workers;
 id |                               type                               |      queue_name
----+------------------------------------------------------------------+-----------------------
  1 | MiqGenericWorker                                                 | generic
...
 16 | ManageIQ::Providers::Azure::CloudManager::MetricsCollectorWorker | azure
 19 | ManageIQ::Providers::Azure::CloudManager::MetricsCollectorWorker | azure
 17 | ManageIQ::Providers::Azure::CloudManager::RefreshWorker          | ems_2
 18 | ManageIQ::Providers::Azure::CloudManager::EventCatcher           | ems_2
(19 rows)
```

I believe it is set via [start_worker_for_ems](https://github.com/ManageIQ/manageiq/blob/8bad745f87faca32f945c8290fe4e76b5583071f/app/models/mixins/per_ems_worker_mixin.rb#L68).    Where is it getting ems_agent_coordinator from in your example?

agrare:
That's right, and the desired queue name is `ems_agent_coordinator` so `queue_name_for_ems("ems_agent_coordinator")` is simply returning the string: https://github.com/ManageIQ/manageiq/blob/8bad745f87faca32f945c8290fe4e76b5583071f/app/models/mixins/per_ems_worker_mixin.rb#L100

Okay one mystery solved, not sure why it is different on podified though

agrare:
Okay I got it now, on an appliance MiqServer is creating the worker record with the create_params (this includes the `:queue_name`).
On Podified the `run_single_worker.rb` script creates the worker record and it doesn't have the same params so there is no queue name.

simaishi:
Kasparov backport details:

```
$ git log -1
commit 0721370426986d8504f2edba601efdfec8d6accf
Date:   Wed Feb 17 15:23:39 2021 -0500

    Merge pull request #686 from agrare/fix_agent_coordinator_worker

    Fix AgentCoordinatorWorker sync_workers

    (cherry picked from commit d7fe2ee98c43f4feddb043d3406a62bb1fe6ce67)
```

simaishi:
Jansa backport details:

```
$ git log -1
commit bde11ab03bc7001ecfa104389be2ed3233868072
Date:   Wed Feb 17 15:23:39 2021 -0500

    Merge pull request #686 from agrare/fix_agent_coordinator_worker

    Fix AgentCoordinatorWorker sync_workers

    (cherry picked from commit d7fe2ee98c43f4feddb043d3406a62bb1fe6ce67)
```

PR #692, opened 2021-03-31
Use pluggable provider secrets for vcr specs

agrare:


config/secrets.defaults.yml
```diff
@@ -0,0 +1,14 @@
+---
+test:
+  amazon_defaults: &amazon_defaults
+    client_id: AMAZON_CLIENT_ID
+    client_secret: AMAZON_CLIENT_SECRET
+  amazon_eks_defaults: &amazon_eks_defaults
+    hostname: hostname
```
Fryguy, line 7: Concerned that "hostname" might be too generic...should we do:

```suggestion
    hostname: amazonhostname
```

agrare: Ok, done

PR #694, opened 2021-04-09
[WIP] Combine Cloud, Network, and EBS refreshes

agrare:
https://github.com/ManageIQ/manageiq/issues/20301


agrare:
This is fully functional against a live provider but it is going to take a bit to get the stubbed refresher specs green

PR #695, opened 2021-04-14
Use Vmdb::Utils.http_proxy_uri when adding AWS provider

agrare:
The typical method for setting a proxy is via the `Settings.http_proxy.ec2` setting which is accessed via the `Vmdb::Utils.http_proxy_uri` method.

This is used during the instance connect/verify methods but not during the class connect/verify methods.  This leads to the inability to add an AWS provider if a proxy is required to access it.

### DIFF
--- a/app/models/manageiq/providers/amazon/manager_mixin.rb
+++ b/app/models/manageiq/providers/amazon/manager_mixin.rb
@@ -59,7 +59,6 @@ module ManageIQ::Providers::Amazon::ManagerMixin
     #     "default" => {
     #       "access_key" => "",
     #       "secret_access_key" => "",
-    #       "proxy_uri => "",
     #       "service_account" => ""
     #     }
     #   }
@@ -68,9 +67,11 @@ module ManageIQ::Providers::Amazon::ManagerMixin
       region           = args["provider_region"]
       default_endpoint = args.dig("authentications", "default")
 
-      access_key, secret_access_key, proxy_uri, service_account = default_endpoint&.values_at(
-        "userid", "password", "proxy_uri", "service_account"
+      access_key, secret_access_key, service_account = default_endpoint&.values_at(
+        "userid", "password", "service_account"
       )
+
+      proxy_uri = http_proxy_uri&.to_s
 
       secret_access_key = ManageIQ::Password.try_decrypt(secret_access_key)
       # Pull out the password from the database if a provider ID is available


agrare:
It looks like this was originally a DDF parameter, but was accidentally dropped in 50311369db1f489ddc1b33792df5441897878b15

Either way we should at least be also using the proxy configured via settings.

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -182,6 +182,12 @@ def self.params_for_create
                           :name      => "endpoints.default.url",
```
agrare, line 182: Not related to this PR, but I don't see where we're using this value anywhere?

Fryguy (approved):

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -68,10 +68,12 @@ def verify_credentials(args)
       region           = args["provider_region"]
       default_endpoint = args.dig("authentications", "default")
 
-      access_key, secret_access_key, proxy_uri, service_account = default_endpoint&.values_at(
-        "userid", "password", "proxy_uri", "service_account"
```
Fryguy, line 72: Does this mean that proxy_uri doesn't come from DFF _at all_ ?

Fryguy: Minor, but if so, we probably should remove line 62 as well.

app/models/manageiq/providers/amazon/manager_mixin.rb
```diff
@@ -68,10 +68,12 @@ def verify_credentials(args)
...
+      access_key, secret_access_key, service_account = default_endpoint&.values_at(
+        "userid", "password", "service_account"
       )
 
+      proxy_uri = http_proxy_uri.to_s
```
Fryguy, line 75: If the proxy is _not_ defined, will an empty string being passed into raw_connect cause a problem?

agrare: Yeah I just noticed this and it actually works but is confusing, fixing

app/models/manageiq/providers/amazon/manager_mixin.rb
agrare: Correct, we don't actually have proxy_uri in the endpoints table so the pre-ddf code must have been updating the settings if you put something here

agrare: > Minor, but if so, we probably should remove line 62 as well.

:+1: done

Fryguy (approved):

app/models/manageiq/providers/amazon/manager_mixin.rb
agrare: So I started out by doing `proxy_uri ||= http_proxy_uri` but then noticed there was no proxy_uri in the ddf params, hence removing it from here

Fryguy:
Backported to `lasker` in commit 7f6d4a936d2b668b0cc3270cad8a1fb0de107e3b.

```text
commit 7f6d4a936d2b668b0cc3270cad8a1fb0de107e3b
Date:   Wed Apr 14 15:42:58 2021 -0400

    Merge pull request #695 from agrare/ec2_proxy_not_used_in_ddf_form
    
    Use Vmdb::Utils.http_proxy_uri when adding AWS provider
    
    (cherry picked from commit 63a39adfd7743868a8af33fb41805a7d361b4533)
```

PR #697, opened 2021-04-20
Added Number Value Validator to Form Schema

DavidResende0:
Found in: Storage->Block Storage->Volumes
Part of the Fix for: https://github.com/ManageIQ/manageiq-ui-classic/issues/7713

Adds the number value validator from Data Driven Forms to the Cloud Volume schema to ensure the user can't enter a negative number for the size field

Preview:

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
@@ -124,7 +124,7 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
           :type       => 'number',
           :step       => 1.gigabytes,
           :isRequired => true,
-          :validate   => [{:type => 'required'}],
+          :validate   => [{:type => 'required'}, {:type => 'min-number-value', :value => 0, :message => _('Size must be greater than or equal to 0')}],
           :condition  => {
             :or => [
               {
@@ -205,7 +205,8 @@ class ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume < ::CloudVol
           :condition => {
             :when => 'volume_type',
             :is   => 'io1',
-          }
+          },
+          :validate  => [{:type => 'min-number-value', :value => 0, :message => _('Number of IOPS Must be greater than or equal to 0')}],
         },
         {
           :component    => 'select',


kbrock (approved):
looking good.

Does this have precident? can I just merge?

DavidResende0:
should be safe to just merge

DavidResende0:
> should be safe to just merge

Hold on I think I missed something

chessbyte:
> should be safe to just merge

the [WIP] in the PR title says otherwise ;-)

chessbyte:
This PR is similar to https://github.com/ManageIQ/manageiq-providers-openstack/pull/708.  Is there a parent issue for both?

DavidResende0:
They are all related to this issue ManageIQ/manageiq-ui-classic#7713, but there's no parent issue that I'm aware of

DavidResende0:
Sorry about that, should be good to merge now.

agrare:
@DavidResende0 take this out of WIP when you are ready for this to be merged

app/models/manageiq/providers/amazon/storage_manager/ebs/cloud_volume.rb
```diff
@@ -205,7 +205,8 @@ def self.params_for_create(ems)
           :condition => {
             :when => 'volume_type',
             :is   => 'io1',
-          }
+          },
+          :validate  => [{:type => 'min-number-value', :value => 0, :message => _('Must be greater than or equal to 0')}],
```
kbrock, line 209: are we including the field name in these?
Wondering if we want to have `Size` in these for localization reasons

DavidResende0: Originally I thought only the size field needed this so I wrote the message accordingly. In hindsight, it may be better to change the size field so `:message` is consistent across all the fields. @kavyanekkalapu thoughts?

kavyanekkalapu: @DavidResende0 Better to have individual messages with name of that attribute, otherwise will we get an avt issue, on screenreader test needs it like that.

kbrock: thnx. @davidresende0 are you going to go ahead and change this message to have the field name in here as well? Lets get these great PRs merged

DavidResende0: yup, made the change

Fryguy:
Backported to `lasker` in commit ae03841f21ee06bab4e72ddafc602b35c2fa688d.

```text
commit ae03841f21ee06bab4e72ddafc602b35c2fa688d
Date:   Tue Apr 20 16:39:12 2021 -0400

    Merge pull request #697 from DavidResende0/cloud-volume-schema-fix
    
    Added Number Value Validator to Form Schema
    
    (cherry picked from commit 4a036c05b41973f35e45a1ea56d3fd2ab54ba205)
```

PR #702, opened 2021-04-27
[KASPAROV] Replace authentication_status_ok? with provider specific version

agrare:
This wasn't required when this was done for master because the provider_worker_mixin refactoring meant this automatically picked up the improved `all_valid_ems_in_zone` method.

Depends:
- [x] https://github.com/ManageIQ/manageiq/pull/21198


agrare:
@Fryguy this is green

PR #706, opened 2021-05-19
Use the AmazonEksCredentials class from Kubeclient

agrare:
Kubeclient has their own AWS EKS Credentials class so we don't need to
do the signing/STS calls ourselves.

### DIFF
--- a/app/models/manageiq/providers/amazon/container_manager.rb
+++ b/app/models/manageiq/providers/amazon/container_manager.rb
@@ -32,30 +32,20 @@ class ManageIQ::Providers::Amazon::ContainerManager < ManageIQ::Providers::Kuber
     auth_options[:bearer_token] = options[:bearer] if options[:bearer].present?
 
     # Otherwise request a token from STS with an access_key+secret_access_key
-    auth_options[:bearer_token] ||= sts_eks_token(options[:username], options[:password], options[:region_name], options[:cluster_name])
+    auth_options[:bearer_token] ||= sts_eks_token(options[:username], options[:password], options[:cluster_name])
 
     auth_options
   end
 
-  def self.sts_eks_token(access_key, secret_access_key, region_name, cluster_name)
-    url = sts_presigned_url(access_key, secret_access_key, region_name, cluster_name)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/container_manager.rb
```diff
@@ -32,30 +32,20 @@ def self.kubernetes_auth_options(options)
...
-  end
-  private_class_method :sts_eks_token
-
-  def self.sts_presigned_url(access_key, secret_access_key, region_name, cluster_name)
+  def self.sts_eks_token(access_key, secret_access_key, cluster_name)
+    require "kubeclient/aws_eks_credentials"
     require "aws-sdk-sts"
-    sts_client = Aws::STS::Client.new(
-      :credentials => Aws::Credentials.new(access_key, secret_access_key),
-      :region      => region_name
```
chessbyte, line 51: potential issue in that kubeclient hardcodes the region [here](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/aws_eks_credentials.rb#L26)

Fryguy: https://github.com/abonas/kubeclient/blob/0dff02a3e501debf60184442456153da74556149/lib/kubeclient/aws_eks_credentials.rb#L23

Fryguy (approved):

app/models/manageiq/providers/amazon/container_manager.rb
agrare: > By default, the AWS Security Token Service (AWS STS) is available as a global service, and all AWS STS requests go to a single endpoint at https://sts.amazonaws.com. AWS recommends using Regional AWS STS endpoints instead of the global endpoint to reduce latency, build in redundancy, and increase session token validity. 

from https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html

So it sounds like it will work with a hard coded region but they recommend not.  I'll put a PR in to kubeclient and see what they say.

PR #707, opened 2021-05-28
Add back an instance_types rake task

agrare:
There used to be a rake task to auto-generate the instance_types.rb (at least as early as https://github.com/ManageIQ/manageiq-providers-amazon/pull/170)

It was removed in https://github.com/ManageIQ/manageiq-providers-amazon/pull/449 with the plan of pulling this info from the AWS API/SDK however that never materialized.

This manifested in provisioning failing because the `vpc_only` attribute was flat out incorrect for a number of flavors which caused the auto-select logic to fail.

### DIFF
--- a/db/fixtures/aws_instance_types.yml
+++ b/db/fixtures/aws_instance_types.yml
@@ -1,176 +1,171 @@
 ---
 a1.2xlarge:
   :architecture:
-  - :x86_64
-  :cluster_networking: 
+  - :arm64
+  :cluster_networking:
   :current_generation: true
   :current_version: true
-  :description: A1 General purpose 2XL
+  :description: A1 Double Extra Large
   :ebs_only: true
-  :ebs_optimized_available: true
+  :ebs_optimized_available: false
   :enhanced_networking: true
   :family: General purpose
   :instance_store_size: 0
   :instance_store_size_gb: 0.0
-  :instance_store_type: 
+  :instance_store_type:
   :instance_store_volumes: 0
   :intel_aes_ni: true
-  :intel_avx: 
-  :intel_avx2: 
-  :intel_turbo: 
+  :intel_avx:
+  :intel_avx2:
+  :intel_turbo:
   :memory: 17179869184
   :memory_gb: 16.0
   :name: a1.2xlarge
-  :network_performance: :up_to_10_gigabit
+  :network_performance: :very_high
   :physical_processor: AWS Graviton Processor
   :processor_clock_speed: 2.3 GHz
   :vcpu: 8
-  :virtualization_type:
-  - :hvm
-  :vpc_only: false
+  :virtualization_type: []
+  :vpc_only: true
 a1.4xlarge:
   :architecture:
-  - :x86_64
-  :cluster_networking: 
+  - :arm64
+  :cluster_networking:
   :current_generation: true
   :current_version: true
-  :description: A1 General purpose 4XL
+  :description: A1 Quadruple Extra Large
   :ebs_only: true
-  :ebs_optimized_available: true
+  :ebs_optimized_available: false
   :enhanced_networking: true
   :family: General purpose
   :instance_store_size: 0
   :instance_store_size_gb: 0.0
-  :instance_store_type: 
+  :instance_store_type:
   :instance_store_volumes: 0
   :intel_aes_ni: true
-  :intel_avx: 
-  :intel_avx2: 
-  :intel_turbo: 
+  :intel_avx:
+  :intel_avx2:
+  :intel_turbo:
   :memory: 34359738368
   :memory_gb: 32.0
   :name: a1.4xlarge
-  :network_performance: :up_to_10_gigabit
+  :network_performance: :very_high
   :physical_processor: AWS Graviton Processor
   :processor_clock_speed: 2.3 GHz
   :vcpu: 16
-  :virtualization_type:
-  - :hvm
-  :vpc_only: false
+  :virtualization_type: []
+  :vpc_only: true
 a1.large:
   :architecture:
-  - :x86_64
-  :cluster_networking: 
+  - :arm64
+  :cluster_networking:
   :current_generation: true
   :current_version: true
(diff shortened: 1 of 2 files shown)


lib/tasks_private/instance_types.rake
```diff
@@ -0,0 +1,83 @@
+namespace 'aws:extract' do
```
agrare, line 1: extract might not be the best name for the namespace here but I wanted to have e.g. `rake aws:extract:regions` and `rake aws:extract:instance_types`

lib/tasks_private/instance_types.rake
```diff
@@ -0,0 +1,83 @@
...
+  desc 'Get / renew instance_types'
+  task :instance_types => [:environment] do
+    # Instance Types provided by the AWS SDK do not have sufficient info for provisioning
+    #
+    # This task creates a db/fixtures/aws_instance_types.yml file with data from
+    #  curl https://raw.githubusercontent.com/powdahound/ec2instances.info/master/www/instances.json > lib/tasks/instance_types_data/instances.json
+    #  https://aws.amazon.com/ec2/instance-types/#instance-type-matrix > lib/tasks/instance_types_data/instance_types.csv
+    #  https://aws.amazon.com/ec2/previous-generation/#Previous_Generation_Instance_Details_and_Pricing_ > lib/tasks/instance_types_data/instance_types_previous.csv
+    #     converted to csv via http://www.convertcsv.com/html-table-to-csv.htm
```
agrare, line 10: TODO need to clean up these comments that aren't useful anymore

lib/tasks_private/instance_types.rake
```diff
@@ -0,0 +1,83 @@
...
+                                instance["network_performance"].downcase.tr(' ', '_').to_sym
+                              end
+
+        result = {
+          :architecture            => instance["arch"].map(&:to_sym).sort,
+          :cluster_networking      => nil,
+          :current_generation      => instance["generation"] != "previous",
+          :current_version         => true,
+          :description             => instance["pretty_name"],
+          # TODO :ebs_only => ???
```
agrare, line 51: I think this is now the last attribute I'm not setting...need to find out where it comes from

db/fixtures/aws_instance_types.yml
```diff
@@ -1,220 +1,233 @@
...
-  :intel_avx2: 
-  :intel_turbo: 
+  :intel_avx:
+  :intel_avx2:
+  :intel_turbo:
   :memory: 17179869184
   :memory_gb: 16.0
   :name: a1.2xlarge
-  :network_performance: :up_to_10_gigabit
+  :network_performance: :very_high
```
agrare, line 23: This was a legit bug in the current yaml, the network performance string is "Up To 10 Gigabit" which didn't match our exact "10 Gigabit" or "20 Gigabit" check

db/fixtures/aws_instance_types.yml
```diff
@@ -1,220 +1,206 @@
...
 a1.2xlarge:
   :architecture:
-  - :x86_64
-  :cluster_networking: 
+  - :arm64
+  :cluster_networking:
   :current_generation: true
   :current_version: true
-  :description: A1 General purpose 2XL
-  :ebs_only: true
```
Fryguy, line 9: Seems `ebs_only` key went away...will that affect anything?

lib/tasks_private/instance_types.rake
Fryguy: This is fine.

lib/tasks_private/instance_types.rake
```diff
@@ -0,0 +1,81 @@
...
+                              instance["network_performance"].downcase.tr(' ', '_').to_sym
+                            end
+
+      result = {
+        :architecture            => instance["arch"].map(&:to_sym).sort,
+        :cluster_networking      => nil,
+        :current_generation      => instance["generation"] != "previous",
+        :current_version         => true,
+        :description             => instance["pretty_name"],
+        # TODO :ebs_only => ???
```
Fryguy, line 50: Oh I see you already have it TODO'd

db/fixtures/aws_instance_types.yml
agrare: https://github.com/ManageIQ/manageiq-providers-amazon/pull/707#discussion_r641597626
Hence WIP 😄

Fryguy:
Overall LGTM - glad the script is relatively tiny.

lib/tasks_private/instance_types.rake
```diff
@@ -0,0 +1,81 @@
+namespace 'aws:extract' do
+  desc 'Get / renew instance_types'
+  task :instance_types => [:environment] do
```
Fryguy, line 3: Do you need environment?  I think you could drop this, and then just require activesupport selectively for things like where you need `.gigabyes`

lib/tasks_private/instance_types.rake
agrare: Okay instance["storage"].nil? co-implies ebs-only :tada:

agrare:
For now I've limited this to just the existing flavors so that the diff cleaner and it is more obvious what is changed by specifically this generator.  I will follow-up with the new flavors that have been added since this was last run in a separate PR.

lib/tasks_private/instance_types.rake
agrare: I was worried I was going to have to pull down the https://aws.amazon.com/ec2/instance-types/ list to get this property but luckily not

Fryguy:
Backported to `lasker` in commit eae927bee2e0954f5f82f22283708ac6ac77f22d.

```text
commit eae927bee2e0954f5f82f22283708ac6ac77f22d
Date:   Fri May 28 14:57:21 2021 -0400

    Merge pull request #707 from agrare/add_instance_types_rake_task
    
    Add back an instance_types rake task
    
    (cherry picked from commit 216036c6845f5553a681cf8929a4abae68130528)
```

PR #708, opened 2021-05-28
Update the current list of aws_instance_types

agrare:


Fryguy:
Backported to `morphy` via https://github.com/ManageIQ/manageiq-providers-amazon/pull/775

PR #718, opened 2021-08-20
Prevent services other than EC2 from causing refresh failures

agrare:
Only access to the EC2 service is required for the refresh to complete, other services are optional and failures there (e.g. due to lack of permissions) shouldn't cause the entire refresh to fail.


app/models/manageiq/providers/amazon/inventory/collector/cloud_manager.rb
```diff
@@ -21,6 +21,10 @@ def cloud_database_flavors
 
   def cloud_databases
     hash_collection.new(aws_rds.client.describe_db_instances.flat_map(&:db_instances))
+  rescue
```
agrare, line 24: NOTE: AWS doesn't have a "base" exception that we could catch here, and there are a number of different exceptions that can be raised in different circumstances (e.g. NetworkingError, AccessDenied, etc...).  These inherit from StandardError only hence no more specific exception being caught here.

PR #721, opened 2021-09-09
[WIP] Normalize the cloud_database status column

agrare:
There are _a lot_ of statuses not sure we need to normalize all of them: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/accessing-monitoring.html#Overview.DBInstance.Status


app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -215,7 +215,7 @@ def cloud_databases
       persister.cloud_databases.build(
         :ems_ref               => cloud_database["dbi_resource_id"],
         :name                  => cloud_database["db_instance_identifier"],
-        :status                => cloud_database["db_instance_status"],
+        :status                => normalize_db_status(cloud_database["db_instance_status"]),
```
chessbyte, line 218: should we store the raw status somewhere as well?

agrare: Yes but in order to backport this I want to merge the normalized status first (which is what the UI is currently calling), backport it, then add a schema migration to add the raw_status in a follow-up

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -224,6 +224,21 @@ def cloud_databases
     end
   end
 
+  def normalize_db_status(raw_status)
+    case raw_status
+    when "available"
+      "available"
+    when "creating", "starting"
```
Fryguy, line 231: Should we add `modifying` here or maybe to `unavailable`?

Fryguy:
Totally separate, I'm wondering if we should eventually add support for the maintenance status (which is different)

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
agrare: Yeah I didn't add a few yet because I wanted to check what the availability of the database is (e.g. I'm assuming that `"backing-up"` is an online backup).  Similarly I'm assuming that `"modifying"` would mean unavailable but want to try to find that out first.

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -224,6 +224,21 @@ def cloud_databases
...
+      "initializing"
+    when "deleting", "failed", "stopped", "stopping", "rebooting"
```
Fryguy, line 233: Thinking we might want `upgrading` here too

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
Fryguy: ah ok that makes sense

Fryguy:
Backported to `morphy` in commit ea57883e9dc116be566a683bf809459397fe445f.

```text
commit ea57883e9dc116be566a683bf809459397fe445f
Date:   Tue Sep 14 09:31:59 2021 -0400

    Merge pull request #721 from agrare/normalize_cloud_database_status
    
    Normalize the cloud_database status column
    
    (cherry picked from commit 96a7d0bc1b48b6c636bde6e4acec0fe35656b335)
```

PR #722, opened 2021-09-24
Convert AuthKey operations to SupportsFeatureMixin

chessbyte:
Goes with https://github.com/ManageIQ/manageiq/pull/21457

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
@@ -1,6 +1,8 @@
 class ManageIQ::Providers::Amazon::CloudManager::AuthKeyPair < ManageIQ::Providers::CloudManager::AuthKeyPair
   AwsKeyPair = Struct.new(:name, :key_name, :fingerprint, :private_key)
 
+  supports :delete
+
   def self.raw_create_key_pair(ext_management_system, create_options)
     ec2 = ext_management_system.connect
     kp = if create_options[:public_key].blank?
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
```diff
@@ -15,16 +17,6 @@ def self.raw_create_key_pair(ext_management_system, create_options)
...
-  def self.validate_create_key_pair(ext_management_system, _options = {})
-    if ext_management_system
-      {:available => true, :message => nil}
-    else
-      {:available => false,
-       :message   => _("The Keypair is not connected to an active %{table}") %
-         {:table => ui_lookup(:table => "ext_management_system")}}
-    end
-  end
-
```
kbrock, line 27: I've noticed similar validation put into `supports` blocks.
It feels like it is not needed here, but technically, the ems is not required for authentications of auth key pairs.

Do we want to:
- add the ems requirement to `AuthKeyPair` or `Authentication`
- add the validation back in?
- not worry about it

kbrock (commented):
good stuff

app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
chessbyte: @kbrock we should be asking the ems if it `supports?(:auth_key_pair_create)`  This set of PRs is focused on AuthKeyPair, so not going off on tangents for other types.

agrare (approved):

PR #742, opened 2022-02-04
Fix bin/setup and bundler conflict

Fryguy:


Fryguy:
Ugh there's a bunch of stuff in the shared setup that reference travis specifically...gonna need to fix that in core first.

.github/workflows/ci.yaml
```diff
@@ -25,13 +25,22 @@ jobs:
...
+    - name: Install system packages
+      run: sudo apt-get install -y libcurl4-openssl-dev
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby-version }}
-        bundler-cache: true
     - name: Prepare tests
       run: bin/setup
+    - name: Set up bundler caching
```
kbrock, line 39: you are doing this twice?

Fryguy: I'm experimenting in this PR...ignore the man behind the curtain 😆

chessbyte:
@Fryguy CI via GitHub Actions is passing for both 2.6 and 2.7.  What else is needed to get this and the related PR over the finish line?

Fryguy:
Updated the PR after some discussion with @agrare.  I've moved the repo cloning and system packages into a `bin/before_install` script, which is called from GitHub Actions before we install Ruby, and is also called from `bin/setup` to preserve that functionality for users.  This also keeps the cloning in one place which is important for the release scripts at branching/tagging time since they need to also change the cloned core ref. (e.g. changing clone of core to najdorf when we cut the najdorf branch on the provider repo).  The `bin/before` script is a shell script since we don't have Ruby yet.

agrare (approved):

Fryguy:
One weird thing is that technically we bundle install twice...this can easily be handled in core, but it only incurs ~15s of overhead, so I'm going to solve that later once we are all back to green

PR #746, opened 2022-03-04
Refactor AuthKeyPair.raw_create_key_pair

agrare:
Merge together with: https://github.com/ManageIQ/manageiq/pull/21759

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
@@ -1,6 +1,4 @@
 class ManageIQ::Providers::Amazon::CloudManager::AuthKeyPair < ManageIQ::Providers::CloudManager::AuthKeyPair
-  AwsKeyPair = Struct.new(:name, :key_name, :fingerprint, :private_key)
-
   supports :create
   supports :delete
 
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/amazon/cloud_manager/auth_key_pair.rb
```diff
@@ -1,6 +1,4 @@
 class ManageIQ::Providers::Amazon::CloudManager::AuthKeyPair < ManageIQ::Providers::CloudManager::AuthKeyPair
-  AwsKeyPair = Struct.new(:name, :key_name, :fingerprint, :private_key)
```
Fryguy, line 2: Was key_name just not used?

agrare: So I'm guessing they created `AwsKeyPair` to match the Openstack object as closely as possible, but key_name was not used by https://github.com/ManageIQ/manageiq/pull/21759/files#diff-b80f64b28488206bf8d41b9b53fb77a5490267a1a91b5d1e4f0897af88951554L43-L46 so I didn't include it.  This is now just returning the params for creating of the record

Fryguy (approved):

kbrock:
- I like the moving away from using an openstruct.
- Would it make sense for `raw_create_*` actually create the object instead of creating parameters? Maybe this is a trend we use elsewhere. sorry for the lazy question

agrare:
> Would it make sense for raw_create_* actually create the object instead of creating parameters? Maybe this is a trend we use elsewhere. sorry for the lazy question

The create_key_pair operation is unique in creating a database record, because if a private key is generated it isn't possible to fetch it again as part of refresh so we have to create the db record here.  The raw_* methods typically do the provider specific bits and creating the db record isn't provider specific which is why I left it out of here.

Fryguy:
Backported to `najdorf` in commit c94d01f4acf4581b76cc755afe9589cbd4f5b63d.

```text
commit c94d01f4acf4581b76cc755afe9589cbd4f5b63d
Date:   Tue Mar 8 14:45:48 2022 -0500

    Merge pull request #746 from agrare/refactor_create_key_pair
    
    Refactor AuthKeyPair.raw_create_key_pair
    
    (cherry picked from commit eab5fa966b6fc7553b1ad0a110e444996dc2b795)
```

PR #748, opened 2022-03-08
Update AWS instance types list

agrare:
Update the list of AWS instance-types


Fryguy:
@agrare What is the mechanism for updating this.  Cause I had an idea 😈 ... We can create a GitHub action that runs, say, weekly, that runs your automation and produces a PR if something changed.  We could do this for other providers too that have these things that we have to keep up to date.  WDYT?

agrare:
Yeah there is a rake task we could definitely automate that

spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/refresher_inventory_object_spec.rb
```diff
@@ -65,7 +65,7 @@
 
   def table_counts_from_api
     counts                           = super
-    counts[:flavor]                  = counts[:flavor] + 5 # Graph refresh collect all flavors, not filtering them by known_flavors
```
agrare, line 68: This has not been a concern for a _long_ time, definitely time to remove this

spec/models/manageiq/providers/amazon/cloud_manager/vcr_specs/refresher_inventory_object_spec.rb
```diff
@@ -63,15 +63,6 @@
     end
   end
 
-  def table_counts_from_api
```
agrare, line 66: Actually `table_counts_from_api` aren't called from anywhere else anymore so we don't need to override this, just moved the service_* counts into the base method https://github.com/ManageIQ/manageiq-providers-amazon/pull/748/files#diff-054ae1b3080de7c9ba1d8a11342cfa4d41b35def7a12a2cf8e1ebaef5d5604b0R169-R171

Fryguy:
Backported to `najdorf` in commit 35fa7d48388f977e03810a1e03962db41c508699.

```text
commit 35fa7d48388f977e03810a1e03962db41c508699
Date:   Tue Mar 15 17:25:25 2022 -0400

    Merge pull request #748 from agrare/update_aws_instance_types
    
    Update AWS instance types list
    
    (cherry picked from commit dd24895cf6e2be14571c5145b400e232d2a94bdd)
```

Fryguy:
Backported to `morphy` via https://github.com/ManageIQ/manageiq-providers-amazon/pull/775

PR #753, opened 2022-03-28
Added fields for create cloud object store container form and support for create

GilbertCherrie:
part of: https://github.com/ManageIQ/manageiq-api/pull/1142

Moved fields to provider repo for the create cloud object store container form as well as add support for creating cloud object store container form.

Fields:
Screen Shot 2022-03-28 at 8 15 50 AM

### DIFF
--- a/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
+++ b/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
@@ -24,6 +24,33 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
     end
   end
 
+  def self.params_for_create(_ems)
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'name',
+          :id         => 'name',
+          :label      => _('Container Name'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
+        },
+        {
+          :component    => 'select',
+          :name         => 'provider_region',
+          :id           => 'provider_region',
+          :label        => _('Region'),
+          :isRequired   => true,
+          :includeEmpty => true,
+          :validate     => [{:type => 'required'}],
+          :options      => ManageIQ::Providers::Amazon::Regions.all.map do |region|
+            {:label => region[:description], :value => region[:name]}
+          end
+        }
+      ]
+    }
+  end
+
   def provider_object(connection = nil)
     connect(connection).bucket(ems_ref)
   end
@@ -45,11 +72,8 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
     # frontend stores name as :name, but amazon expects it as :bucket
-    options[:bucket] = options.delete(:name) unless options[:name].nil?
-
-    region = options[:create_bucket_configuration][:location_constraint]
-    connection = ext_management_system.connect(:region => region)
-    bucket = connection.create_bucket(options)
+    connection = ext_management_system.connect(:region => options[:region])
+    bucket = connection.create_bucket(:bucket => options[:name], :create_bucket_configuration => {:location_constraint => options[:region]})
     {
       :key                   => bucket.name,
       :ems_ref               => bucket.name,


GilbertCherrie:
@miq-bot add-label enhancement

GilbertCherrie:
@miq-bot assign @agrare

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -22,6 +24,38 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
     end
   end
 
+  def self.params_for_create(_ems)
+    managers = ManageIQ::Providers::CloudManager.permitted_subclasses.select(&:supports_regions?)
+    ems_regions = managers.each_with_object({}) do |manager, provider_regions|
+      regions = manager.module_parent::Regions.all.sort_by { |r| r[:description] }
+      provider_regions[manager.name] = regions.map { |region| [region[:description], region[:name]] }
+    end
```
agrare, line 32:
```suggestion
```

agrare (commented):
Hey @GilbertCherrie the Amazon Regions are seeded from the https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/config/regions.yml file so you don't need to loop though all of the managers to get their regions

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -22,6 +24,33 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
...
+          :component    => 'select',
+          :name         => 'provider_region',
+          :id           => 'provider_region',
+          :label        => _('Region'),
+          :isRequired   => true,
+          :includeEmpty => true,
+          :validate     => [{:type => 'required'}],
+          :options      => ManageIQ::Providers::Amazon::Regions.all.map do |region|
+            {:label => region[:description], :value => region[:name]}
+          end
```
agrare, line 48: NOTE we can do this in a follow-up but this is almost exactly what we do for the region selection when creating a provider so we should be able to deduplicate this

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -22,6 +24,33 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
...
+          :name       => 'name',
+          :id         => 'name',
+          :label      => _('Container Name'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
+        },
+        {
+          :component    => 'select',
+          :name         => 'provider_region',
+          :id           => 'provider_region',
```
agrare, line 41: @GilbertCherrie the create method below expects this to be `options[:create_bucket_configuration][:location_constraint]` does that need to change?
Also that says it is a location_constraint so should this be optional?

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -22,6 +24,33 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
     end
   end
 
+  def self.params_for_create(_ems)
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'name',
+          :id         => 'name',
```
agrare, line 33: >     options[:bucket] = options.delete(:name) unless options[:name].nil?

We should confirm if these options are passed in as symbol or string keys in this hash

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
GilbertCherrie: @agrare This is all handled in the ui and api code to convert it to `options[:create_bucket_configuration][:location_constraint]`. Also, I made it required because the old ui form had it required.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
GilbertCherrie: @agrare Just confirmed they are being passed in as symbols:
{:name=>"0", :ems_id=>"87", :create_bucket_configuration=>{:location_constraint=>"ap-east-1"}, :emsType=>"ManageIQ::Providers::Amazon::StorageManager::S3", :parent_emstype=>"ManageIQ::Providers::Amazon::CloudManager"}

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
agrare: > This is all handled in the ui and api code to convert it to options[:create_bucket_configuration][:location_constraint].

That feels very amazon specific and not something that should be done in the core UI/API.  Thinking that we should build the AWS specific `create_bucket_configuration` options in the `raw_cloud_object_store_container_create` method.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -43,9 +70,9 @@ def raw_cloud_object_store_container_clear
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
     # frontend stores name as :name, but amazon expects it as :bucket
-    options[:bucket] = options.delete(:name) unless options[:name].nil?
```
agrare, line 46: I think you still need to set `options[:bucket]` for the aws sdk opts

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -43,9 +70,9 @@ def raw_cloud_object_store_container_clear
...
+    region = options.delete(:region)
 
-    region = options[:create_bucket_configuration][:location_constraint]
+    options[:create_bucket_configuration] = {:location_constraint => region}
     connection = ext_management_system.connect(:region => region)
     bucket = connection.create_bucket(options)
```
agrare, line 77: @GilbertCherrie if `:bucket` and `:create_bucket_configuration` are the only two options you could do:
```
    connection = ext_management_system.connect(:region => options[:region])
    bucket = connection.create_bucket(:bucket => options[:name], :create_bucket_configuration => {:location_constraint => options[:region]})
```

agrare (approved):

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -45,11 +72,8 @@ def raw_cloud_object_store_container_clear
 
   def self.raw_cloud_object_store_container_create(ext_management_system, options)
     # frontend stores name as :name, but amazon expects it as :bucket
-    options[:bucket] = options.delete(:name) unless options[:name].nil?
-
-    region = options[:create_bucket_configuration][:location_constraint]
-    connection = ext_management_system.connect(:region => region)
-    bucket = connection.create_bucket(options)
+    connection = ext_management_system.connect(:region => options[:region])
+    bucket = connection.create_bucket(:bucket => options[:name], :create_bucket_configuration => {:location_constraint => options[:region]})
```
kbrock, line 76: for some reason I was expecting `with_provider_object`.

Not sure if `with_provider_object` works for us here, but I guess I always thought it would close the connection or something.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
```diff
@@ -24,6 +24,33 @@ class ManageIQ::Providers::Amazon::StorageManager::S3::CloudObjectStoreContainer
     end
   end
 
+  def self.params_for_create(_ems)
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'name',
```
kbrock, line 32: for some reason, I expected this to be `bucket` so `:bucket` would come in on the `options` hash.

app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb
kbrock: ooh. I see. you converted `:name` to `:bucket` here.

kbrock (commented):
this looks nice. Just the connection/with_provider_object question.

PR #773, opened 2022-05-24
Change update_instance_types committer/author

agrare:
* Change back to running weekly now that we have solved the permissions issue
* Set the committer/author to miq-bot to prevent unverified commits from the author of the most-recent master commit
* Set the enhancement label on PRs created

### DIFF
--- a/.github/workflows/instance_types.yaml
+++ b/.github/workflows/instance_types.yaml
@@ -28,6 +28,7 @@ jobs:
         author: ManageIQ Bot <bot@manageiq.org>
         committer: ManageIQ Bot <bot@manageiq.org>
         delete-branch: true
+        labels: enhancement
         push-to-fork: miq-bot/manageiq-providers-amazon
         title: Update AWS instance_types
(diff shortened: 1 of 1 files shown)


.github/workflows/instance_types.yaml
```diff
@@ -25,7 +25,10 @@ jobs:
         add-paths: db/fixtures/aws_instance_types.yml
         commit-message: Update AWS instance_types
         branch: update_aws_instance_types
+        author: miq-bot <bot@manageiq.org>
+        committer: miq-bot <bot@manageiq.org>
```
Fryguy, line 29: For consistency, elsewhere I've been doing this:

```suggestion
        author: ManageIQ Bot <bot@manageiq.org>
        committer: ManageIQ Bot <bot@manageiq.org>
```

Like here: https://github.com/ManageIQ/miq_bot/blob/master/bin/before_install#L9-L10 and https://github.com/ManageIQ/font-fabulous/blob/master/bin/deploy#L5-L6

PR #780, opened 2022-07-12
Updates to database schema

MelsHyrule:
Related to https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/403

~~db.t1.micro is now deprecated, as seen by this error: `["status", "Error"], ["message", "Invalid DB Instance class: db.t1.micro"]` and we can see that it is no longer listed [here](https://aws.amazon.com/rds/instance-types/).~~ <- Handled in seperate PR, see discussion below.

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager/cloud_database.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/cloud_database.rb
@@ -59,6 +59,8 @@ class ManageIQ::Providers::Amazon::CloudManager::CloudDatabase < ::CloudDatabase
           :id        => 'username',
           :name      => 'username',
           :label     => _('Master Username'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
         },
         {
           :component => 'password-field',
@@ -66,6 +68,8 @@ class ManageIQ::Providers::Amazon::CloudManager::CloudDatabase < ::CloudDatabase
           :id        => 'password',
           :name      => 'password',
           :label     => _('Master Password'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
         },
       ],
     }


MelsHyrule:
cc @nasark

nasark:
Based on https://aws.amazon.com/rds/instance-types/ it looks like m1-m3 and r3 are also no longer supported. i.e. getting the following `Aws::RDS::Errors::InvalidParameterValue (Invalid DB Instance class: db.m1.micro)`.

MelsHyrule:
@agrare in that case do i remove m1-m3 and r3 as well in this PR and any updates to this list be made via an updated weekly github actions sync task ?

agrare:
Yes lets remove the ones we know at this point to not work, i'm working on the automated method now

agrare:
@MelsHyrule looks good we just need to update the specs to look for a flavor that still exists.

Fryguy:
Do we remove deprecated ones?  I thought they still get listed in case we need the data for old existing instances?

agrare:
@Fryguy we don't have a deprecated attribute on flavors, so what we do is if a flavor isn't found we use an "unknown" flavor: https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb#L252 and deprecated flavors are no longer collected/parsed/saved

Fryguy:
Oh interesting - do we do the same for images?  If not, I wonder if we _should_ store the deprecated types similar to how we do with instance types.

MelsHyrule:
@miq-bot assign @agrare

Fryguy:
The database types were updated in #781

MelsHyrule:
@Fryguy with the merging of that PR then do we keep the deprecated instance types and leave `aws_database_types.yml` as is or do i still remove them?

Fryguy:
Up to @agrare

agrare:
I think we would keep them as long as amazon keeps them, but filter them out for selection when creating new cloud_databases

agrare:
@MelsHyrule I think the only changes needed here now are the `params_for_create`

agrare (approved):

agrare (approved):

PR #781, opened 2022-07-13
Add a rake task to auto-update database types

agrare:
Automatically pull the latest list of rds database instance types from https://instances.vantage.sh/rds/instances.json
Also added this to the weekly cron run to update instance types

### DIFF
--- a/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
@@ -206,7 +206,7 @@ class ManageIQ::Providers::Amazon::Inventory::Parser::CloudManager < ManageIQ::P
       persister.cloud_database_flavors.build(
         :ems_ref => flavor[:name],
         :name    => flavor[:name],
-        :enabled => true,
+        :enabled => !flavor[:deprecated],
         :cpus    => flavor[:vcpu],
         :memory  => flavor[:memory]
       )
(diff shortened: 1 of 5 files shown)


Fryguy:
Will this also be added to the cron job?

lib/tasks_private/database_types.rake
```diff
@@ -0,0 +1,30 @@
+namespace 'aws:extract' do
+  desc 'Get / refresh database_types'
+  task :database_types do
+    def instances
```
Fryguy, line 4: Minor, but why is this a method?  Seems like this could just be inlined.

spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb
```diff
@@ -1726,7 +1726,7 @@ def assert_specific_cloud_database_flavor
       :ems_ref => "db.r3.large",
       :type    => "ManageIQ::Providers::Amazon::CloudManager::CloudDatabaseFlavor",
       :cpus    => 2,
-      :memory  => 16106127360
+      :memory  => 16374562816
```
agrare, line 1729: NOTE the previous value was `15GiB` where the new one has `15.25GiB` I forget where I pulled the old list from but it either didn't include fractional memory or this has been updated since

app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb
```diff
@@ -206,7 +206,7 @@ def cloud_database_flavors
       persister.cloud_database_flavors.build(
         :ems_ref => flavor[:name],
         :name    => flavor[:name],
-        :enabled => true,
+        :enabled => !flavor[:deprecated],
```
agrare, line 209: @Fryguy turns out we do have an `enabled` column already for this

PR #788, opened 2022-08-25
Test with ruby 2.7, based on changes in ManageIQ/manageiq#22073

bdunne:
Test with ruby 2.7, based on changes in ManageIQ/manageiq#22073

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,19 +1,17 @@
+---
 name: CI
-
-on:
+'on':
   push:
   pull_request:
   workflow_dispatch:
   schedule:
-  - cron: '0 0 * * *'
-
+  - cron: 0 0 * * *
 jobs:
   ci:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         ruby-version:
-        - '2.6'
         - '2.7'
     services:
       postgres:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,19 +1,17 @@
+---
 name: CI
-
-on:
+'on':
```
agrare, line 3: Hm don't love this, realize you have the issue with ruby dumping this to `true:` but think a little `sed -i 's/true:/on:/g` would be better (this is what we've done before with blaster outer changes to ci.yaml)

bdunne: Yeah, I don't love it either, I'm surprised they used `on` as a key.

I didn't see any other blaster outer scripts updating ci.yaml

agrare: I didn't create a PR for it but I still have it locally, basically just:
```
gh_actions_ci_yaml = gh_actions_ci.to_yaml
gh_actions_ci_yaml.gsub!(/^true:/, "on:") # YAML replaces on: with true:

File.write(gh_actions_ci_path, gh_actions_ci_yaml)
puts "Wrote updated ci.yml at: #{gh_actions_ci_path}"
```

PR #833, opened 2023-10-26
Refactor regions spec to not hard-code atypical regions

agrare:
The regions spec has a hard-coded list of "atypical regions" which really just means regions which exist but are not available to us via the aws-sdk `describe_regions` API call.

It would be simpler to just check that all of the regions that we do have access to are in the regions list.  This would allow us to update the `config/regions.yml` list in an automated way.


agrare:
I'm questioning how helpful this spec is, since the "online" list of regions is coming from a vcr rather than live then we still cannot update the regions automatically without re-recording the VCR to pick up new ones.

Fryguy (approved):

PR #870, opened 2024-09-30
Testing with ruby 3.2, 3.3

kbrock:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,8 +12,8 @@ jobs:
     strategy:
       matrix:
         ruby-version:
-        - '3.0'
         - '3.1'
+        - '3.3'
     services:
       postgres:
         image: manageiq/postgresql:13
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
         ruby-version:
         - '3.0'
         - '3.1'
+        - '3.2'
+        - '3.3'
```
jrafanie, line 18: Just offering my thoughts on the plugins and core... I think we should be supporting 3.0-3.3 in plugins until we drop 3.0 in the Gemfile.  I don't think we need to test all of those versions in plugins/core as I think it's fine to test only 3.0, 3.1, and 3.3 with the intention of dropping 3.0 as soon as we have smoke tested appliances and pods with rebuilt rpms against 3.3.

Repositories that don't load the core application (aka not core and not a plugin/engine) can test as many rubies and rails versions as make sense based on their usage.  Things tied somewhat to the core rails app without loading it should stray closer to the rails app's dependencies (pg ha admin, logical replication, etc.) while other gems that are completely independent should have their own ruby/rails dependencies (ancestry, default_value_for, optimist, etc.).

Fryguy:
Backported to `radjabov` in commit a4b50649d0d1365aad03a66e9afbc02e56e31467.

```text
commit a4b50649d0d1365aad03a66e9afbc02e56e31467
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Tue Nov 5 16:49:20 2024 -0500

    Merge pull request #870 from kbrock/ruby3.3
    
    Testing with ruby 3.2, 3.3
    
    (cherry picked from commit 5ef88e007b56f5ec266adc223f2d1be0d965b80f)
```

PR #871, opened 2024-10-28
Handle missing networkPerformance and "X Gpbs"

agrare:
The updated AWS database instance types list has some depreacted types without a networkPerformance value.

The old code also wasn't properly handling the "X Gpbs" value only "X Gigabit" which was causing some values to be set as `:20_gbps` instead of `:very_high`

### DIFF
--- a/lib/tasks_private/database_types.rake
+++ b/lib/tasks_private/database_types.rake
@@ -8,11 +8,8 @@ namespace 'aws:extract' do
     end
 
     results = instances.map do |instance|
-      network_performance = if instance["networkPerformance"].match?(/(\d+ Gigabit)/)
-                              :very_high
-                            else
-                              instance["networkPerformance"].downcase.tr(' ', '_').to_sym
-                            end
+      network_performance = instance["networkPerformance"]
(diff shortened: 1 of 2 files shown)


lib/tasks_private/database_types.rake
```diff
@@ -8,11 +8,8 @@ namespace 'aws:extract' do
     end
 
     results = instances.map do |instance|
-      network_performance = if instance["networkPerformance"].match?(/(\d+ Gigabit)/)
-                              :very_high
-                            else
-                              instance["networkPerformance"].downcase.tr(' ', '_').to_sym
-                            end
+      network_performance = instance["networkPerformance"]
+      network_performance = network_performance.match?(/(\d+ Gigabit)|(\d+ Gbps)/) ? :very_high : network_performance.downcase.tr(' ', '_').to_sym if network_performance
```
agrare, line 12: I'm not sure that :very_high is better than :25_gigabit especially with the number of >1gbps continuing to increase we're losing out on some valuable information.  I figured I'd fix it so that we were consistent then change it in the future.

agrare: Actually it looks like we don't display this anywhere...

PR #874, opened 2024-10-30
Delegate refresh_ems class method to parent cloud manager

nasark:
Refreshes should be done via the cloud manager. Instance methods are already delegated however the class level refresh methods were not delegated.

Following changes in https://github.com/ManageIQ/manageiq-providers-azure/pull/564

@miq-bot assign @agrare 
@miq-bot add_label bug
@miq-bot add_reviewer @agrare


agrare (approved):
:+1: LGTM
I notice we still do the old "queue refresh of child managers" in the CloudManager/Refresher this is something we should convert over here also

Fryguy:
Backported to radjabov via merge of master into radjabov

PR #883, opened 2025-01-30
Rename default_blacklisted_event_names

agrare:
The Amazon CloudManager ems_type is ec2 not amazon so we still have to override the filtered_event_names method.

https://github.com/ManageIQ/manageiq/pull/23320

### DIFF
--- a/app/models/manageiq/providers/amazon/cloud_manager.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager.rb
@@ -108,7 +108,7 @@ class ManageIQ::Providers::Amazon::CloudManager < ManageIQ::Providers::CloudMana
     false
   end
 
-  def self.default_blacklisted_event_names
+  def self.filtered_event_names
     Settings.ems.ems_amazon.blacklisted_event_names
   end
 
--- a/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/runner.rb
@@ -28,7 +28,7 @@ class ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Runner < ManageIQ
   private
 
   def filtered?(event)
-    filtered_events.include?(event["eventType"])
+    @ems.filtered_event_names.include?(event["eventType"])
   end
 
   def event_monitor_handle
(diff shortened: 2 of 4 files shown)


spec/models/manageiq/providers/amazon/network_manager_spec.rb
```diff
@@ -21,16 +21,6 @@
...
-    it "returns the expected value for the default_blacklisted_event_names method" do
-      array = %w(
-        ConfigurationSnapshotDeliveryCompleted
-        ConfigurationSnapshotDeliveryStarted
-        ConfigurationSnapshotDeliveryFailed
-      )
-
-      expect(described_class.default_blacklisted_event_names).to eq(array)
-    end
-
```
agrare, line 33: The NetworkManager::EventCatcher was removed a long time ago in 15db06f9a423af55afed7267448b615507703225 and the default_blacklisted_event_names method was removed in 991a51a4c9d4eba63a3694a169007d899c56ac11 so this test is not necessary.

app/models/manageiq/providers/amazon/cloud_manager.rb
```diff
@@ -108,7 +108,7 @@ def self.hostname_required?
     false
   end
 
-  def self.default_blacklisted_event_names
+  def self.filtered_event_names
     Settings.ems.ems_amazon.blacklisted_event_names
```
Fryguy, line 112: Can we add a comment here, similar to vmware, about why we aren't inheriting the base class method?

app/models/manageiq/providers/amazon/cloud_manager/event_catcher/runner.rb
```diff
@@ -28,7 +28,7 @@ def process_event(event)
   private
 
   def filtered?(event)
-    filtered_events.include?(event["eventType"])
+    @ems.filtered_event_names.include?(event["eventType"])
```
Fryguy, line 31: I think this can be just 
```suggestion
    filtered_event_names.include?(event["eventType"])
```

because the base class provides the `filtered_event_names` method that goes through the `@ems`

Fryguy: Oh no, I misread, the method is on the EMS, but this is the runner.

PR #887, opened 2025-03-12
Lock down aws-sdk-core to before 3.220.1

agrare:
Prevent a change to aws-sdk-core stubbing causing failures in our amazon authenticator specs.
This can be unlocked once we fix stubbing issue.


agrare:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-amazon/pull/888

Issue #895, opened 2025-05-28
Event Triggered Targeted Refresh broken


agrare:
When starting/stopping instances events are caught and handled properly, but parsing the event targets for the purpose of queuing a targeted refresh returns `[]` so no refresh is queued.

This appears to be due to the `event_hash[:full_data]` going through a `deep_symbolize` process causing our `EventTargetParser#parse` to not find any data.

```
EmsEvent.add_queue:
[...]

PR #897, opened 2025-05-29
Fix EventTargetParser using string keys

agrare:
After we started using kafka for the event handler we were doing a deep_symbolize_keys on the message payload which caused this to start breaking.

We will need some further changes to ensure this works for the miq_queue dequeue_method as well, and I need to check all of the other EventTargetParsers to see if they are also impacted or not.

Fixes https://github.com/ManageIQ/manageiq-providers-amazon/issues/895


app/models/manageiq/providers/amazon/cloud_manager/event_parser.rb
```diff
@@ -53,13 +53,13 @@ def self.event_to_hash(event, ems_id)
     end
 
     log_header = "ems_id: [#{ems_id}] " unless ems_id.nil?
-    _log.debug("#{log_header}event: [#{event[:message]}]")
+    _log.debug("#{log_header}event: [#{event_hash[:message]}]")
```
agrare, line 56: Unrelated to this change, but this was always `nil` because we set `:message` in the `event_hash` it isn't in the native `event` payload from amazon.  Think this was just a mistake in the original implementation 609f666a

spec/models/manageiq/providers/amazon/cloud_manager/event_target_parser_spec.rb
```diff
@@ -1669,6 +1669,6 @@ def response(path)
   def create_ems_event(path)
     event = ManageIQ::Providers::Amazon::CloudManager::EventCatcher::Stream.new(double).send(:parse_event, response(path))
     event_hash = ManageIQ::Providers::Amazon::CloudManager::EventParser.event_to_hash(event, @ems.id)
-    EmsEvent.add(@ems.id, event_hash)
+    EmsEvent.add(@ems.id, event_hash.deep_symbolize_keys)
```
agrare, line 1672: This simulates how the event_hash is processed by the event_handler: https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_event_handler/runner.rb#L3

Fryguy:
I'm happy to merge as is, but I am curious why we are deep symbolizing a payload that isn't ours.  The full_data column was meant to hold the "original" event in case we need any internal data.

Fryguy:
Backported to `spassky` in commit 88fbdb824658c38a12d5a625c6625a6f970b2373.

```text
commit 88fbdb824658c38a12d5a625c6625a6f970b2373
Date:   Fri May 30 09:36:48 2025 -0400

    Merge pull request #897 from agrare/fix_event_target_parser_symbol_keys
    
    Fix EventTargetParser using string keys
    
    (cherry picked from commit cb7a982ba2a1fb1ed7d8b447d8af31b65c3ebc0d)
```

PR #967, opened 2026-09-18
[WIP] Support running ContainerManager EventCatcher as a non-Rails worker

ohhAndy:
Highlight: `~300MB -> ~89MB (~70% lower)`

Part of https://github.com/ManageIQ/manageiq/issues/22955

Depends on: 
- https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/569

@miq-bot assign @agrare
@miq-bot add-label enhancement


app/models/manageiq/providers/amazon/container_manager/event_catcher.rb
```diff
@@ -1,4 +1,10 @@
 class ManageIQ::Providers::Amazon::ContainerManager::EventCatcher < ManageIQ::Providers::BaseManager::EventCatcher
+  self.worker_settings_paths = [
+    %i[http_proxy ec2],
+    %i[log level_aws],
```
Fryguy: The logging is interesting - Wonder if this should also be in the base? I'm surprised this lands in worker settings at all.

ohhAndy: Bob actually hallucinated the logging settings, from the vmware infra example, so its actually not needed.

